PR #1, opened 2015-05-05
Create basic django project foundation

carsongee:
- Load configuration from environment or yaml file
- Use dj-static for static file serving
- Add docker images for postgresql based development environment
- Add instructions to README for getting started.


.gitignore
```diff
@@ -0,0 +1,58 @@
...
+db.sqlite
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
+
+node_modules/
+
```
pwilkins, line 58: Consider adding the common editor backup file extensions such as these from the edx-platform .gitignore:

### Editor and IDE artifacts

*~
*.swp
*.orig

/nbproject
.idea/
.redcar/
codekit-config.json
.pycharm_helpers/

### OS X artifacts

*.DS_Store

carsongee, line 58: Cool, added

bdero, line 58: I usually side with it being the individual developer's responsibility to add these things to their [global excludes](http://git-scm.com/docs/gitignore) file, which caters to their specific OS and tooling (since I tend to think of versioned gitignores as only encompassing project-specific generated files, rather than every possible generated file produced by every possible environment).

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig.

That being said, I don't actually have a really strong preference on this. :wink:

pwilkins, line 58: Hearing you say that, it makes better sense since you could go nuts trying to accommodate every editor or IDE that someone could choose. :-)

carsongee, line 58: I don't think there is anything wrong with belt and suspenders here, but thanks for the info @bdero

pdpinch:
@carsongee should we expect this to be passing tests? Is there enough of the skeleton in place yet?

carsongee:
No, it was a messed up travis config, it defaults to testing all branches even if they don't have a travis.yml, though since no one has really looked at this, we could just do the review on #2 since it has all this plus the testing

pdpinch:
Ok. Merging despite the test failures.

PR #7, opened 2015-05-11
Updates README

ShawnMilo:
"setup" is a noun, "set up" is a verb
adds hyphenation and capitalization


pdpinch:
@ShawnMilo we have to find you something to do with your free time... :+1:

PR #9, opened 2015-05-13
Add taxonomy app with models

noisecapella:
This isn't done, there are no API methods yet, but how does this look from a Django perspective? Is everything in the right place? Any other comments on the model?


lore/settings.py
```diff
@@ -67,6 +67,7 @@ def get_var(name, default):
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'taxonomy.apps.TaxonomyConfig',
```
ShawnMilo: The import in `INSTALLED_APPS` name is usually single value (like `TaxonomyConfig`, and lower-case. So the "app" may be on the wrong path. Also, it appears that `TaxonomyConfig` is a class, and the class name isn't what's added to `INSTALLED_APPS`; the app folder itself (the one usually containing `models.py`) is.

Unless I'm behind and this is all good in Django 1.7+...

taxonomy/models.py
```diff
@@ -0,0 +1,31 @@
+"""Models for lore taxonomy application"""
+
+from django.db import models
+
+
+class Vocabulary(models.Model):
+    repository = models.ForeignKey("Repository")
+    name = models.TextField()
```
ShawnMilo, line 15: To prevent `null` values in the database, if this field isn't required it should have `blank=True, default=""` in the definition. Also, I believe a `TextField` requires a `max_length` value.

taxonomy/models.py
```diff
@@ -0,0 +1,31 @@
...
+    description = models.TextField()
+    required = models.BooleanField()
+    type = models.CharField(max_length=1, choices=(
```
ShawnMilo: `type` is a reserved word in Python, so this column name should be modified.

ShawnMilo:
@noisecapella: Cool, it's shaping up! Since you asked for Django comments I added a couple. The first part about the `INSTALLED_APPS` might be wrong since I haven't used Django 1.7 or greater, so FYI.

taxonomy/models.py
noisecapella: It's not technically a reserved word (https://docs.python.org/2/reference/lexical_analysis.html#keywords) but I see your point. Should I call it _type and specify 'type' as the database column name?

carsongee: I would probably just change it to vocabulary_type, it adds some typing and redundancy, but is clear, and underscore naming it would break the private variable/method convention

ShawnMilo: I agree with Carson about the name. Also, I made a mistake; it's not a keyword, but it _is_ a [built-in function](https://docs.python.org/2/library/functions.html) in Python.

taxonomy/models.py
noisecapella, line 15: From the [TextField](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.TextField) docs, it looks like max_length is enforced only in the UI for TextField (since Django 1.7). It is required for CharField though.

In MySQL there's the VARCHAR type which stores text inline and is indexable and the TEXT type which is somewhat out of band and is not indexable, but is able to store larger amounts of text (> 64k). Does this translate to postgres? Any insight on whether I should use CharField or TextField in general? I'm guessing CharField should be used here if postgres has similar ideas about storing text

ShawnMilo:
@noisecapella: I received a comment from you via e-mail which I now can't find on github, so I'm quoting it here:

> In MySQL there's the VARCHAR type which stores text inline and is indexable and the TEXT type which is somewhat out of band and is not indexable, but is able to store larger amounts of text (> 64k). Does this translate to postgres? Any insight on whether I should use CharField or TextField in general? I'm guessing CharField should be used here if postgres has similar ideas about storing text

I don't know the current implementation details of what field types the ORM creates in its migrations based on the field types, but I think @trski has more recent and in-depth knowledge of this and might have an opinion. You could also create and run the migrations in Postgres and dump the schema to see (or poke around in Django's codebase, which might be rewarding but take a lot longer).

noisecapella:
It's under the second 'show outdated diff', got buried because I pushed a change to models.py.

I did some research and it looks like postgres doesn't do anything different for its TEXT and VARCHAR types (some more information [here](http://stackoverflow.com/questions/7354588/django-charfield-vs-textfield)), so I'll use TextField from now on.

noisecapella:
It looks like [nulls are disabled by default](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.null), blank=False by default and all fields are required by default. Should any of these fields provide default values? I'm thinking it's probably safer to require everything to be specified when creating the Field

noisecapella:
The latest commit adds the API methods. They still need work and I have a few questions about these:
- do functions take id numbers or Django models?
- Should the LearningObjectType functions take an LearningObjectType id number or should it use the name column to do lookup?
- Should they return Django models or managers, or something abstracted away from that?
- Ok to input kwargs instead of enumerating the parameters?

taxonomy/models.py
```diff
@@ -0,0 +1,110 @@
...
+    """Model for term table"""
+    vocabulary = models.ForeignKey(Vocabulary)
+    label = models.TextField()
+    weight = models.IntegerField()
+
+    learning_objects = models.ManyToManyField(LearningObject,
+                                              related_name="terms")
+
+
+def get_vocabulary(vocabulary_id):
```
carsongee: If we decide to go this way, I think we would want these in a different file, like api.py or similar, we would also want to have a unit test for each of these to verify they return what we expect.

Ferdi: That would be great !

carsongee:
> do functions take id numbers or Django models?

If they take models, then they already have the object and not need to call the API, so I would think IDs, but @ShawnMilo would understand more what we are trying to achieve.

> Should the LearningObjectType functions take an LearningObjectType id number or should it use the name column to do lookup?

We can do both, just do a type check to figure out which, or write two functions

> Should they return Django models or managers, or something abstracted away from that?

Another question for @ShawnMilo, but I would think we would want to return just a dictionary

> Ok to input kwargs instead of enumerating the parameters?

Speaking for myself, I would want them enumerated so that pydoc and or sphinx can tell me what I should be passing in that makes up a new object.  A docstring in google style with each argument would be preferred, but we can slack on that a little bit for now.

noisecapella:
I'd rather not do a type check if possible for LearningObjectType, maybe just input and output the name and leave the id number as a database implementation detail. Everything else sounds good

noisecapella:
Regarding this code snippet:

```
def add_learning_object_type_for_vocabulary(learning_object_type, vocabulary_id):
    vocabulary = Vocabulary.objects.get(id=vocabulary_id)
    learning_object_type = LearningObjectType.objects.get(name=learning_object_type)
    vocabulary.learning_object_types.add(learning_object_type)
```

The two `get()`s don't seem to be required since the SQL is just inserting a row with two integer ids for the m2m table. Is there another Django-friendly way to do this? This may also be premature optimization, just wondering

ShawnMilo:
I'd say don't worry about it for now. If we take what @carsongee said and just develop them as we need them, they'll make more sense as we write them -- and they should also have tests written alongside them, which will make us think about them more. And then there will be code reviews. 

As for the original question, a look at your sample code makes me think there should be a way to insert a record just having the primary keys, but [the documentation](https://docs.djangoproject.com/en/1.8/ref/models/relations/) doesn't seem to bear that out.

taxonomy/apis.py
```diff
@@ -0,0 +1,226 @@
+"""APIs for lore taxonomy application"""
+
+from taxonomy.models import Vocabulary, Term
+from learningobjects.models import LearningObject, LearningObjectType
+
+
+def get_vocabulary(vocabulary_id):
+    """
+    Lookup vocabulary given its id
+    :param vocabulary_id: Vocabulary id
```
carsongee: Totally open for discussion, but we've actually been using https://pypi.python.org/pypi/sphinxcontrib-napoleon/0.3.3 and the Google style guide for doc strings: http://google-styleguide.googlecode.com/svn/trunk/pyguide.html

I just tend to prefer the look, but we can change our minds, no problem.  Also the params are missing types, like `str`, `int`, etc.

taxonomy/apis.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def _get_term_from_model(term):
+    """
+    Convert Term model to dictionary
+    :param term: Term
+    :type term: Term
+    :return: Dictionary
+    """
+    return {
```
carsongee: Instead of these wrapper functions, you can just to model to object with `.values()`https://docs.djangoproject.com/en/dev/ref/models/querysets/#values or django.forms.models.model_to_dict

taxonomy/apis.py
noisecapella: Unfortunately PyCharm doesn't support Google style docstrings which makes me prefer reStructuredText instead. I agree with the need for types.

taxonomy/apis.py
noisecapella: One complication is that _get_vocabulary_from_model returns a list of learning_object_type strings, I'm not sure what `values()` would return, probably nothing for that.

But we don't really know that we need to include this in the vocabulary dictionary, we could require a separate function for that instead. Might be best to leave it out until we need it.

noisecapella:
This assumes certain changes from the learningobjects portion of pull request #21 are merged in before this is merged

noisecapella:
Will reopen as pull request off skm_import branch

PR #10, opened 2015-05-13
(WIP) learning objects app

ghost:


lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+    org = models.TextField()
+    course_number = models.IntegerField()
+    semester = models.TextField()
+    import_date = models.DateField()
+    imported_by = models.TextField()
+
+class LearningObject(models.Model):
+    course = models.ForeignKey(Course)
+    learning_object_type = models.ForeignKey(LearningObjectType)
+    uuid = models.TextField()
```
ShawnMilo: News to me, but evidently Django has added a [uuid](https://docs.djangoproject.com/en/1.8/ref/models/fields/#uuidfield) field type.

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+    title = models.TextField()
+    description = models.TextField()
+    content_xml = models.TextField()
+    path_xml = models.TextField()
+    mpath = models.TextField()
```
Ferdi: maybe rename to "materialized_path" ?

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class LearningObjectType(models.Model):
+    learning_object = models.ForeignKey(LearningObject)
+    name = models.TextField()
+    
+class Repository(models.Model):
+    name = models.TextField()
+    description = models.TextField()
+    create_date = models.DateField()
+    created_by = models.TextField()
```
Ferdi: reference to some "User" table in Django ?

ShawnMilo:
We discussed (elsewhere, so it should be repeated here anyway) the possibility of subclassing `models.Model` and adding some features our models will use, including datetime fields for date added and date last modified, using Django's `auto_now` and `auto_now_add` [kwargs](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.DateField.auto_now).

Also, the topic of using UUIDs as primary keys was discussed favorably. Do we still want to consider that? There is a potential performance hit, but it is nice in other ways as well.

ShawnMilo:
Should the `learningobjects` directory be under `lore` instead of `lore/lore`?

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Course(models.Model):
+    repository = models.ForeignKey(Repository)
+    org = models.TextField()
+    course_number = models.IntegerField()
+    semester = models.TextField()
+    import_date = models.DateField()
+    imported_by = models.TextField()
+
+class LearningObject(models.Model):
```
carsongee: Why aren't we using MPTT to do materialized path for us?

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
+from django.db import models
+
+
+class Course(models.Model):
```
carsongee: I think it would be nice to create a base model as per a suggestion from @ShawnMilo that these all inherit from that add auditing fields, date_modified, date_created

lore/learningobjects/models.py
carsongee: yep, FK to `django.contrib.auth.models.User`

carsongee:
yeah, if we are doing these as sep django apps, this should be in `repo_root/learningobjects`, also, sorry for the redundant comment, I really need to remember to read everything before commenting

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+    url_path = models.TextField()
+    parent_id = models.ForeignKey("self")
+    copyright = models.TextField()
+    xa_nr_views = models.TextField()
+    xa_nr_attempts = models.IntegerField()
+    xa_avg_grade = models.FloatField()
+    xa_histogram_grade = models.FloatField()
+
+class LearningObjectType(models.Model):
+    learning_object = models.ForeignKey(LearningObject)
```
noisecapella: This foreign key should be removed since there's no reason for learning object types to have an id to a particular learning object

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+    course = models.ForeignKey(Course)
+    learning_object_type = models.ForeignKey(LearningObjectType)
+    uuid = models.TextField()
+    title = models.TextField()
+    description = models.TextField()
+    content_xml = models.TextField()
+    path_xml = models.TextField()
+    mpath = models.TextField()
+    url_path = models.TextField()
+    parent_id = models.ForeignKey("self")
```
noisecapella: We may want to rename this 'parent' for consistency, not 100% sure on naming convention however

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
+from django.db import models
+
+
+class Course(models.Model):
+    repository = models.ForeignKey(Repository)
```
noisecapella: We should change Repository to a string since they are defined after this point

lore/learningobjects/models.py
```diff
@@ -0,0 +1,37 @@
...
+    org = models.TextField()
+    course_number = models.IntegerField()
+    semester = models.TextField()
+    import_date = models.DateField()
+    imported_by = models.TextField()
+
+class LearningObject(models.Model):
+    course = models.ForeignKey(Course)
+    learning_object_type = models.ForeignKey(LearningObjectType)
```
noisecapella: We should change LearningObjectType to a string since it's defined below this point

lore/learningobjects/models.py
noisecapella: Also I think this needs to have null=True on it to allow a root node

lore/learningobjects/models.py
ghost: so this is not part of lore? if they are sep apps, then, being not part of lore, having them in repo_root makes sense. maybe we can change 'lore' to 'lorecore' or something less confusing in this architecture if learningobjects is a part of lore.

ShawnMilo:
This looks fine to me. I'm already using it successfully.

PR #13, opened 2015-05-14
Update to latest pytest-watch for watching in docker

carsongee:
Be explicit about None type for config file in settings
Correct minor Dockerfile permission issue
Sleep before running migrations to wait for the DB

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,21 +36,22 @@ overrides.
 Testing
 =======
 
-The project is set up with
-`tox<https://tox.readthedocs.org/en/latest/>`_ and
-`py.test<http://pytest.org/latest/>`_. It will run pylint, pep8, and
+The project is setup with
+`tox <https://tox.readthedocs.org/en/latest/>`_ and
+`py.test <http://pytest.org/latest/>`_. It will run pylint, pep8, and
 py.test tests with coverage. It will also generate an HTML coverage
 report. To run them all inside the docker image, run ``docker-compose
 run web tox``, or if you are running locally, after installing the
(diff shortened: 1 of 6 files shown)


carsongee:
@noisecapella mind reviewing?  I think you've already tried the docker change.

README.rst
```diff
@@ -44,13 +44,14 @@ report. To run them all inside the docker image, run ``docker-compose
...
-because the file events it uses are fired on the host OS, and not the
-docker OS. I have added an
-`issue<https://github.com/joeyespo/pytest-watch/issues/9>`_ to the
-`pytest-watch repo<https://github.com/joeyespo/pytest-watch>`_
+If you want test to run on file changes, the ``test_requirements.txt``
+adds pytest-watcher, which can be started with:
+``docker-compose run web ptw --poll``
+For additional options like having it say "passed"
+out loud, or sending desktop notifications for failures see the
+`README<https://github.com/joeyespo/pytest-watch/blob/master/README.md>`_.
```
noisecapella: These links don't seem to work for me. Could we convert them to something like this format? [MIT](http://mit.edu)

noisecapella:
The functionality works great for me, only nitpicking about link format

README.rst
carsongee: Good catch, they are restructured text links, and they need a space between the name and reference.  This should be fixed.  The markdown link format you have there won't work with RST unfortunately, and I always get them mixed up a bit.

carsongee:
Awesome, thanks for the review :do_not_litter:

Issue #4, opened 2015-05-11
Importer app


Ferdi:
Importer module is responsible for providing the APIs for :
- Import course content as tarball file that users can generate by a) exporting the course from studio or b) clicking “download” repository from github
- Parse each item of the course (chapter, sequential, vertical, problem etc.), extract the relevant metadata and the XML representation of the item and insert them in LORE database.
- Assign each item a universal unique identifier (UUID) if it does not have one. 

Uses the interface provided by #3

ShawnMilo:
How should we handle the same course being imported multiple times? How do we
detect it -- is it unique by repository + course number? What if the "duplicate"
course has been modified?

If it is a re-import, do we version the imported Learning Modules, so
there are now multiple versions of those Learning Modules to choose from?
Do we mark older versions of the Learning Modules as deprecated and link
to the newer version?

Deleting or replacing Learning Modules seems like a
bad idea, because it could be destructive to users who have already selected
Learning Modules for export, or who have exported them and may wish to re-export
the same content.

ShawnMilo:
Courses have static content (JavaScript, CSS, images). Is there a benefit in
uploading this to a CDN or online storage and embedding links in the imported
Learning Objects? It seems that something like this is needed (URL embedding)
even if only to point to an edX instance, and since there are no guarantees
that a course's content will be live on a particular instance, a CDN seems
like a good alternative. If permissions are an issue, S3 and IS&T are options.

Ferdi:
org + coursenumber + semester should uniquely identify the course.

carsongee:
I think we will need to upload the static assets to be able to export problems, but we should use django storage to do that so we can swap out the backend at will

ShawnMilo:
@carsongee, that implies that we need to parse the XML to determine which Learning Objects consume which static content. Looking at a course, it seems that there's no other way to determine which static file belongs to which Learning Object; multiple items could refer to the same resource in their XML, and due to the hierarchy, there will be overlap (if a problem uses a file, then by definition so will the enclosing chapter). So we need a  table with a foreign key to Learning Object for storing metadata for the static content.

carsongee:
Agreed, static files are tricky, and I'm not sure how completely we want to deal with them at first.  We can always just store everything for a course in a single folder/container, and then parse the exported XML files for what static content it needs on export which is how edX does it using their "magic" `/static/file_name.jpg` url expansion to find the correct files.  If we do that we don't have to worry about storing anything other than the files.

[...]

PR #19, opened 2015-05-15
merges migrations and first API functions

ShawnMilo:
Adds the migrations file for `learningobjects`, and the first API functions and tests.

### DIFF
--- a/apt.txt
+++ b/apt.txt
@@ -1,4 +1,7 @@
+git
 libpq-dev
+libxml2-dev
+libxslt1-dev
 python-pip
 python-dev
 python3-dev
--- a/apt.txt
+++ b/apt.txt
@@ -5,3 +5,4 @@ libxslt1-dev
 python-pip
 python-dev
 python3-dev
+git-core
--- a/apt.txt
+++ b/apt.txt
@@ -5,4 +5,3 @@ libxslt1-dev
 python-pip
 python-dev
 python3-dev
-git-core
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -1,0 +1,100 @@
+"""
+Import OLX data into LORE.
+"""
+
+# stdlib
+from glob import glob
+from shutil import rmtree
+from tempfile import mkdtemp
+from os.path import join
+import tarfile
+import zipfile
+
+# PyPi
+
+from xbundle import XBundle, DESCRIPTOR_TAGS
+from lxml import etree
+
+# MITODL
+# commented out for now, until it's merged
+# from lore.learningobjects.models import Course, LearningObject
+
+def import_course_from_file(filename):
+    """
+    Import OLX from .zip or tar.gz.
+    """
+    tempdir = mkdtemp()
+    if filename.endswith(".tar.gz"):
+        course = tarfile.open(filename, "r")
+    elif filename.endswith(".zip"):
+        course = zipfile.ZipFile(filename, "r")
+    else:
+        raise ValueError("Unexpected file type (want tarball or zip).")
+    course.extractall(tempdir)
+    dirs = glob(join(tempdir, "*"))
+    assert len(dirs) == 1
+    import_course_from_path(dirs[0])
+    rmtree(tempdir)
+
+def import_course_from_path(path):
+    """
+    Import course from an OLX directory.
+    """
+    bundle = XBundle()
+    bundle.import_from_directory(path)
+    return import_course(bundle)
+
+
+def import_course(bundle):
+    """
+    Import a course from an XBundle object.
+    """
+    course = Course()
+    course.course_number = bundle.course.attrib.get("course", "MISSING")
+    course.save()
+    import_children(course, bundle.course, None)
+
+
+def import_children(course, element, parent):
+    """
+    Create LearningObject instances for each element
+    of an XML tree.
+    """
+    lox = LearningObject()
+    lox.course = course
+    lox.display_name = element.attrib.get("display_name", "MISSING")
+    lox.tag = element.tag
+    lox.xml = etree.tostring(element)
+    if parent is not None:
+        lox.parent = parent
+    else:
+        lox.display_name = "/".join([
+            element.attrib.get(x, "MISSING")
+            for x in ("org", "course", "semester")
+        ])
+    lox.save()
+    for child in element.getchildren():
+        if child.tag in DESCRIPTOR_TAGS:
+            import_children(course, child, lox)
+
+# Dummy placeholders.
+
+class FakeModel(object):
+    """
+    Mock
+    """
+    def save(self):
+        return True
+
+class Course(FakeModel):
+    """
+    dummy
+    """
+    pass
+
+class LearningObject(FakeModel):
+    """
+    dummy
+    """
+    pass
+
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -2,26 +2,29 @@
 Import OLX data into LORE.
 """
 
-# stdlib
 from glob import glob
+from datetime import datetime
 from shutil import rmtree
 from tempfile import mkdtemp
 from os.path import join
 import tarfile
 import zipfile
 
-# PyPi
-
 from xbundle import XBundle, DESCRIPTOR_TAGS
 from lxml import etree
 
-# MITODL
-# commented out for now, until it's merged
-# from lore.learningobjects.models import Course, LearningObject
+from learningobjects.api import create_course
 
-def import_course_from_file(filename):
+
+def import_course_from_file(filename, user_id):
     """
     Import OLX from .zip or tar.gz.
+
+    Args:
+        filename (unicode): Path to archive file (zip or .tar.gz)
+    Raises:
+        ValueError: Unable to find single path inside archive file.
+    Returns: None
     """
     tempdir = mkdtemp()
     if filename.endswith(".tar.gz"):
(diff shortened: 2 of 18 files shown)


ShawnMilo:
This is to support #4. 

To import a course, we needed the migrations for the `learningobjects` app and the function to create a course.

ShawnMilo:
Re-pushed with `learningobjects` added to INSTALLED_APPS. That's what caused it to fail Travis last time.

learningobjects/api.py
```diff
@@ -0,0 +1,39 @@
...
+    # TODO: Determine rules for accepting/requiring a repository.
+    kwargs = {
+        "org": org, "course_number": course_number, "semester": semester,
+        "import_date": datetime.now(), 'imported_by_id': user_id,
+        "repository_id": get_temp_repository(user_id).id
+    }
+    courses = Course.objects.filter(**kwargs)
+    if courses.count() > 0:
+        return None, "course already exists"
+    return Course.objects.create(**kwargs), ""
```
noisecapella: Should we be returning a dictionary here or a Django model? It seems like the Django model would break the abstraction layer since now the caller can access it directly. I don't have a strong opinion here, a dictionary may be more work than it's worth

ShawnMilo:
Github is confused; `learningobjects` _is_ in `INSTALLED_APPS` now, and github knows it, but the build still failed again.

Disregard; `__init__.py` was missing from `learningobjects`. Re-pushed latest branch which is known to have tests run and pass locally.

And now tox is failing, probably because of a dependency I added. @carsongee or @bdero: Please take a look when you have some time. Thanks.

learningobjects/api.py
ShawnMilo: No, I don't get that crazy. I think it's not worth the trouble to do that. I'm not saying never use models; just not to use the ORM in most cases.

importer/api/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+# stdlib
+from glob import glob
+from shutil import rmtree
+from tempfile import mkdtemp
+from os.path import join
+import tarfile
+import zipfile
+
+# PyPi
```
carsongee: I think it is safe to leave these comments out, but it is worth noting that we should follow the three import group convention, of core, third-party and internal app

ShawnMilo: No problem. Personal habit, easy to break.

importer/api/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+from xbundle import XBundle, DESCRIPTOR_TAGS
+from lxml import etree
+
+# MITODL
+# commented out for now, until it's merged
+# from lore.learningobjects.models import Course, LearningObject
+
+def import_course_from_file(filename):
```
carsongee: This is assuming you are getting a file name, I don't think that is a safe assumption, I think it would probably be better to assume this is a file like object so we can use something like django storage or other file abstraction, after it is unzipped, you can go back to using regular filename/path semantics

importer/api/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    tempdir = mkdtemp()
+    if filename.endswith(".tar.gz"):
+        course = tarfile.open(filename, "r")
+    elif filename.endswith(".zip"):
+        course = zipfile.ZipFile(filename, "r")
+    else:
+        raise ValueError("Unexpected file type (want tarball or zip).")
+    course.extractall(tempdir)
+    dirs = glob(join(tempdir, "*"))
+    assert len(dirs) == 1
```
carsongee: no asserts in prod code (ok in tests), as they don't always run depending on how the python interpreter is invoked

importer/api/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+from xbundle import XBundle, DESCRIPTOR_TAGS
+from lxml import etree
+
+# MITODL
+# commented out for now, until it's merged
+# from lore.learningobjects.models import Course, LearningObject
+
+def import_course_from_file(filename):
+    """
+    Import OLX from .zip or tar.gz.
```
carsongee, line 21: full doc strings with expected types and return values would be good.
edit, and expected exceptions like the ValueError you possibly raise

importer/api/__init__.py
ShawnMilo: This is because we will be getting a filename when the course is uploaded as a tarball. If we're getting it from elsewhere we'll still have to save it as a file before opening it, because that's how `zipfile` and `tarfile` work.

importer/api/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    return import_course(bundle)
+
+
+def import_course(bundle):
+    """
+    Import a course from an XBundle object.
+    """
+    course = Course()
+    course.course_number = bundle.course.attrib.get("course", "MISSING")
+    course.save()
```
carsongee: This means we are overwriting courses with the same ID right? I thought we decided this would be one time only, so we should raise if it already exists I think

importer/api/__init__.py
ShawnMilo: Okay, will fix.

importer/tests/test_import.py
```diff
@@ -0,0 +1,56 @@
...
+# local
+from importer.api import import_course_from_file
+
+# TODO: Also test this course:
+# https://github.com/pmitros/edX-Insider/archive/master.zip
+def get_course_zip():
+    """
+    Get the path to the demo course.
+    """
+    url = "https://github.com/edx/edx-demo-course/archive/master.zip"
```
carsongee, line 20: Please don't put external dependencies in tests, it makes it really hard to work offline, makes the tests slower, and means that our tests are dependent on github being available.  I would rather add this as a git submodule to the repo, or just straight drop it into the repo.

importer/tests/test_import.py
```diff
@@ -0,0 +1,56 @@
...
+class TestImportDemo(TestCase):
+    """
+    Test import functionality.
+    """
+    def setUp(self):
+        """
+        Make sure we have a local copy of the edX Demo Course
+        for testing.
+        """
+        self.course_zip = get_course_zip()
```
carsongee, line 52: please always add super to setUp and tearDown

learningobjects/api.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Add a course to the database.
+    """
+    # TODO: Determine rules for accepting/requiring a repository.
+    kwargs = {
+        "org": org, "course_number": course_number, "semester": semester,
+        "import_date": datetime.now(), 'imported_by_id': user_id,
+        "repository_id": get_temp_repository(user_id).id
+    }
+    courses = Course.objects.filter(**kwargs)
```
carsongee, line 30: this isn't thread safe, or transaction safe. Please use get_or_create and wrap it in a transaction

importer/api/__init__.py
ShawnMilo: Yeah, sorry -- this is leftover code from when I was writing this functionality before I had the `learningmodules` branch available. If you look at `create_course` in `learningobjects/api.py` you'll see this will be a non-issue.

I'll clean it up.

learningobjects/api.py
carsongee, line 30: That also implies the model should be setup with the proper unique fields so that get_or_create will work properly

learningobjects/api.py
```diff
@@ -0,0 +1,39 @@
...
+    if courses.count() > 0:
+        return None, "course already exists"
+    return Course.objects.create(**kwargs), ""
+
+
+def get_temp_repository(user_id):
```
carsongee, line 35: This is a really strange thing to have, to me, at least make it deprecated, and possibly an _ since I don't think we want this to be part of our public api

learningobjects/migrations/0001_initial.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+from django.conf import settings
+
+# auto-generated code, so shut up
+# pylint: disable=line-too-long,invalid-name,missing-docstring
```
carsongee: instead of adding this, I believe I suggested you fix it in the pylintrc, that way it will work for all migration files

learningobjects/tests/test_api.py
```diff
@@ -0,0 +1,62 @@
...
+
+class TestCreateCourse(TestCase):
+    """
+    Test course creation API.
+    """
+    def setUp(self):
+        """
+        Set some test data
+        """
+        self.user, _ = User.objects.get_or_create(username="test")
```
carsongee, line 24: again, need super

importer/tests/test_import.py
ShawnMilo, line 20: So, add a 12Mb edx-demo-course zip file to this repository?

learningobjects/tests/test_api.py
```diff
@@ -0,0 +1,62 @@
...
+            'course_number': '42',
+            'semester': 'Febtober',
+            'user_id': self.user.id,
+        }
+
+    def tearDown(self):
+        """
+        Clean up test data.
+        """
+        kwargs = self.kwargs
```
carsongee, line 37: super

learningobjects/api.py
carsongee: Ok, so we need to decide on this one way or the other... I thought @ShawnMilo were specifically recommending  not using models, but several of the use cases you suggested for the API approach could happen if we are returning models and caller functions could/will become dependent on model coming back and the fields and methods therein. Sounds like we need to talk

learningobjects/api.py
ShawnMilo, line 35: Yeah, it's just there to get things moving, since this part had to be developed in isolation. I don't know what you mean by "make it deprecated," though.

learningobjects/api.py
ShawnMilo, line 30: I'll do something to make it safer, but I don't want to use `get_or_create` because it shouldn't do anything if the course already exists.

importer/tests/test_import.py
carsongee, line 20: or a git submodule, but then we all have to do recursive clones.  I believe I mentioned it on the issue, but I think we should actually just create a small course for testing, one that exhibits bad behaviors we have found.  The DevOps course would be perfect for this and is much smaller (1.1M), it just needs to have some sanitizing of it's policy file before becoming open.  In fact, for this test, we really should just be having a minimal course, i.e. a single course.xml file in a tar.gz or zip file, since we should assume (or create the tests to do so, since we know it isn't well tested) the piece that xbundle does is well tested, and we just need to test our code here.

carsongee, line 20: It should probably also have a small file in the /static folder as well that we reference to verify we can do static assets eventually

ShawnMilo, line 20: Okay. I did see that comment, but I'm not yet comfortable making my own course by hand. Whom should I pester to sanitize and publish the DevOps course?

carsongee, line 20: We could also just use the main one edX uses: https://github.com/edx/edx-platform/tree/master/common/test/data/two_toys and copy that in here.

importer/api/__init__.py
carsongee: Makes sense, we'll have to have that probably in the form handler/view then.

learningobjects/api.py
ShawnMilo: Yeah, my stance is evidently misunderstood. I never advocated never handling models. I simply meant that functions like `create_course` should exist and be used rather than having the same logic simply in a view or whatever. Because it could change, and it would certainly end up being non-DRY.

learningobjects/api.py
carsongee, line 30: But you are doing the get with your filter here, you can always ignore the object created if created is false, i.e.:
`course, created = ...get_or_create(...)`
`if created: return object; else raise Course.ValidationError`
I also think you should raise instead of tuple as was discussed somewhat in chat

apt.txt
```diff
@@ -1,4 +1,7 @@
+git
```
noisecapella, line 1: This is nitpicking but git-core is included a few lines down. Not a problem for me though

learningobjects/api.py
carsongee, line 30: Sorry, I should have helped you out on what I meant by making it safe: just import `from django.db import transaction`
and add `@transaction.atomic` as a decorator

apt.txt
ShawnMilo, line 1: Nits are good. It wasn't there when I added it. Also, `git-core` is obsolete.

```
~ 15:07 0 job(s)  
shawn@julius    
 $ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"

~ 15:07 0 job(s)  
shawn@julius    
 $ apt-cache search git-core
devscripts - scripts to make the life of a Debian Package maintainer easier
git - fast, scalable, distributed revision control system
git-core - fast, scalable, distributed revision control system (obsolete)
```

learningobjects/api.py
carsongee, line 35: Just add to the docstring, normally we would use: http://sphinx-doc.org/markup/para.html#directive-deprecated, but we don't really have an alternative...

apt.txt
carsongee, line 1: that is my bad, I added git-core in there, feel free to swap it

ShawnMilo:
I have pushed an update, but don't worry about reviewing it right now; I still have more issues to address.

apt.txt
ShawnMilo, line 1: No worries, already removed. That _was_ annoying when `apt-get install git` didn't work and you had to train yourself to type `git-core`. I'm glad they "fixed" it, but it's still annoying. Like on Ubuntu, there's already a dock applet nobody uses called `docker`, so you have to install `docker.io` from the repo. Of course I use the official repo, that's `lxc-docker`. This is why we can't have nice things.

importer/tests/test_import.py
ShawnMilo, line 20: Added two_toys to the repo, removed the `requests` dependency from the tests.

ShawnMilo:
Superseded by PR #21. Closing.

PR #21, opened 2015-05-18
Basic import functionality

ShawnMilo:
The code behind this demo: https://youtu.be/tvwzq3e06k0


learningobjects/tests/test_api.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+from django.test.testcases import TestCase
+
+from django.contrib.auth.models import User
+
+from learningobjects.models import Course
+from learningobjects.api import create_course
+
+# pylint: disable=no-member
```
carsongee: These are in almost all of these, what is triggering these?  If it is super common we should just add it to the pylintrc, but it would be nice to have a description as to why, i.e. `# matplotlib dynamically adds members to its imports, so disabling`.  Also, these are covering the whole file, is there a way to just disable a line item?

ShawnMilo: It happens with everything Django -- no model has member `objects`, nor `get_or_create` or `create`.

Also, it happens with `lxml.etree`.

I've kept them local when possible, but in files where models are used from multiple places I put them at the top. My logic is that if something actually _doesn't_ have a member the code will be broken and tests will fail anyway, so the warning is pretty worthless.

carsongee: We can try and fix them with the RC, something like:

``` ini
[TYPECHECK]
generated-members=objects,_meta,id,etree
```

or whatever else you need.  It may not be worth it in the long run if we end up with a ton of these, but would help keep the code cleaner. I'm also mostly ok with just nuking that message altogether if it is too tedious to keep the list up

ShawnMilo: I like the idea of disabling them globally. If we have `thing.whatever()` in our code, and `thing` doesn't have method `whatever`, I'm pretty sure we'll find out pretty quickly.

I think it has to do with metaclass programming. If you parse the Python they actually _don't_ have those methods, but at runtime they do. I know that's how Django models work; I'm just guessing on lxml.

carsongee:
Also, after we get tests passing, this is going to need a squash

carsongee:
Just to cross post since this PR needs it too:
I mentioned it in chat last week, but the migration fix to pep8 is:

``` ini
pep8ignore =
    */migrations/* ALL
```

in pytest.ini

Ferdi:
@ShawnMilo can you move views/templates/url in the UI app ?

ShawnMilo:
@carsongee: Would you be willing to help with the squash? I've squashed it before, but the past couple of iterations the rebase has gotten really messy and I gave up. Or I could always just check out master and copy everything into the folder again.

ShawnMilo:
@Ferdi I will, but since the listing page (Pivotal task 24) is already underway and is based on this branch, I'll do it after they're merged together.

carsongee:
Still working on the weird pylint violations on migrations, but there are still lots of little things to cleanup here.  I am happy to try and do the squash myself, can you post the final commit message you want here?

importer/templates/importer/upload.html
```diff
@@ -0,0 +1,6 @@
+<h1>upload here</h1>
```
carsongee: templates should always extend base.html. https://code.djangoproject.com/wiki/DosAndDontsForApplicationWriters

ShawnMilo: No arguments there. There's no base.html yet and there's no HTML header or anything in these templates that would conflict with it once it exists. I'll create an empty base.html and extend it just so it's there.

Ferdi: Can we create one real fast ?

ShawnMilo: Yes, as I said above I'll add one, then push again.

pylintrc
```diff
@@ -1,3 +1,4 @@
 [BASIC]
 # Allow django's urlpatterns, and our log preference
 const-rgx = (([A-Z_][A-Z0-9_]*)|(__.*__)|log|urlpatterns)$
+ignore=migrations
```
carsongee: so this kind of sucks, but this appears to be a limitation of pylint, if you run `pylint learningobjects/migrations/0001_initial.py --ignore=migrations` or `pylint learningobjects/migrations/0001_initial.py --ignore=0001_initial.py` then it still checks the file.  I think we just need to add `#pylint: skip-file` unfortunately.  The pep8 rule will work, and I'll file an issue to see if we can't fix it either in pylint or at least pytest-pylint.

carsongee: https://github.com/carsongee/pytest-pylint/issues/2

importer/templates/importer/upload.html
carsongee: You can just use the one I put in the cookiecutter that you reviewed: https://github.com/mitodl/cookiecutter-djangoapp/blob/master/%7B%7B%20cookiecutter.app_name%20%7D%7D/%7B%7B%20cookiecutter.app_name%20%7D%7D/templates/base.html

pylintrc
ShawnMilo: Okay, so adding `#pylint: skip-file` to the migration file fixes it. So I'll add that to all migrations for now.

requirements.txt
```diff
@@ -6,3 +6,4 @@ uwsgi==2.0.10
 psycopg2==2.6
 tox==1.9.2
 static3==0.5.1
+git+https://github.com/mitodl/xbundle.git
```
carsongee: Looks like we are still getting the beautiful soup stuff installed, maybe try changing this to:
`git+https://github.com/mitodl/xbundle.git@d6eb740e023f52f726b8830990ac108df547b0a8#egg=xbundle`
to make sure nothing is cached

test_requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov
 pytest-django
 git+https://github.com/joeyespo/pytest-watch@cd236e6507e6463a9949d2515391ebe68599ad55#egg=pytest-watch-poll
 mock
+requests
```
carsongee: this doesn't need to be here anymore right?

ui/templates/importer/upload.html
```diff
@@ -0,0 +1,9 @@
+{% extends "lore/base.html" %}
```
carsongee: I think just base.html should work, and would be preferred, but happy to be wrong if that is wrong...

test_requirements.txt
ShawnMilo: Correct. I'll take it out in the following branch.

ui/templates/importer/upload.html
ShawnMilo: I removed `lore/` and it broke it. I think it's because they're under different subfolders in `ui/templates`.

Ferdi: There is no need for subfolders under /ui/templates

ShawnMilo:
I fixed the `requests` import. Is there anything else keeping this from being merged? 

It's passing the Python 2 tests, and the Python3 tests are only failing because the [xbundle on PyPi](https://pypi.python.org/pypi/xbundle/0.2.0) still has `BeautifulSoup` as a requirement, and `BeautifulSoup` won't install on Python 3. 

I know once that's fixed it'll still break until Amir finishes [fixing xbundle](https://github.com/mitodl/xbundle/issues/30), but I don't think that should hold this up.

carsongee:
That was one of my suggested fixes, point it at the github hash, it may have gotten lost in the diffs, but just do: `git+https://github.com/mitodl/xbundle.git@d6eb740e023f52f726b8830990ac108df547b0a8#egg=xbundle` instead of `xbundle=0.2.0`and we should pass, we will fix it once we get 0.3.0 out there

carsongee:
Oh, also the no-member thing, just kill the whole check in pylintrc and move it out of the code files

carsongee:
I think there is probably a coverage issue in `importer/api/__init__.py` too, but that will be easier to checkout once the build passes so that coveralls runs its report.

requirements.txt
```diff
@@ -6,3 +6,4 @@ uwsgi==2.0.10
 psycopg2==2.6
 tox==1.9.2
 static3==0.5.1
+-e git+https://github.com/mitodl/xbundle.git@7799a0957c502103fb030fc0b16c64204f551c75#egg=xbundle
```
carsongee: Sorry for the trifle, you shouldn't need to use -e here, i.e. https://github.com/mitodl/lore/blob/master/test_requirements.txt#L6

ShawnMilo: Thanks. Tested, confirmed, and fixed.

carsongee:
Hurray :fireworks:, the tests pass.  It looks like there is quite a bit of coverage missing on newly added code, like the form and the views.  It also looks like you aren't testing exceptions in the api module which is probably the piece we care most about since it will likely live longer than the rest.  Could you add tests for those at least, and hopefully for the other pieces as well, but they are closer to "nice to haves".

importer/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    Args:
+        course (learningobjects.Course): Course
+        element (lxml.etree): XML element within xbundle
+        parent (learningobjects.LearningObject): parent LearningObject
+    """
+    # pylint: disable=no-member
+    # for tostring() below.
+
+    lox = create_lox(
+        course=course, parent=parent, lox_type=element.tag,
```
carsongee: If we have an mpath tree, which it looks like we do, based on the model (https://github.com/mitodl/lore/pull/21/files#diff-69e311fb6dd79b2f1d831c899b5113c4L33), we should be setting the path here as well.  There are also several other fields we aren't populating like the xml path. We should either use them, or nuke them from the models I think.

importer/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    elif filename.endswith(".zip"):
+        course = zipfile.ZipFile(filename, "r")
+    else:
+        raise ValueError("Unexpected file type (want tarball or zip).")
+    course.extractall(tempdir)
+    dirs = glob(join(tempdir, "*"))
+    if len(dirs) != 1:
+        raise ValueError("Unable to get course directory.")
+    import_course_from_path(dirs[0], user_id)
+    rmtree(tempdir)
```
carsongee, line 43: You nuke the tempdir, but this should also nuke the original file as well so we don't end up with a cluttered system.  Especially since we should be able to reconstruct the tar ball from the DB entries we are making.

importer/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        filename (unicode): Path to archive file (zip or .tar.gz)
+    Raises:
+        ValueError: Unable to find single path inside archive file.
+    Returns: None
+    """
+    tempdir = mkdtemp()
+    if filename.endswith(".tar.gz"):
+        course = tarfile.open(filename, "r")
+    elif filename.endswith(".zip"):
+        course = zipfile.ZipFile(filename, "r")
```
carsongee: Last I checked, these are very unsafe operations as it can overwrite our software and other users data by clever pathing. Here is an example implementation of a safe unzipping: https://github.com/edx/edx-platform/blob/c08ccee45a5722755e6c47efda487dcd509001d0/openedx/core/lib/extract_tar.py or from SO: http://stackoverflow.com/questions/10060069/safely-extract-zip-or-tar-using-python

requirements.txt
```diff
@@ -6,3 +6,4 @@ uwsgi==2.0.10
 psycopg2==2.6
 tox==1.9.2
 static3==0.5.1
+git+https://github.com/mitodl/xbundle.git@7799a0957c502103fb030fc0b16c64204f551c75#egg=xbundle
```
carsongee, line 9: I just thought of this, but I think we should add commented sections for each app so we can pull apart deps.  Both here, and in apt.txt as well.  So far everything in there has been core, but this and the xml libs in apt, would be either import app deps with a comment stating such.

importer/api/__init__.py
ShawnMilo: Thanks. I checked out the StackOverflow page and the [code snippet](https://github.com/edx/edx-platform/blob/c08ccee45a5722755e6c47efda487dcd509001d0/openedx/core/lib/extract_tar.py). I'm assuming this isn't part of a "pip-installable" package, so I should just copy/paste the whole file into LORE. Or is there some kind of `mitodl/utils` package floating around with bits and bobs like this? If not, should we create one?

learningobjects/tests/test_api.py
carsongee: I'm fine with that, then just add

``` ini
[MESSAGES CONTROL]

disable = no-member
```

to the pylintrc file and remove all the per file ones.

importer/api/__init__.py
carsongee: https://pypi.python.org/pypi/Archive/0.3 suggests it can do that.

learningobjects/models.py
```diff
@@ -32,12 +32,12 @@ class LearningObject(models.Model):
     path_xml = models.TextField()
     mpath = models.TextField()
```
noisecapella: This may be the wrong place to mention this but are we renaming this to materialized_path? Or using MPTT per @carsongee suggestion? I have no opinion either way, just making sure it wasn't accidentally overlooked

importer/api/__init__.py
ShawnMilo: It does, and works well. However, not in Python 3 because it uses `basestring` to check if the filename passed is a string. It should be a simple fix and I'd like to fork and do a PR to the original package author. Should I fork it under `mitodl`, or just my own account?

carsongee: mitodl is what I've been doing so far, then nuke it and repoint requirements after it merges upstream.

ShawnMilo: This turned out to be unnecessary; the latest commit on github is Python 3 compatible.

learningobjects/models.py
ShawnMilo: Last I heard, mptt was not considered necessary, since the individual `LearningObject`'s relationship with the other `LearningObject`s is unimportant once we're past the import stage.

I don't remember hearing about renaming the field, but it's simple, so someone let me know if I should do that.

learningobjects/forms.py
```diff
@@ -0,0 +1,20 @@
+"""
+Forms for LearninObjects
```
noisecapella: Typo

learningobjects/models.py
Ferdi: I think it's highly likely that we might need MPTT, but I don't know yet. It should not block this PR though.  The same for renaming the field. I think I suggested that , but not a blocker.

carsongee: If so, I think we should include it here.  I imagine the migration to using MPTT is going to be pretty nasty, but just a guess, and we can always just reset migrations if it proves to problematic.

learningobjects/tests/test_api.py
```diff
@@ -0,0 +1,59 @@
...
+class TestCreateCourse(TestCase):
+    """
+    Test course creation API.
+    """
+    def setUp(self):
+        """
+        Set some test data
+        """
+        super(TestCreateCourse, self).setUp()
+        self.user, _ = User.objects.get_or_create(username="test")
```
noisecapella: The tuple unpacking here should be removed (forgive me if you're already aware of this or see this many times in this code review, just being cautious)

noisecapella: Whoops, nevermind

importer/urls.py
```diff
@@ -0,0 +1,12 @@
+"""
+URLs for the importer app.
+"""
+from django.conf.urls import url, patterns
```
carsongee: Patterns is no longer recommended: https://docs.djangoproject.com/en/1.8/releases/1.8/#django-conf-urls-patterns

learningobjects/models.py
pdpinch: To clarify @ShawnMilo rename the field but we'll deal with MPTT later.

learningresources/models.py
```diff
@@ -1,30 +1,30 @@
...
 
-class LearningObject(models.Model):
+class LearningResource(models.Model):
     """
     The units that compose an edX course:
     chapter, sequential, vertical, problem, video, html, etc.
     """
     course = models.ForeignKey(Course)
-    learning_object_type = models.ForeignKey('LearningObjectType')
+    learning_object_type = models.ForeignKey('LearningResourceType')
```
noisecapella: Should learning_object_type be renamed to learning_resource_type here?

ShawnMilo: Yep, good catch. Thanks.

learningresources/models.py
```diff
@@ -32,17 +32,17 @@ class LearningObject(models.Model):
     path_xml = models.TextField()
     mpath = models.TextField()
```
ShawnMilo: There was a comment about renaming this, but the conversation was lost when this file changed. To what are we renaming the `mpath` field?

pdpinch: From @Ferdi 

> maybe rename to "materialized_path" ?

importer/api/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Create LearningResource instances for each element
+    of an XML tree.
+
+    Args:
+        course (learningresources.Course): Course
+        element (lxml.etree): XML element within xbundle
+        parent (learningresources.LearningResource): parent LearningResource
+    """
+    lox = create_lox(
```
carsongee: I think we probably want to rename this too, and all the other lox things

learningresources/api.py
```diff
@@ -0,0 +1,103 @@
...
+    Returns:
+        lox (learningresources.LearningResource): new LearningResource
+    """
+
+    lox = LearningResource(
+        course=course,
+        parent_id=parent.id if parent is not None else None,
+        learning_resource_type_id=type_id_by_name(lox_type),
+        title=title,
+        content_xml=content_xml,
```
carsongee: still no path set here.

learningresources/api.py
```diff
@@ -0,0 +1,103 @@
...
+    a single query.
+
+    Args:
+        name (unicode): LearningResourceType.name
+    Returns:
+        type_id (int): pk of learningresources.LearningResourceType
+    """
+    if name in TYPE_LOOKUP:
+        return TYPE_LOOKUP[name]
+    obj, _ = LearningResourceType.objects.get_or_create(name=name.lower())
```
carsongee: this should be wrapped in a transaction

learningresources/api.py
```diff
@@ -0,0 +1,103 @@
...
+        "repository_id": get_temp_repository(user_id).id
+    }
+    if Course.objects.filter(**unique).exists():
+        raise ValueError("Duplicate course")
+    kwargs = {
+        "org": org, "course_number": course_number, "semester": semester,
+        'imported_by_id': user_id,
+        "repository_id": get_temp_repository(user_id).id
+    }
+    course, _ = Course.objects.get_or_create(**kwargs)
```
carsongee: still needs to be wrapped in transaction

learningresources/api.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Args:
+        user_id (int): primary key of user creating the course
+    Returns:
+        repository (learningresources.Repository): repository
+    """
+    repos = Repository.objects.all()
+    if repos.count() > 0:
+        return repos[0]
+    return Repository.objects.create(
```
carsongee: probably should be a transaction, but since this will likely go away, not a big deal.

test_requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov
 pytest-django
 git+https://github.com/joeyespo/pytest-watch@cd236e6507e6463a9949d2515391ebe68599ad55#egg=pytest-watch-poll
 mock
+ipython
```
carsongee, line 8: yes, thank you. Been meaning to add this

ShawnMilo, line 8: :+1:

importer/views.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+
+def upload(request):
+    """
+    Upload a OLX archive.
+    """
+    form = UploadForm()
+    if request.user.is_anonymous():
+        request.user, _ = User.objects.get_or_create(username="dirty_hack")
```
carsongee: I know this is just getting stuff in here, but as soon as this merges to master, anyone on the Internet will be able to upload large tar balls to our app.  Maybe at least make a note in the doc string that this view should promptly be protected?

carsongee:
Looks like you need to rebase this on master as well, otherwise looking good after comments.

importer/views.py
ShawnMilo: I didn't expect that this was going live anywhere before the authentication kicked in. Is it?

I added `TODO` comments, but then it fails tox, so I took them out.

Also, as for large tarballs, I assumed that the server (nginx?) will have max sizes to protect us from people uploading huge files. I know we can set a max size in Django also, but then you have to maintain it in two places. So, just a note: The server software should have a setting to limit the size.

carsongee: Master is running live on Heroku: https://lore-ci.herokuapp.com.  Yeah, you can either keep the TODO and disable that pylint check, or put the comment in without it as is.  Heroku has a 30MiB file limit, which will of course be a problem pretty quickly, so I don't think we can run prod on it without switching to using something like S3 client side uploads (which we probably want to do anyway).

importer/api/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns: None
+    """
+    tempdir = mkdtemp()
+    try:
+        extract(path=filename, to_path=tempdir, method="safe")
+    except ArchiveException:
+        remove(filename)
+        raise ValueError("Invalid OLX archive, unable to extract.")
+    dirs = glob(join(tempdir, "*"))
+    if len(dirs) != 1:
```
carsongee, line 38: Looks like we still need a test case to exercise this.

importer/views.py
```diff
@@ -0,0 +1,43 @@
...
+    if request.method == "POST":
+        form = UploadForm(data=request.POST, files=request.FILES)
```
carsongee, line 37: Looks like we are missing a test case for this based on coverage, would be nice to have.

pdpinch:
If the missing coverage in https://github.com/mitodl/lore/pull/21/files#r30840968 and https://github.com/mitodl/lore/pull/21/files#r30841116 are the last issues, then I think we can merge this.

ShawnMilo:
I'm trying to run tests, but they're failing after merging master with `ImportError: No module named 'compressor'`.  @carsongee 

Disregard, fixed. Removed `compressor` from `INSTALLED_APPS`.

ShawnMilo:
Pushed with rebase.

learningresources/api.py
```diff
@@ -0,0 +1,113 @@
...
+        return repos[0]
+    with transaction.atomic():
+        return Repository.objects.create(
+            name="demo",
+            create_date=datetime.now(),
+            created_by_id=user_id,
+        )
+
+
+def create_resource(kwargs):
```
carsongee: nit: Why not have the parameters in the signature? It always seems clearer to me to have defined args, and I think that helps with IDEs and annotations

learningresources/api.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Create a learning resource.
+
+    Args:
+        course (learningresources.Course): course
+        parent (learningresources.LearningResource): parent LearningResource
+        resource_type (unicode): name of LearningResourceType
+        title (unicode): title of resource
+        content_xml (unicode): XML
+        mpath (unicode): materialized path
```
carsongee, line 80: nit: don't we want to rename these everywhere? We don't have a sealed API or anything

learningresources/models.py
```diff
@@ -1,48 +1,48 @@
...
     chapter, sequential, vertical, problem, video, html, etc.
     """
     course = models.ForeignKey(Course)
-    learning_object_type = models.ForeignKey('LearningObjectType')
+    learning_resource_type = models.ForeignKey('LearningResourceType')
     uuid = models.TextField()
     title = models.TextField()
     description = models.TextField()
     content_xml = models.TextField()
     path_xml = models.TextField()
```
carsongee: This is an unused field, do we have plans for it?  It looks like we are storing etree path in in mpath, so wouldn't that make this redundant anyway?

learningresources/api.py
ShawnMilo: It did originally, but then it broke the build; too many parameters.

learningresources/api.py
ShawnMilo, line 80: I did that intentionally to keep the line length down. I don't mind if you prefer to change it.

learningresources/api.py
carsongee: That is what I thought, I would just pylint comment disable it, or if we want to allow more for the whole project (no objections from me), just change the config

``` ini
[DESIGN]
# Maximum number of arguments for function / method
max-args=10
```

requirements.txt
carsongee, line 9: You don't have to fix this here, I'll fix this on a sep PR

learningresources/api.py
carsongee, line 80: for local representations, I'm fine with shortening, but generally for arguments to an API I think the full name is clearer in pydoc

carsongee:
the compressor problem is something due to a merge to master, looks like we need another rebase and for a clean image, `rm -r .tox; docker-compose build --no-cache`.  @pwilkins hit that same compressor thing.

ShawnMilo:
I already fixed the compressor issue; it was in INSTALLED_APPS erroneously.

carsongee:
That wasn't erroneous, http://django-compressor.readthedocs.org/en/latest/quickstart/#installation, it is our static asset pipeline and is required.

ShawnMilo:
I just merged `master` again. There were merge conflicts, tests are broken, and tox fails because the Docker image will no longer build. I can no longer work unless I undo the merge, but there have been multiple issues with merging `master` over the past week and I don't want to get further out of sync.

Ferdi:
I think you need some sleep first :-)
- Sent from my phone

ShawnMilo:
Authentication added. If tests pass, I think it's ready for merge to master (after another rebase).

learningresources/api.py
carsongee: Are you not going to fix this before merge?

ShawnMilo: Yeah, I will. Sorry, forgot about that one for a while.

learningresources/models.py
carsongee: I think this is my biggest concern before merging.   I know we are pre-creating fields before use to lay stuff down, but we need some design purpose for this that isn't duplicated in another field.

ShawnMilo: So, delete the field `path_xml`? It looks redundant, and we can always migrate new fields in later if necessary.

carsongee: Yep

ShawnMilo:
Recently updated:
- fix "kwargs" usage
- fix coverage
- remove xml_path field
- import `unicode_literals` in more files

ShawnMilo:
@carsongee Am I good to rebase for a merge?

carsongee:
Awesome on the coverage, everything looks great :honey_pot:. Just squash and I'll merge

ShawnMilo:
Squashed. Just awaiting Travis and it should be good.

carsongee:
Looks like maybe the zip file got messed up, tests failing

ShawnMilo:
Yeah, it's zero bytes for some reason. I'll fix.

importer/tests/test_import.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Test import functionality on an actual course. These tests should
+    be expanded as needed to test regressions and handle valid but
+    non-standard courses.
+    """
+    def setUp(self):
+        """
+        Return location of the local copy of the "two_toys"
+        course for testing.
+        for testing.
```
noisecapella, line 45: Looks like a typo, wouldn't hold up a merge for it though

tests/test_settings.py
```diff
@@ -1,6 +1,7 @@
 """
 Validate that our settings functions work, and we can create yaml files
 """
+
```
noisecapella, line 4: Should we add `from __future__ import unicode_literals` here?

ShawnMilo, line 4: I had it in there, but it broke the tests.

`AssertionError: ((u'Admins', ''),) is not false` in `TestSettings.test_admin_settings`. 

It passes in Python 3, but fails in Python 2. I decided to stop messing with it for the sake of time.

carsongee, line 4: Weird, modernize doesn't suggest that it is needed.  I can look into fixing that in another PR since I wrote the tests.  It looks like we are missing a bunch of absolute imports anyway.

carsongee:
Hitting the big red merge button

PR #31, opened 2015-05-19
Add requirejs and django-compressor for asset pipeline

### LINKED ISSUE #17: Create Static Asset pipeline with optional minification and JavaScript dependencies
As a user, I would like the fastest and leanest JavaScript and CSS files possible to make the user experience faster, and as a developer, I would like to use require.js style semantics for including javascript libraries into my templates.
https://www.pivotaltracker.com/story/show/94671190

carsongee:
Enable attempting to parse environment variables as python types
Closes #17
[Finished  #94671190]


bdero:
This pipeline looks really nice to work with. I reviewed the code and everything looks good to me, now I'm just pulling it down to try it out.

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ web:
   environment:
     PORT: 8070
     DATABASE_URL: postgres://postgres@db:5432/postgres
+    DEBUG:
```
bdero: This addition results in DEBUG being False for me in the settings by default. Should this just be True since I don't think we'll be using docker-compose in production?

carsongee: Cool, I set it to True in docker-compose.yml and set it to default to false in the settings.py file.

bdero:
:eggplant:

PR #32, opened 2015-05-19
Add optional CAS integration - Take two

### LINKED ISSUE #16: Authentication
- [x] Allow users to authenticate using MIT certificate authentication via CAS
- [x] MIT auth should be pluggable , easily enabled / disabled in any LORE instance.

carsongee:
Add ADMIN e-mail configuration
Closes #16
[finished #94639216]

We can wait until https://github.com/mingchen/django-cas-ng/pull/34 merges and point at their repo, or use this as is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@ RUN pip install pip --upgrade
 # Install project packages
 RUN pip install -r requirements.txt
 RUN pip install -r test_requirements.txt
+RUN pip3 install -r requirements.txt
+RUN pip3 install -r test_requirements.txt
 
 # Add, and run as, non-root user.
 RUN adduser --disabled-password --gecos "" mitodl
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -86,6 +86,8 @@ MIDDLEWARE_CLASSES = (
     'django.middleware.security.SecurityMiddleware',
 )
 
+AUTHENTICATION_BACKENDS = ('django.contrib.auth.backends.ModelBackend',)
+
 ROOT_URLCONF = 'lore.urls'
 
 TEMPLATES = [
@@ -155,3 +157,26 @@ STATICFILES_DIRS = (
 COMPRESS_PRECOMPILERS = (
     ('text/requirejs', 'requirejs.RequireJSCompiler'),
 )
+
+# e-mail configurable admins
+ADMIN_EMAIL = get_var('LORE_ADMIN_EMAIL', None)
+if ADMIN_EMAIL is not None:
+    ADMINS = (('Admins', ADMIN_EMAIL),)
+
+CAS_ENABLED = False
+
+# Optional CAS Integration
+if get_var('LORE_USE_CAS', None):
+    CAS_ENABLED = True
+    MIDDLEWARE_CLASSES += ('django_cas_ng.middleware.CASMiddleware',)
+    INSTALLED_APPS += ('django_cas_ng',)
+    AUTHENTICATION_BACKENDS += ('django_cas_ng.backends.CASBackend',)
+    LOGIN_URL = '/login/'
+    LOGIN_REDIRECT_URL = '/'
+    CAS_SERVER_URL = get_var(
+        'LORE_CAS_URL', 'https://auth-r.mitx.mit.edu/login'
+    )
+    CAS_EXTRA_LOGIN_PARAMS = {
+        'provider': 'touchstone',
+        'appname': "LORE"
+    }
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ static3==0.5.1
 django-compressor==1.5
 git+https://github.com/bpeschier/django-compressor-requirejs@889d5edc4f2acaa961c170084f1171c700649519#egg=githubdjango-compressor-requirejs-develop
 django-appconf==1.0.1
+git+https://github.com/mingchen/django-cas-ng@69463fe7c23f025be7165c2967f4ceb983d60472#egg=django-cas-ng
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -1,18 +1,34 @@
 """
 Validate that our settings functions work, and we can create yaml files
 """
+import imp
 import os
+import sys
 import tempfile
-import unittest
 
+from django.conf import settings
+from django.core import mail
+from django.test import TestCase
 import mock
 import yaml
 
 from lore.settings import load_fallback, get_var
 
 
-class TestSettings(unittest.TestCase):
+class TestSettings(TestCase):
     """Validate that settings work as expected."""
+
+    def reload_settings(self):
+        """
+        Reload settings module with cleanup to restore it.
+
+        Returns:
+            dict: dictionary of the newly reloaded settings ``vars``
+        """
+        imp.reload(sys.modules['lore.settings'])
+        # Restore settings to original settings after test
+        self.addCleanup(imp.reload, sys.modules['lore.settings'])
+        return vars(sys.modules['lore.settings'])
 
     def test_load_fallback(self):
         """Verify our YAML load works as expected."""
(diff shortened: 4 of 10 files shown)


Dockerfile
```diff
@@ -16,6 +16,8 @@ RUN pip install pip --upgrade
 # Install project packages
 RUN pip install -r requirements.txt
 RUN pip install -r test_requirements.txt
+RUN pip3 install -r requirements.txt
+RUN pip3 install -r test_requirements.txt
```
bdero, line 20: Are the python2 and python3 dependencies being roped in to allow for running the test suite against both at the same time?

lore/settings.py
```diff
@@ -139,3 +141,26 @@ def get_var(name, default):
...
+    AUTHENTICATION_BACKENDS += ('django_cas_ng.backends.CASBackend',)
+    LOGIN_URL = '/login/'
+    LOGIN_REDIRECT_URL = '/'
+    CAS_SERVER_URL = get_var(
+        'LORE_CAS_URL', 'https://auth-r.mitx.mit.edu/login'
+    )
+    CAS_EXTRA_LOGIN_PARAMS = {
+        'provider': 'touchstone',
+        'appname': "LORE"
+    }
```
bdero, line 182: Should these extra params also be somehow pulled using get_var?

requirements.txt
```diff
@@ -6,3 +6,4 @@ uwsgi==2.0.10
 psycopg2==2.6
 tox==1.9.2
 static3==0.5.1
+git+https://github.com/mitodl/django-cas-ng@feature/python3_compat#egg=django-cas-ng
```
bdero: I think it's cool doing it like this :+1:

tests/test_settings.py
```diff
@@ -41,3 +60,60 @@ def test_get_var(self):
...
+            'LORE_CAS_URL': 'foobar'
+        }, clear=True):
+            settings_vars = self.reload_settings()
+            self.assertTrue(
+                set(dict(
+                    CAS_ENABLED=True,
+                    LOGIN_URL='/login/',
+                    LOGIN_REDIRECT_URL='/',
+                    CAS_SERVER_URL='foobar'
+                )).issubset(set(settings_vars))
```
bdero, line 95: Cool use of sets!

tests/test_urls.py
```diff
@@ -0,0 +1,56 @@
...
+            ))
+
+    def test_cas_disable(self):
+        """Verify that when CAS is disabled, login is default"""
+        with self.settings(
+            CAS_ENABLED=False
+        ):
+            reload_url_conf()
+            response = self.client.get('/login', follow=True)
+            # pylint: disable=no-member
```
bdero: Can you place this pylint comment at the end of the relevant line so that it isn't applied to the method scope?

Dockerfile
carsongee, line 20: This was so I can runserver with python 3 for manual testing. Tox does everything on its own without it for 3.x

lore/settings.py
carsongee, line 182: They should always be static for LORE per the spec, but we could make them configurable if/when we need to.

bdero, line 182: Cool, I figured it wasn't really important right now

requirements.txt
carsongee: They merged it last night, so swapping it back to theirs and deleting our fork

tests/test_settings.py
carsongee, line 95: That is because they took away my beloved (but ambiguous)[assertDictContainsSubset](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertDictContainsSubset), but thanks

tests/test_urls.py
carsongee: I added it to pylintrc, since we will likely be doing a lot of .status_codes....though I think we may just disable the entire no-member check entirely since it is all over the place in Django

Ferdi: Can we move this to a separate app pls?  So that settings,  urls and test
are self contained within the app
- Sent from my phone

carsongee: No, these specifically are the tests of the project which requires settings, and urls in order to have anything run at all, it therefore makes sense to have tests of those settings and urls at that same level.

pwilkins:
@carsongee I can confirm that this now passes both py27 and py34 tests running in the Docker container.  Is that sufficient for merge?

carsongee:
Maybe make sure @bdero is good with the changes I made per his review?

pdpinch:
Resurrecting a thread that was lost in a recent update:

Ferdi wants the settings, urls and tests to be self-contained in its own (authentication?) app.

How can we do that?

https://github.com/mitodl/lore/pull/32/files#r30702401

Ferdi:
Sorry, I replied from my phone and the comment went somewhere inline.   What I'm asking is :  Can we move the cas authentication to a separate app, like mit_cas, if possible ? So that the only thing one does at the project level is to enable / disable the app. Have less "if" conditions

https://goo.gl/MlkrRU
https://goo.gl/AYCKWc

Can that app also tests its own settings? 
https://goo.gl/PNz6PD

carsongee:
This can potentially all be done, but the code here isn't actually an app, there is no logic except one if statement.  This just an integration of the builtin `django.contrib.auth` app and `django-cas-ng` app, so it didn't make any sense to me to make a third app on top of two other apps.  That said, this can technically be done, but it will probably take another day, and it won't get rid of any of the common tests for project level URLs or settings, since we still need to do integration testing of application specific URLs and settings at the project level to make sure everything is configuring properly with environment variables.  It will also likely require a lot of duplicative code in the app that is currently in settings.py.  It will likely also make it more difficult as an ops person to determine what settings are in use for any given installation of LORE.

Let me know if this seems worth it to have the extra app, and I will work towards it.

carsongee:
Ok, so technically two if conditions, and those still have to be somewhere, that is just shifting them into the app since you wanted the feature to be configurable on or off, there is going to have to be some if statements to enable or disable URLs and configure `django.contrib.auth`.

carsongee:
Would this also mean I should create an "ADMIN" app for setting which e-mail address error reports go to since that also has an if statement and a test?  That setting is unrelated to the CAS change, and I just snuck it in there since it was small and needed to be done

Ferdi:
Obviously, I'm not asking we duplicate code or do other hacks to achieve this. If there is a clean way, that would be really great.  
For example,  if Harvard had to use LORE, they would integrate with their own auth system. They should not need to have MIT specific code (https://goo.gl/PNz6PD) in their code base.  Or, if taxonomy app came up with its own settings, it would be great to test those settings within the app itself. 
Anything we can do toward this goal , would be awesome.

Dockerfile
bdero, line 20: Ah, that's awesome - tox really is great. This change makes sense to me.

carsongee:
That is not MIT specific code, that is an industry standard CAS protocol client http://en.wikipedia.org/wiki/Central_Authentication_Service used by thousands, nothing in this commit would pin them to any specifics at MIT whatsoever.  In fact, we use CAS to proxy to the slightly more specific implementation of MIT touchstone shibboleth as well as more standard OpenID implementations like Google and Yahoo, and in fact, you can use our CAS server to login to Harvard with a slight configuration change.  The code you are pointing there, is just to test that our environment variable to enable CAS is also turning on all the other `django.contrib.auth`, and `django-cas-ng` specific application settings.

That said, I'm fine with the application specific settings for something like the taxonomy, that is an application, for which the code would live in this repo, and thus would be testable there, but even in that case there will still likely be a need to add tests to `<repo_root>/tests/test_settings.py` and `<repo_root>/tests/test_urls.py` to verify the app is properly configured and the URLs are enabled for the LORE project in order to prevent people from accidentally removing those apps from the `lore/urls.py` and overrides done in `lore/settings.py`.

For this case this auth enablement `app` just like the `ADMINS` app is 100% pure configuration and there is no code included to do the actual protocol handling, views, ticket negotiation, or redirection logic, so this is just the testing of that configuration to prevent regressions and ensure that when the flag is set to turn it on, all the other pieces are properly enabled as well.

bdero:
Did another pass over the code and I have no additional comments. :penguin:

Ferdi:
@carsongee the piece of code that I'm referencing is CAS specific which is going to be used by the MIT instance of LORE (thus MIT specific code).   Harvard (was just an example) might not use CAS so there is not need for them to have CAS specific code in their code base.

tests/test_urls.py
carsongee: I just wanted to apologize, my response came back pretty gruffly here.  I was in a rush as I was working on some other topics and didn't think it through all the way before hitting the submit button on my comment.

PR #37, opened 2015-05-21
Taxonomy models and API

noisecapella:


noisecapella:
This depends on #21 to be merged first

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+    Term,
+)
+from learningresources.models import (
+    LearningResource,
+    LearningResourceType,
+    Repository,
+)
+
+
+# pylint: disable=no-member
```
ShawnMilo: We've now disabled this globally, so you can remove these.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Lookup vocabulary given its id
+
+    Args:
+        vocabulary_id (int): Vocabulary id
+
+    Returns:
+        Vocabulary: The vocabulary from the database
+    """
+    return Vocabulary.objects.get(id=vocabulary_id)
```
ShawnMilo: This could blow up if given a bad ID. A couple of options are [get_object_or_404](https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404) or doing a `.filter()` and checking whether its `.count() == 0` and returning `None` if so. 

I like the 404 option better, because:
- You're not returning different data types (`None` versus a `Vocabulary` instance)
- If someone causes that then they did something wrong
- Calling code doesn't have to check if the returned values `is None` before trying to use it

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+        required (bool): Is Vocabulary required?
+        vocabulary_type (unicode): Vocabulary type
+        weight (int): Weight
+
+    Returns:
+        Vocabulary: The Vocabulary which was just added to the database
+
+    """
+
+    if vocabulary_type != "m" and vocabulary_type != "f":
```
ShawnMilo: This manual checking is easy to get out-of-sync with the model choices. It's probably better to use Django's built-in [clean methods](https://docs.djangoproject.com/en/1.8/ref/models/instances/#validating-objects).

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+        raise TaxonomyAPIException("vocabulary_type must be either"
+                                   " 'm' (managed) or 'f' (free tagging)")
+
+    repository = Repository.objects.get(id=repository_id)
```
ShawnMilo: Same comment about the instance not existing from above.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+
+    return Vocabulary.objects.create(
+        repository=repository,
```
ShawnMilo: There's no need to actually do a database query to get the repository first, if this is changed to `repository_id=repository_id`. Note that the fact that they're both `repository_id` is a coincidence:

`repository_id` in the model works because there's a field named `repository` that is a foreign key, so you can refer to the _primary key_ (as opposed to the _instance_) by appending `_id`.

`repository_id` happens to be the variable name you're using in the function. Had it been `foo`, the line would be `repository_id=foo`.

Of course, this will cause a _worse_ problem if the repository doesn't exist, so this may not be the best solution here. Please check and see if the `clean` methods above handle this automatically.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+        name=name,
+        description=description,
+        required=required,
+        vocabulary_type=vocabulary_type,
+        weight=weight,
+    )
```
ShawnMilo, line 58: Also, this might need to be in a transaction.

taxonomy/api.py
noisecapella: If the id is assigned directly without the database check the number may be invalid. Do you think that's a reasonable thing to check for here or should we rely on the caller already having validated that? 

Would it be more appropriate to not check and get a DoesNotExist error the next time Vocabulary.repository is called instead? It may be the more pythonic way, not sure

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def delete_vocabulary(vocabulary_id):
+    """
+    Delete Vocabulary with given id
+
+    Args:
+        vocabulary_id (int): Vocabulary id
+    """
+    Vocabulary.objects.filter(id=vocabulary_id).delete()
```
ShawnMilo: This is dangerous, because it could cascade delete records from other tables.

Please see the [on_delete](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.on_delete) documentation and update the model as appropriate.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Get Term with existing id
+
+    Args:
+        term_id (int): Term id
+
+    Returns:
+        Term: The Term with the id
+    """
+    return Term.objects.get(id=term_id)
```
ShawnMilo: Same comment about `get_object_or_404` above.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Args:
+        vocabulary_id (int): Vocabulary id
+        label (unicode): Term label
+        weight (int): Weight for term
+
+    Returns:
+        Term: The newly created Term
+    """
+    vocabulary = Vocabulary.objects.get(id=vocabulary_id)
```
ShawnMilo: Same comment about get_object_or_404 above. I'll stop adding comments about this so I don't get too annoying, so please check any `.get()` statements.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def delete_term(term_id):
+    """
+    Delete an existing Term
+
+    Args:
+        term_id (int): Term id
+    """
+    Term.objects.filter(id=term_id).delete()
```
ShawnMilo: Same comment about unanticipated `cascade delete` actions as above. I'll also stop commenting on these; please set appropriate `on_delete` settings on any relevant fields.

taxonomy/api.py
ShawnMilo: I think using the model's `full_clean` method first to see if it works is best. If that doesn't work we'll come up with another plan. I don't think the caller should do the validation in this case, because that could lead to duplicated checks.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Get Terms using a particular LearningResource
+
+    Args:
+        learning_resource_id (int): Learning object id
+
+    Returns:
+        sequence of Term relating to existing LearningResource
+    """
+    return (LearningResource.objects.get(id=learning_resource_id).terms.
```
ShawnMilo: It's more Pythonic not to use extra parens for wrapping, since you get that "for free" from the parens already in the line.

``` python
return LearningResource.objects.get(
    id=learning_resource_id).terms.order_by('id')
```

ShawnMilo: It may be better to order by `title`.

taxonomy/api.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Get vocabularies supporting the given learning object type
+
+    Args:
+        learning_resource_type (unicode): The learning resource type
+
+    Returns:
+        sequence of Vocabulary objects supporting the learning resource type
+    """
+    return (LearningResourceType.objects.get(name=learning_resource_type).
```
ShawnMilo: Same "extra parens" comment. I'll also stop mentioning these if there are any more.

taxonomy/models.py
```diff
@@ -0,0 +1,37 @@
...
+        ("m", "managed"),
+        ("f", "free tagging")
+    ))
+    weight = models.IntegerField()
+
+    learning_resource_types = models.ManyToManyField(
+        LearningResourceType,
+        related_name="vocabularies"
+    )
+
```
ShawnMilo, line 32: There should probably be a [unique_together](https://docs.djangoproject.com/en/1.8/ref/models/options/#unique-together) here so we don't get duplicate names within the same repository.

taxonomy/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Term(models.Model):
+    """Model for term table"""
+    vocabulary = models.ForeignKey(Vocabulary)
+    label = models.TextField()
```
ShawnMilo, line 42: Probably another use for `unique_together`, if the label can exist for multiple vocabularies, but not multiple times for the same vocabulary.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+        add_term_to_learning_resource(self.learning_resource.id, self.term.id)
+        self.assertEquals([self.term],
+                          list(self.learning_resource.terms.all()))
+        add_term_to_learning_resource(self.learning_resource.id, self.term.id)
+
+        # duplicate term is ignored
+        self.assertEquals([self.term],
+                          list(self.learning_resource.terms.all()))
+
+        self.assertRaises(LearningResource.DoesNotExist,
```
ShawnMilo: These are good tests, but if these cases happened in production it would probably be a runtime error, resulting in a 500 response to the user.

The suggestions I made about using `full_clean` within the API calls will help with this.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def test_add_type_for_vocabulary(self):
+        """
+        Test add_type_for_vocabulary
+        """
+        self.assertEquals(0, self.vocabulary.learning_resource_types.count())
+        add_type_for_vocabulary(self.learning_resource_type1.name,
+                                self.vocabulary.id)
+        self.assertEquals([self.learning_resource_type1],
+                          list(self.vocabulary.learning_resource_types.all()))
```
ShawnMilo, line 115: I like these tests. Normally I would check that the `count()` had increased by the exact number I expected, but this kills three birds with one stone by asserting the count is correct, what we think is in the database is what's actually there, and that there's nothing unexpected in the database.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+        self.assertEquals([self.term, term],
+                          list(Term.objects.order_by('id')))
+
+        self.assertRaises(Vocabulary.DoesNotExist,
+                          lambda: create_term(456, "label", 9))
+        self.assertRaises(IntegrityError,
+                          lambda: create_term(self.vocabulary.id, None, 9))
+        create_term(self.vocabulary.id, "label2", 99.7)
+        # term weight is truncated after being stored in database
+        self.assertEquals(1, Term.objects.filter(weight=99).count())
```
ShawnMilo: This is testing Django's `IntegerField`, rather than our code.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def test_create_vocabulary(self):
+        """
+        Test create_vocabulary
+        """
+        self.assertEquals([self.vocabulary],
+                          list(Vocabulary.objects.all()))
+
+        v = create_vocabulary(self.repository.id,
```
ShawnMilo: Single-letter variable name. Does this pass pylint? I think `vocab` would be nicer.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+                              "vocabulary",
+                              "description",
+                              False,
+                              "f",
+                              9)
```
ShawnMilo: It could be just me, but I think all the `"f"` and `"m"` values here should be replaced with variables at the top of the function. Maybe even with constants defined in models.py and used within the field `choices()` and imported here (and elsewhere) when creating instances.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+                          lambda: create_vocabulary(9,
+                                                    "vocabulary",
+                                                    "description",
+                                                    False,
+                                                    "m",
+                                                    9))
+
+        # int is cast to string
+        create_vocabulary(self.repository.id,
+                          9,
```
ShawnMilo: This is testing Django, not our code. Also, it should never be possible for an integer to be passed to the API because all the values will come from `ModelForm`s, which will treat the `POST` value as text.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+                          "m",
+                          9)
+        self.assertEquals("9", Vocabulary.objects.order_by('id').
+                          last().description)
+
+        # "abc" is cast to boolean True
+        create_vocabulary(self.repository.id,
+                          "vocabulary",
+                          "description",
+                          "abc",
```
ShawnMilo: Same comments as above about testing Django's code, and that forms will not allow the wrong types to be passed in. I'll stop mentioning these even if I see more.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+                          "f",
+                          9)
+        self.assertEquals(True, Vocabulary.objects.order_by('id').last().
+                          required)
+
+        # pass in None, violates constraints
+        self.assertRaises(IntegrityError,
```
ShawnMilo: These kinds of problems will be prevented by the `is_valid()` method of `ModelForm`s, so these checks are unnecessary. If the API is used directly by a yet-unplanned app, the model `full_clean` will also prevent this.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,346 @@
...
+        self.assertEquals([],
+                          list(get_terms_for_learning_resource(
+                              self.learning_resource.id)))
+
+        self.term.learning_resources.add(self.learning_resource)
+        self.assertEquals([self.term],
+                          list(get_terms_for_learning_resource(
+                              self.learning_resource.id)))
+        self.assertRaises(LearningResource.DoesNotExist,
+                          lambda: get_terms_for_learning_resource(9))
```
ShawnMilo, line 255: The "magic numbers" used in these tests are distracting. Maybe use `0` or `-1` in all places (instead of sometimes `3`, sometimes `9`), or set `FAKE_ID = -1` at the top and use it everywhere.

taxonomy/api.py
noisecapella, line 58: The function should already be wrapped in a transaction, or did you notice something else?

taxonomy/api.py
noisecapella: What would make more sense after delete, setting the foreign key to null (and null=True in the model) or leaving the foreign key with its old value? I agree cascade deletes should be avoided

ShawnMilo: I think it would be best if we did not allow deletes if there are related records.

noisecapella: Sounds good. Is on_delete supported for ManyToManyFields too?

taxonomy/tests/test_api.py
noisecapella: Edited out your image in an effort to reduce nightmares :)

taxonomy/tests/test_api.py
noisecapella: `ModelForm`s would bypass the API functions anyway, right? The users of the API functions would be other pieces of code which should be passing in data with correct types.

Though I agree that asserting the casting behavior is probably not very useful.

taxonomy/tests/test_api.py
noisecapella: If `create_vocabulary` were using `full_clean` it would be raising a `ValidationError` instead. Wouldn't it still be worth asserting how it handles `None`?

taxonomy/api.py
ShawnMilo, line 58: Yeah, my mistake. I'd lost track of the function decorator when I wrote that.

taxonomy/api.py
ShawnMilo: It didn't jump out at me in the documentation. I'm guessing not, because it's not required, so it wouldn't cause any integrity issues when deleted. Since it wouldn't lose course data (just taxonomy data which is a form of metadata), this is probably just a design decision that needs to be made by Ferdi.

taxonomy/tests/test_api.py
ShawnMilo: You're right, when dealing with a single model, the ModelForm won't use the API. That suggests that maybe the API should use ModelForms and all their baked-in goodness. I've actually done that in the past and it works out well. But it does cause some overhead when you have Python objects and you need to massage them to look like they arrived via HTTP POST.

I'd say as long as the API uses the `full_clean` method of the ModelForm we should be good, as far as not duplicating validation logic.

taxonomy/tests/test_api.py
ShawnMilo: If you get a `ValidationError` for passing in bad or empty values, then specifically testing for a `None` is testing Django, so I'd say it's not worthwhile.

noisecapella:
Recreating...

PR #38, opened 2015-05-21
Add forms for taxonomy app

noisecapella:


learningresources/models.py
```diff
@@ -47,6 +47,9 @@ class LearningResourceType(models.Model):
     """
     name = models.TextField()
 
+    def __str__(self):
```
noisecapella: Is this the right place to tell the form widget how to describe the object?

taxonomy/forms.py
```diff
@@ -0,0 +1,33 @@
...
+    ModelForm,
+    TextInput,
+    CheckboxSelectMultiple,
+    RadioSelect,
+)
+
+from .models import Vocabulary
+
+
+class VocabularyForm(ModelForm):
```
noisecapella, line 15: There are descriptions in the screenshot for each field which aren't included in this form but maybe that can be done via CSS? Not sure

taxonomy/urls.py
```diff
@@ -0,0 +1,11 @@
+"""
+URLs for taxonomy app
+"""
+from django.conf.urls import url
+
+from taxonomy import views
+
+urlpatterns = [
+    url(r'^edit/$', views.edit),
+    url(r'^edit/(?P<vocabulary_id>[0-9]+)/$', views.edit),
```
noisecapella: Is there a way to combine these two?

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+from django.shortcuts import render, redirect
+
+from .forms import VocabularyForm
+from learningresources.models import Repository
+from taxonomy.models import Vocabulary
+
+
+def edit(request, vocabulary_id=None):
```
noisecapella: Should this be broken into two different functions? Or better organized for code reuse?

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+    form.instance.required = False
+    form.instance.weight = 1000
+
+    if request.POST:
+        form = VocabularyForm(request.POST)
+
+        form.instance.repository = repository
+        # TODO: are these good defaults?
+        form.instance.required = False
+        form.instance.weight = 1000
```
noisecapella, line 71: Good defaults?

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+
+            form.save()
+            # TODO: get_absolute_url here?
+            return redirect(
+                edit, vocabulary_id=form.instance.id
+            )
+        if vocabulary_id is None:
+            return render(
+                request,
+                "vocabulary_create.html",
```
noisecapella: Are these filenames appropriate given our file structure? Does it make sense to have two different templates for create and edit?

taxonomy/views.py
noisecapella: Also, is 'edit' an appropriate name for a view here? What is the Django preferred way?

learningresources/models.py
ShawnMilo: It's usually `__unicode__`. I don't thing `__str__` is necessary.

taxonomy/forms.py
ShawnMilo, line 15: Can you give an example? I think maybe the [help_text](https://docs.djangoproject.com/en/1.8/topics/forms/modelforms/) could be what you're looking for.

taxonomy/urls.py
ShawnMilo: Yes, you could make the regex pattern optional with a `?`.

taxonomy/views.py
ShawnMilo: Generally, create and edit should be the same template. They'll be using the same form.

taxonomy/forms.py
noisecapella, line 15: I'm referring to the text below each field on this screenshot, for example "Describe how content editors..." under the description textarea. Would help_text render in that way?

Screenshot

ShawnMilo, line 15: Yes, `help_text` is the best thing for that.

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # TODO: where to specify this?
+    repository = Repository.objects.first()
+
+    form.instance.repository = repository
+    # TODO: are these good defaults?
+    form.instance.required = False
+    form.instance.weight = 1000
+
+    if request.POST:
```
ShawnMilo: Use `if request.method == "POST"` instead of checking the post dict directly.

taxonomy/views.py
ShawnMilo: `vocabulary_id` should not be optional. I'd remove it now and simplify the function (and urls.py).

taxonomy/views.py
ShawnMilo: Handling both a GET and a POST in the same view can be handled with a little less code. See this for an example: https://github.com/mitodl/lore/blob/skm_import/importer/views.py#L29

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+        form = VocabularyForm(request.POST)
+
+        form.instance.repository = repository
+        # TODO: are these good defaults?
+        form.instance.required = False
```
ShawnMilo, line 70: The defaults should be in the model. I don't have a comment on whether these are good.

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+        form.instance.weight = 1000
+        if form.is_valid():
+
+            form.save()
+            # TODO: get_absolute_url here?
+            return redirect(
+                edit, vocabulary_id=form.instance.id
```
ShawnMilo: Using `reverse` is good here. [example](https://github.com/mitodl/lore/blob/skm_022_welcome/learningobjects/views.py#L34)

taxonomy/views.py
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    'form': form,
+                    'vocabulary_id': vocabulary_id,
+                }
+            )
+    else:
+        if vocabulary_id is None:
+            return render(
+                request,
+                "vocabulary_create.html",
```
ShawnMilo: nit: There doesn't need to be so much whitespace in these == `{'form': form}` is fine.

ui/templates/vocabulary_create.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% block content %}
+    <form action="{% url 'taxonomy.views.edit' %}"
```
ShawnMilo: nit: I don't think it's necessary or desirable to break up the `<form` line. Also, I believe `POST` is preferable to `post`.

taxonomy/forms.py
noisecapella, line 15: How does Django handle localization of `help_text`?

taxonomy/views.py
noisecapella: What if we're creating a new vocabulary? Should that go in a separate view?

taxonomy/forms.py
ShawnMilo, line 15: The same way Django does all its localization:

`from django.utils.translation import ugettext_lazy`

It's often imported as `_`, as in:

`from django.utils.translation import ugettext_lazy as _`

Then you handle it with `.po` files as usual in Python.

taxonomy/views.py
ShawnMilo: It should definitely be at a different URL, but it should use the same template or at least shared template fragment. It can be a different view, and it's so small it may as well be to avoid some `if` statements in the edit view to handle both.

Issue #8, opened 2015-05-12
Taxonomy app


Ferdi:
Provides the APIs for a) defining the taxonomy system, managing vocabularies and taxonomy terms and adding them for each repository and b) assigning taxonomy terms to learning objects.

See also https://goo.gl/gmpedb

noisecapella:
Some initial thoughts on an API for taxonomy. Does this make sense given the data?

GET taxonomy/learning_object/1/vocabularies - List of vocabularies for a learning object
GET taxonomy/learning_object/1/terms - List of valid terms for a learning object

GET taxonomy/vocabulary/1 - vocabulary information
POST taxonomy/vocabulary - create new vocabulary
PUT taxonomy/vocabulary/1 - update existing vocabulary
DELETE taxonomy/vocabulary/1 - delete vocabulary

GET taxonomy/term/1 - Get information about term
POST taxonomy/term - create term
PUT taxonomy/term/1 - update existing term
DELETE taxonomy/term/1 - delete term

GET taxonomy/term/1/learning_object - get a list of learning objects for a particular term
POST taxonomy/term/1/learning_object/7 - create link between the two
DELETE taxonomy/term/1/learning_object/7 - delete link

GET taxonomy/vocabulary/1/types - List of valid learning object types for vocabulary (Or maybe included in taxonomy/vocabulary/1)?
POST taxonomy/vocabulary/1/type - append a new learning object type to that list
DELETE taxonomy/vocabulary/1/type/type_name_here (is type name guaranteed to fit here?)

Ferdi:
It would be a good start on the REST api.  This should just be a  Django app that simply implements the models. What you have here would go in the REST app https://goo.gl/P3NsMd

Ferdi:
By "API" , I meant, public methods like :

```
taxonomy_get_vocabularies(type)
```

will return all vocabularies that applies to  learning object of certain type (e.g. "problems")
You'd then use that in REST to do what you have above 

GET taxonomy/vocabulary/1/types - List of valid learning object types for vocabulary

noisecapella:
Do these models make sense?

class Vocabulary(models.Model):
    vid = models.AutoField(primary_key=True)
    rid = models.ForeignKey("Repository")
    name = models.TextField()
    description = models.TextField()
    required = models.BooleanField()
    type = models.TextField()  # TODO: or enum? CharField?
    weight = models.IntegerField()  # TODO: or float?

class VocabularyUnitType(models.Model):
[...]

PR #42, opened 2015-05-26
Taxonomy models and API

noisecapella:


noisecapella:
Pull request for #8

taxonomy/api.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Returns:
+        list of unicode: sequence of learning object types supported by
+            a LearningResource
+    """
+    learning_resource_types = get_object_or_404(
+        Vocabulary, id=vocabulary_id
+    ).learning_resource_types.order_by('name')
+
+    return (x.name for x in learning_resource_types)
```
ShawnMilo: This is a [generator expression](https://docs.python.org/2/reference/expressions.html#generator-expressions). Was this intentional? It seems like it should be a list expression.

taxonomy/models.py
```diff
@@ -0,0 +1,51 @@
...
+    """Model for term table"""
+    vocabulary = models.ForeignKey(Vocabulary, on_delete=models.PROTECT)
+    label = models.TextField()
+    weight = models.IntegerField()
+
+    learning_resources = models.ManyToManyField(LearningResource,
+                                                related_name="terms")
+
+    # pylint: disable=missing-docstring,no-init,too-few-public-methods,
+    # pylint: disable=old-style-class
```
ShawnMilo, line 49: @carsongee Should we disable `old-style-class` universally? It seems like it'll come up a lot in Django models and ModelForms, and I doubt it'll be an actual issue.

carsongee, line 49: I'm down for that, it is pretty annoying and it would be weird really rare for us to do that in reality

ShawnMilo, line 49: Cool. @noisecapella: please add to `pylintrc`.

taxonomy/api.py
carsongee: It just depends on the use case really, generators are nice for memory if the caller won't need to index the list (not that I ever expect to have a huge lists of types (i.e. will probably only ever max out at 50-100), but the doc string suggests it should be a list, not a generator/iterable.

ShawnMilo: I was thinking along the lines of maybe someone checking the `len()` of the output and being surprised. I agree generators are nice. That's why I asked if it was intentional, so it could be made more clear, at least with a comment.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,328 @@
...
+                          list(get_learning_resources_for_term(self.term.id)))
+
+        self.assertRaises(Http404,
+                          lambda: get_learning_resources_for_term(9))
+
+    def test_get_term(self):
+        """
+        Test get_term
+        """
+        invalid_id = -1
```
ShawnMilo, line 239: This fixes one of the test failures. Before, instead of `invalid_id`, this was hard-coded to `9`. During testing, `self.term` was created at least 15 times, so often a term with pk `9` existed.

taxonomy/tests/test_api.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def test_remove_term_from_learning_resource(self):
+        """
+        Test remove_term_from_learning_resource
+        """
+        self.learning_resource.terms.add(self.term)
+        self.assertEquals([self.term],
+                          list(get_terms_for_learning_resource(
+                              self.learning_resource.id)))
+        remove_term_from_learning_resource(self.learning_resource.id,
```
ShawnMilo, line 302: This fixes the other bug. The arguments here were reversed (it was sending term ID then learning resource ID, but the function takes them in the other order). This probably doesn't fail in certain circumstances if they're both the same, probably `1`.

ShawnMilo:
I think this is ready to commit, but it would be good if someone could review [this commit](https://github.com/mitodl/lore/commit/b5c536eb10f7b16ce45da245c203d6ff50e2dd05), since I changed it.

You can also see my [notes from today](https://github.com/mitodl/lore/pull/42/files) explaining the changes.

carsongee:
Looks good but I'm going to run tests locally to double check.  I'll ping again when they are done and you can merge

carsongee:
Tests pass locally, you are good to merge  @ShawnMilo :fire_engine:

taxonomy/models.py
noisecapella, line 49: What about the other disabled items? We should keep missing-docstring globally, no strong opinion on the other two however

ShawnMilo, line 49: I agree on the docstring -- we should keep it. I think `no-init` and `too-few-public-methods` can be globally disabled, since the former happens all the time with subclasses and the latter (as well as too _many_ methods) often happen when subclassing Django and testcase classes. I'd like to get a second from @carsongee on those, though.

carsongee, line 49: I'm torn, but ok with either globally disabling, too-few-public-methods or no-init, or at least scoping those to the top of the models file.  I generally like the `too-few-public-methods` when making new classes, so I would hate to see it go, but we are mostly doing inheritance in this project anyway, so I don't think it would hurt us very bad to disable it all together. The missing-docstring however, should actually be scoped to the meta class though, as opposed to the root model class, but not a big deal, and definitely shouldn't be added to the top or global disable list.

noisecapella, line 49: Do pylint comments have scope or is that suggestion just a stylistic thing?

carsongee, line 49: It does respect block level and line level scope, though I don't know if I ever personally checked it working in memory: http://docs.pylint.org/faq.html#message-control

Issue #22, opened 2015-05-18
Welcome page


Ferdi:
Implement LORE  welcome page

![01_welcome](https://cloud.githubusercontent.com/assets/125561/7682973/ecdc4500-fd49-11e4-8002-b66418f98383.png)

PR #29, opened 2015-05-19
WIP: welcome page w/demo

ShawnMilo:
Demo [video](http://youtu.be/SiD65S37-lI).

PR #44, opened 2015-05-26
Welcome page. Closes issue #22.

ShawnMilo:
Supersedes #29 after a rebase against master.

### DIFF
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -114,3 +114,15 @@ def type_id_by_name(name):
         obj, _ = LearningResourceType.objects.get_or_create(name=name.lower())
     TYPE_LOOKUP[name] = obj.id
     return obj.id
+
+
+def get_repos(user):
+    """
+    Get all repositories a user may see.
+
+    Args:
+        user (auth.User): request.user
+    Returns:
+        repos query set of learningobject.Repository: repositories
+    """
+    return Repository.objects.filter(created_by__id=user.id).order_by('name')
--- a/learningresources/tests/test_views.py
+++ b/learningresources/tests/test_views.py
@@ -1,0 +1,38 @@
+"""
+Test the importer views to make sure they work.
+"""
+
+from __future__ import unicode_literals
+
+from learningresources.models import Repository
+
+from .base import LoreTestCase
+
+HTTP_OK = 200
+
+
+class TestViews(LoreTestCase):
+    """Hit each view."""
+
+    def test_welcome(self):
+        """Welcome page."""
+        resp = self.client.get("/lore/welcome", follow=True)
+        self.assertTrue(resp.status_code == HTTP_OK)
+        body = resp.content.decode("utf-8")
+        self.assertTrue("<h1>Welcome</h1>" in body)
+
+    def test_create_repo_get(self):
+        """GET repo creation page."""
+        resp = self.client.get("/lore/create_repo", follow=True)
+        body = resp.content.decode("utf-8")
+        self.assertTrue('<h1>Create repository</h1>' in body)
+
+    def test_upload_post(self):
+        """POST upload page."""
+        self.assertTrue(Repository.objects.count() == 0)
+        self.client.post(
+            "/lore/create_repo/",
+            {"name": "test name", "description": "test description"},
+            follow=True
+        )
+        self.assertTrue(Repository.objects.count() == 1)
--- a/learningresources/views.py
+++ b/learningresources/views.py
@@ -1,0 +1,40 @@
+"""
+Views for learningresources app.
+"""
+
+from django.shortcuts import render, redirect
+from django.core.urlresolvers import reverse
+from django.contrib.auth.decorators import login_required
+
+from learningresources.api import get_repos
+from learningresources.forms import RepositoryForm
+
+
+@login_required
+def welcome(request):
+    """
+    Greet the user, show available repositories.
+    """
+    return render(
+        request,
+        "welcome.html",
+        {"repos": get_repos(request.user)}
+    )
+
+
+@login_required
+def create_repo(request):
+    """
+    Create a new repository.
+    """
+    form = RepositoryForm()
+    if request.method == "POST":
+        form = RepositoryForm(data=request.POST)
+        if form.is_valid():
+            form.save(request.user)
+            return redirect(reverse("welcome"))
+    return render(
+        request,
+        "create_repo.html",
+        {"form": form},
+    )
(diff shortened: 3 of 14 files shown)


importer/templates/importer/status.html
```diff
@@ -0,0 +1,11 @@
+<p>
```
carsongee: I think this is already in, but a diff place, same with others here

learningresources/api.py
```diff
@@ -114,3 +114,19 @@ def type_id_by_name(name):
...
+
+    Args:
+        user (auth.User): request.user
+    Returns:
+        repos query set of learningobject.Repository: repositories
+    """
+
+    # Note: This fails pylint because "user" is not being used,
+    # but it can't be used just yet.
+    _ = user
```
carsongee: We have users, is there any reason not to actually use that?  If not, then I don't think we should run codee in runtime to make our static code checker happy, maybe just do a disable or change the param in the function to an underscore if that makes it happy, and we don't have any callers using kwargs yet.

ShawnMilo: You're right. It was only this morning that this was added (to remove the `dirty_hack` user). This code is kind of old at this point. I'll update it.

learningresources/tests/test_views.py
```diff
@@ -0,0 +1,54 @@
...
+
+from learningresources.models import Repository
+
+HTTP_OK = 200
+
+
+class TestViews(TestCase):
+    """Hit each view."""
+
+    def setUp(self):
```
carsongee: This is going to be such a common pattern I think we should run this out to a base class either in each app if we want to keep these apps de-composable, or into the project, and inherit from anywhere that needs a user in their tests.  I also think this would be cleaner as a one-liner ala:

``` python
self.user = User.objects.create_user(
            username='test', password='test_pass'
)
```

learningresources/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def welcome(request):
+    """
+    Greet the user, show available repositories.
+    """
+    return render(
+        request,
+        "welcome.html",
+        {"repos": get_repos(request.user)}
```
carsongee, line 21: I think you want to protect this with `@login_required` unless we want to allow for repos owned by anonymous as you are using the request.user here and that can be the stubby anonymous user object.

lore/urls.py
```diff
@@ -20,10 +20,13 @@
 from django.contrib import admin
 
 import importer.urls as importer_urls
+from learningresources.views import welcome, create_repo
 import cas.urls as cas_urls
 
 urlpatterns = [
     url(r'^admin/', include(admin.site.urls)),
     url(r'^importer/', include(importer_urls)),
+    url(r'^welcome/', welcome, name="welcome"),
```
carsongee: Can we please use an include here for URLs? If we still want to use the root url space, you can keep these two and redirect them to the app, or just have them be vanity URLs, but learningresources should have a urls.py of its own with its views.

carsongee: Also wouldn't these be "UI" urls with our design and go in the UI app?

learningresources/tests/test_case.py
```diff
@@ -0,0 +1,29 @@
...
+    """Handle often-needed things in tests."""
+
+    def setUp(self):
+        """set up"""
+        super(LoreTestCase, self).setUp()
+        self.username = 'test'
+        self.password = 'test'
+        self.user = User.objects.create(username=self.username)
+        self.user.set_password(self.password)
+        self.user.save()
```
carsongee: I think this is still probably clearer as just a single `User.objects.create` statement as mentioned earlier

learningresources/tests/test_case.py
```diff
@@ -0,0 +1,29 @@
...
+
+        self.client = Client()
+        self.client.login(username=self.username, password=self.password)
+
+    def tearDown(self):
```
carsongee: Why do we need this if it is just calling super? I think you can either remove this, or add a deletion statement for the user we created in setup if we want to be super explicit maybe?  Your call

lore/urls.py
```diff
@@ -20,10 +20,12 @@
 from django.contrib import admin
 
 import importer.urls as importer_urls
+import learningresources.urls as lore_urls
 import cas.urls as cas_urls
 
 urlpatterns = [
     url(r'^admin/', include(admin.site.urls)),
+    url(r'^lore/', include(lore_urls)),
```
carsongee, line 29: Are we sure this is the URL extension we want?  I know we have already used importer to house views, but if we are putting all the templates in the UI app, don't we also want to have all the views there?  @Ferdi and @pdpinch?  I'm fine with any way we do it, I'm just hoping for architectural guidance on reviews and future PRs with URLs, views, static assets and templates.

carsongee, line 29: Also I would think we want to redirect the root url to lore/welcome right?

ShawnMilo:
@carsongee Updated URLs so blank URL goes to `welcome` page.

carsongee:
There are still two open comments other than the question for Ferdi and Peter

pdpinch:
@cg is this the question you wanted @Ferdi or I to weigh in on? https://github.com/mitodl/lore/pull/44#discussion_r31077262

carsongee:
Here is the comment pulled out of the diff

> ```
> `url(r'^lore/', include(lore_urls)),`
> ```
> 
> Are we sure this is the URL extension we want? I know we have already used importer to house views, but if we are putting all the templates in the UI app, don't we also want to have all the views there? @Ferdi and @pdpinch? I'm fine with any way we do it, I'm just hoping for architectural guidance on reviews and future PRs with URLs, views, static assets and templates.

lore/urls.py
pdpinch, line 29: I have to agree, it seems confusing to have the templates in /ui and the views in /learningresource

But I'm inclined to leave as is for now until we work out the /ui app the way we want it -- hopefully next week.

pdpinch:
Ok... I don't know why my e-mailed comments aren't showing up here. Earlier, I wrote:

> I have to agree, it seems confusing to have the templates in /ui and the views in /learningresource
> 
> But I'm inclined to leave as is for now until we work out the /ui app the way we want it -- hopefully next week. 

If Carson's two other comments have been addressed, this should be merged.

ShawnMilo:
Pushed a rebase.

carsongee:
Ready for squash and merge :put_litter_in_its_place:

ShawnMilo:
Do you want your commits to be lost in another rebase? I was going to leave it as-is for proper attribution.

ShawnMilo:
Okay, rebase pushed. Just waiting on a last run of tests, then hopefully :accept:.

ShawnMilo:
@carsongee: Is this good to go?

PR #50, opened 2015-05-27
Add logging, especially for use during development and testing.

ShawnMilo:
Logging config.

### DIFF
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -47,6 +47,10 @@ def get_var(name, default):
 
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+LOGDIR = os.path.join(BASE_DIR, 'logs')
+
+if not os.path.exists(LOGDIR):
+    os.mkdir(LOGDIR)
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.8/howto/deployment/checklist/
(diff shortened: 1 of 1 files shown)


carsongee:
I actually don't want to log to a file, or at least if we are going to log to a file, we should also log to stdout/stderr for it to be a 12 factor app.  I would also like the log level to be configurable with an environment variable that defaults to `DEBUG`.  While we are here, it would also be nice to configure it to log to syslog.  As a specific example, our [flask logging extension](https://github.com/mitodl/flask-log/blob/master/flask_log.py) has some pointers on how to set things up, though it will be different in Django

carsongee:
Also, you have set here to disable existing loggers, but we still need the uncaught exception logger to be there so that we get e-mails on uncaught exceptions i.e.

``` python
...
        'filters': {
...
            'require_debug_false': {
                '()': 'django.utils.log.RequireDebugFalse',
            }
        },
...
        'handlers': {
...
            'mail_admins': {
                'level': 'ERROR',
                'filters': ['require_debug_false'],
                'class': 'django.utils.log.AdminEmailHandler'
            },
...
        'loggers': {
...
            'django.request': {
                'handlers': ['mail_admins'],
                'level': 'ERROR',
                'propagate': True,
            },
```

ShawnMilo:
My main goal was to having logging I can see when running tests. When I use a standard logging setup it doesn't show anywhere on the screen. If you know a way I can log during tests (even if it's on stdout -- I don't need a file either), I'd be happy to adopt that method.

carsongee:
I can go ahead and do this sometime soon, for the short term, you might try installing https://pypi.python.org/pypi/pytest-capturelog as it will spit out captured logging like nose does I think.  If that works, maybe drop a small PR for that, and we can hold off on the more sensitive logging situation to work for testing and production.

ShawnMilo:
This should be closed because it seems like we want to take an alternate approach.

PR #54, opened 2015-05-28
fixup! Welcome page and add repository

carsongee:
@ShawnMilo just a rename of the test file since it doesn't have tests in it, small rework of the setup method, and removed a comment that wasn't relevant anymore.

### DIFF
--- a/learningresources/tests/base.py
+++ b/learningresources/tests/base.py
@@ -11,14 +11,17 @@ from django.contrib.auth.models import User
 
 class LoreTestCase(TestCase):
     """Handle often-needed things in tests."""
+    USERNAME = 'test'
+    PASSWORD = 'test'
 
     def setUp(self):
         """set up"""
         super(LoreTestCase, self).setUp()
-        self.username = 'test'
-        self.password = 'test'
-        self.user, _ = User.objects.get_or_create(username=self.username)
-        self.user.set_password(self.password)
-        self.user.save()
+        self.user = User.objects.create_user(
+            username=self.USERNAME, password=self.PASSWORD
+        )
         self.client = Client()
-        self.client.login(username=self.username, password=self.password)
+        self.client.login(
+            username=self.USERNAME,
+            password=self.PASSWORD
+        )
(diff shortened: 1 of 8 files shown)


learningresources/tests/base.py
```diff
@@ -11,14 +11,17 @@
...
     def setUp(self):
         """set up"""
         super(LoreTestCase, self).setUp()
-        self.username = 'test'
-        self.password = 'test'
-        self.user, _ = User.objects.get_or_create(username=self.username)
-        self.user.set_password(self.password)
-        self.user.save()
+        self.user = User.objects.create_user(
+            username=self.TEST_USERNAME, password=self.TEST_PASSWORD
```
ShawnMilo: This works, but it's not technically correct. I read a great blog post that explained it really well a while back. I can't find it now, but [this one](https://timothyawiseman.wordpress.com/2012/10/06/class-and-instance-variables-in-python-2-7/) says pretty much the same thing.

The short version is that setting class variables is not the same as setting defaults for instance variables, and changing a value may not take effect in other places as intended, or may change things for other instances.

Short example to illustrate:

``` python
class Thing(object):
    fred = "bicycle"

    def something(self):
        print "self:", self.fred
        print "thing:", Thing.fred
        print ""

b = Thing()

print "before"
b.something()

b.fred = "monkey"

print "after"
b.something()
```

I think using  `LoreTestCase.TEST_USERNAME` and password would be better. Also, I don't think `TEST_` is necesasry in either variable name.

carsongee: Sure, that is basically the same case for not allowing empty lists as default arguments, but for this case, I think this is still the right choice.  These are literals, and the person inheriting the class should be able to modify them if they want to override `__init__` and change the user password and name which may be desirable.  I am also not using them as class variables, but object properties here as I don't mind if they are modified, i.e. no `LoreTestCase.TEST_USERNAME` here, I'm using self, so recognizing that they aren't intended to be at the class scope.  I also don't like defining properties outside of `__init__`, so If you have a strong preference I'm happy to move those into the `__init__`, I was just avoiding writing extra code.

ShawnMilo: Okay. They do work as-written. If it's intentional and not likely to bite another developer, then it's fine.

I still have the nit about the redundant `TEST_`, but not enough to not merge. Let me know if you want to change that and I'll merge.

carsongee: Dumped the TEST_

ShawnMilo:
Cool. Will merge once tests pass.

Issue #55, opened 2015-05-28
Create API documentation using Sphinx autodoc


pwilkins:
Add Sphinx documentation tools to generate API docs from code docstrings. Uses Google doc style through Napoleon and the Sphinx Bootstrap theme.
1. Add Sphinx autodoc tooling
# . Create TOX env to create Sphinx doc
# . Create ReadTheDocs site for LORE documentation

pwilkins:
closed by merge of PR#56

PR #56, opened 2015-05-28
Add Sphinx autodoc tooling

pwilkins:
Add Sphinx documentation tools to generate API docs
from code docstrings. Uses Google doc style through Napoleon
and the Sphinx Bootstrap theme.

See PT story here: https://www.pivotaltracker.com/story/show/95632294

Closes Github issue #55 - https://github.com/mitodl/lore/issues/55

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ MAINTAINER ODL DevOps <mitx-devops@mit.edu>
 # Add package files
 ADD requirements.txt /tmp/requirements.txt
 ADD test_requirements.txt /tmp/test_requirements.txt
+ADD doc_requirements.txt /tmp/doc_requirements.txt
 ADD apt.txt /tmp/apt.txt
 WORKDIR /tmp
 
@@ -16,6 +17,7 @@ RUN pip install pip --upgrade
 # Install project packages
 RUN pip install -r requirements.txt
 RUN pip install -r test_requirements.txt
+RUN pip install -r doc_requirements.txt
 RUN pip3 install -r requirements.txt
 RUN pip3 install -r test_requirements.txt
 
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -1,0 +1,4 @@
+-r requirements.txt
+sphinx>=1.3.1
+sphinx_bootstrap_theme>=0.4.5
+sphinxcontrib-napoleon>=0.3.5
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+import sys
+import os
+import sphinx_bootstrap_theme
+
+#pylint: skip-file
+
+sys.path.insert(0, os.path.abspath('..'))
+
+# django-specific configuration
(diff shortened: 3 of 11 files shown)


pwilkins:
@carsongee @ShawnMilo @pdpinch @noisecapella @giocalitri This is a first pass at API docs.  I'd welcome your comments and suggestions for improvement.

ShawnMilo:
It looks pretty straightforward to me -- mostly Sphinx boilerplate. Is there a one-step command to build the documentation with `docker-compose`?

pwilkins:
@ShawnMilo - @carsongee suggested creating a TOX env task to build the doc.  I don't know if he wanted that to run by default or only explicitly.

carsongee:
There are pep8 violations in the conf, either fix or disable in `pytest.ini`.  Also, there is a tox example right in the docs for sphinx: http://tox.readthedocs.org/en/latest/example/general.html#integrating-sphinx-documentation-checks

Dockerfile
```diff
@@ -16,8 +17,10 @@ RUN pip install pip --upgrade
 # Install project packages
 RUN pip install -r requirements.txt
 RUN pip install -r test_requirements.txt
+RUN pip install -r doc_requirements.txt
 RUN pip3 install -r requirements.txt
 RUN pip3 install -r test_requirements.txt
+RUN pip3 install -r doc_requirements.txt
```
carsongee: We just need to choose one or the other, no reason to slow the build for testing in both python languages.  I think readthedocs uses 2, so we can just do it there.

docs/conf.py
```diff
@@ -0,0 +1,271 @@
...
+#texinfo_appendices = []
+
+# If false, no module index is generated.
+#texinfo_domain_indices = True
+
+# How to display URL addresses: 'footnote', 'no', or 'inline'.
+#texinfo_show_urls = 'footnote'
+
+# If true, do not generate a @detailmenu in the "Top" node's menu.
+#texinfo_no_detailmenu = False
```
carsongee: I would just nuke everything that is commented out.

Dockerfile
pwilkins: @carsongee Do I need to change tox.ini to explicitly run the docs only in py27?

carsongee:
Looks like the rtfd build is failing on the django conf stuff: https://readthedocs.org/builds/lore/2886899/

pwilkins:
ReadTheDocs was failing because it initially tried to build docs from the master branch.  I switched it to the proper branch for the docs.

doc_requirements.txt
```diff
@@ -0,0 +1,3 @@
+sphinx>=1.3.1
+sphinx_bootstrap_theme>=0.4.5
+sphinxcontrib-napoleon>=0.3.5
```
carsongee, line 4: I think you want to add `-r requirements.txt` here and give rtfd this as the requirements to install.  That is how we've done it in the past.   That should simplify the tox build a bit too I think

carsongee:
Woot, this looks great! :loud_sound:

Issue #25, opened 2015-05-18
Add vocabulary page


Ferdi:
Simple page, no need for modal or special theming. 

![08_managevocab2](https://cloud.githubusercontent.com/assets/125561/7683265/d3d7ef08-fd4b-11e4-9e84-3db25554030a.png)

noisecapella:
Closed by #63

PR #63, opened 2015-06-01
Add and edit forms for vocabularies

noisecapella:


noisecapella:
For #25

taxonomy/models.py
```diff
@@ -16,25 +19,46 @@ class Vocabulary(models.Model):
...
-        (FREE_TAGGING, "free tagging")
-    ))
+    vocabulary_type = models.CharField(
+        max_length=1,
+        choices=(
+            (MANAGED, _("Managed")),
+            (FREE_TAGGING, _("Tag Style (on the fly)"))
+        ),
+        default="m",
+        help_text=_("Should terms be created in advance or on the fly?")
```
pwilkins, line 35: consider changing this message to, "Select whether vocabulary terms are managed or created as tags on-the-fly."  I am trying to find a help message that isn't a question.

pwilkins, line 35: this is the message in the UI mock-up from the designer, so I'll let it go as is.

PR #64, opened 2015-06-01
Listing page

ShawnMilo:
Pagination (hidden if only one page)

Adds repository ID implicitly to listing and upload pages.

Adds and populates slug field to repository.

Updates repository name to be a charfield instead of textfield, with corresponding correction to HTML (text input instead of textarea input).

Listing page itself, containing courses, runs, and LearningResource types.

Renamed `semester` to `run` throughout application.

Fix upload to accept a single course or multiple courses in subfolders (with @pwilkins).

Adds unique constraints to `LearningObject` models.

Add two links (import listing) to the "Welcome" page for each repository.


ui/templates/welcome.html
```diff
@@ -6,9 +6,26 @@
 
 <p>
```
Ferdi, line 9: The changes you have in this view are not necessary

importer/api/__init__.py
```diff
@@ -17,8 +16,10 @@
 
 from learningresources.api import create_course, create_resource
 
+log = logging.getLogger(__name__)
 
-def import_course_from_file(filename, user_id):
+
+def import_course_from_file(filename, repo_id, user_id):
     """
     Import OLX from .zip or tar.gz.
```
noisecapella, line 26: Docstring needs update for new argument

importer/api/__init__.py
```diff
@@ -56,10 +62,10 @@ def import_course_from_path(path, user_id):
     """
```
noisecapella, line 70: Docstring needs update for new argument

importer/api/__init__.py
```diff
@@ -56,10 +62,10 @@ def import_course_from_path(path, user_id):
...
     bundle = XBundle()
     bundle.import_from_directory(path)
-    return import_course(bundle, user_id)
+    return import_course(bundle, repo_id, user_id)
 
 
-def import_course(bundle, user_id):
+def import_course(bundle, repo_id, user_id):
     """
     Import a course from an XBundle object.
```
noisecapella, line 79: Docstring needs update for new argument

importer/api/__init__.py
```diff
@@ -36,17 +37,22 @@ def import_course_from_file(filename, user_id):
     except ArchiveException:
         remove(filename)
         raise ValueError("Invalid OLX archive, unable to extract.")
-    dirs = glob(join(tempdir, "*"))
-    if len(dirs) != 1:
-        rmtree(tempdir)
-        remove(filename)
-        raise ValueError("Invalid OLX archive, bad directory structure.")
-    import_course_from_path(dirs[0], user_id)
+    course_imported = False
```
noisecapella, line 48: Are we only interested in course.xml in the root directory or immediate subdirectory? If recursive we should use os.walk. If there's only one place course.xml should be we should only check there, or add a comment if the intent is to be flexible with input (I don't know the xbundle format so I'm not sure about my reasoning here)

importer/api/__init__.py
noisecapella, line 26: Also, can you update docstring to say filename will get deleted?

importer/forms.py
```diff
@@ -6,31 +6,28 @@
...
-        repos = Repository.objects.all().values_list('id', 'name')
         super(UploadForm, self).__init__(*args, **kwargs)
 
-        options = (('', '-- SELECT --'),) + tuple(repos)
-        self.fields['repository'].options = options
-
-    def save(self, user):
+    def save(self, user, repo_id):
         """
         Receives the request.FILES from the view.
```
noisecapella, line 46: Update docstring for new function arg

noisecapella, line 46: Also, would it make sense to pass in repo as an object given that user is an object as well? I'm not sure myself

importer/api/__init__.py
ShawnMilo, line 48: This is based on @carsongee's suggestion (made while you were out) that we only go two levels deep to prevent issues. Also, yes, we only expect it to be in the root or one level deep.

importer/forms.py
ShawnMilo, line 46: I'm just going to reverse that and pass just the user ID, as is being done elsewhere. I just missed that one. Thanks.

importer/urls.py
```diff
@@ -10,5 +10,5 @@
 
 urlpatterns = [
     url(r'^status/$', status, name='status'),
-    url(r'^upload/$', upload, name='upload'),
+    url(r'^upload/(?P<repo_id>\d+)', upload, name='upload'),
```
noisecapella: I'm guessing this should have a $ at the end otherwise it may match other suffixes, not sure what Django's behavior is here

importer/forms.py
ShawnMilo, line 46: The next thing we're going to do is add async calls, and passing around primary keys is better than passing around objects for various reasons. Also, I want to prevent people from having to instantiate objects when only the pk is going to be used anyway, which it is in these cases.

importer/views.py
```diff
@@ -28,17 +34,28 @@ def status(request):
...
+            except ValueError as ex:
+                log.debug("ex args: %s", ex.args)
+                if "Duplicate course" not in ex.args:
+                    raise ex
+                message = "Duplicate course"
     return render(
         request,
         "upload.html",
-        {'form': form},
+        {'form': form, "message": message, "repo_id": repo_id},
```
noisecapella: Is this the Django way to pass error messages? I'm not familiar with Django but their docs show an `errors` property which might be a better place to do that, not sure

learningresources/api.py
```diff
@@ -120,13 +103,93 @@ def type_id_by_name(name):
     return obj.id
 
 
-def get_repos(user):
+def get_repos(user_id):
     """
     Get all repositories a user may see.
 
     Args:
         user (auth.User): request.user
```
noisecapella, line 112: Docstring needs update

learningresources/api.py
```diff
@@ -120,13 +103,93 @@ def type_id_by_name(name):
...
+        repos query set of learningresource.Repository: repositories
+    """
+    return Repository.objects.filter(created_by__id=user_id).order_by('name')
+
+
+def get_repo_courses(repo_id):
+    """
+    Get courses for a repository.
+    Args:
+        repository (int): pk of learningresource.Repository
```
noisecapella: Docstring needs update (sorry for all these messages, trying to make it easier to spot these)

importer/views.py
ShawnMilo: You are correct about how errors are handled. This is not an error on the form because the form is, in fact, valid. So I added this message. It may be more correct to add a `clean()` override to the form with a `transaction.atomic` and have it check for a duplicate course so it makes `form.is_valid()` false. It just seemed like a lot more work for no gain. Opinion, @carsongee?

learningresources/api.py
```diff
@@ -120,13 +103,93 @@ def type_id_by_name(name):
...
+def get_runs(repo_id):
+    """
+    Get runs in all user's courses for the repo.
+    Args:
+        repo_id (int): primary key of the repository
+    Returns:
+        runs (list of strings): run names
+    """
+    courses = get_courses(repo_id)
+    return sorted(list(set([x.run for x in courses])))
```
noisecapella, line 167: Use courses.order_by('run').distinct() here instead? Probably not very important

learningresources/forms.py
```diff
@@ -35,11 +37,18 @@ class Meta:
...
     # not part of the form, and shouldn't be.
+    @transaction.atomic
     def save(self, user):
         """
         Save a newly-created form.
         """
         repo = super(RepositoryForm, self).save(commit=False)
+        slug = slugify(repo.name)
+        count = 1
+        while Repository.objects.filter(slug=slug).exists():
```
noisecapella: Is this typically how slugs are generated? Looks fine to me, just double checking that there's no awesome Django `generate_unique_slug()` function

ui/templates/welcome.html
pdpinch, line 9: All we want here are the links to the respositories. The user will just click through to the listing page and then import from there. L21-27 could just go.

learningresources/migrations/0001_initial.py
```diff
@@ -20,7 +20,7 @@ class Migration(migrations.Migration):
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
```
noisecapella, line 20: We should probably generate a new migration since Lore is already in production, to some degree.

learningresources/api.py
ShawnMilo, line 167: If I understand that properly, then no, because the same run may be in multiple courses, and we just want unique runs. I think this would be simpler with a raw SQL query, like

``` sql
SELECT DISTINCT(run) FROM learningresources_course ORDER BY run
```

Maybe I should do that, if nobody objects to the use of `objects.raw()`.

noisecapella, line 167: Wouldn't `courses.order_by('run').distinct('run')` do the same thing as that SQL?

learningresources/forms.py
ShawnMilo: I don't think so. I did some looking and although Django does have a [SlugField](https://docs.djangoproject.com/en/1.8/ref/models/fields/#slugfield), [this post](https://keyerror.com/blog/automatically-generating-unique-slugs-in-django) seems to indicate that Django won't automatically prevent dupes even if you set `unique=True`. If you find something that lets me eliminate this loop please let me know, though. That would be nice.

learningresources/migrations/0001_initial.py
ShawnMilo, line 20: In general, you're correct. At this point, it's fine to delete the dummy DB on the Heroku server, as per @carsongee. During development it's best to try to collapse migrations as much as possible because they quickly get out of hand.

noisecapella, line 20: Makes sense

ui/templates/welcome.html
ShawnMilo, line 9: As discussed yesterday, I thought we didn't want to allow the user to go to the listing page when there are no imported courses. Does this change that? 

My understanding of the comments above is that the "Welcome" page will contain a single link per repo, which brings you to the (possibly blank) listing page, and the listing page will now have the import link. Is that correct?

pdpinch, line 9: That's correct: 

> My understanding of the comments above is that the "Welcome" page will contain a single link per repo, which brings you to the (possibly blank) listing page, and the listing page will now have the import link. Is that correct?

Sorry about the confusion. I think someone made a comment about going to an empty listing page, but the solution is to add an import link to the listing page, not on this welcome page. We want to direct the user flow always through the repository.

importer/api/__init__.py
pwilkins, line 48: So the comment could be: "A valid OLX archive has a single occurrence of the file `course.xml` in its root directory, or no `course.xml` in its root and a single occurrence of `course.xml` in one or more of the root directory's children." 

I've shied away from adding many comments since it's so easy for comments to become out of sync with the code they describe.

ui/templates/welcome.html
ShawnMilo, line 9: Okay, that makes sense. So despite the "empty" listing page, it wouldn't be a dead-end, because the next logical step (an import) is available right there.

In fact, all the headers should just be hidden if there are no imported courses, so they get nothing except the import link. I'll do that.

importer/api/__init__.py
ShawnMilo, line 48: @pwilkins: agreed, we should probably be that specific, since that does reflect the real-world.

code commends

taxonomy/migrations/0001_initial.py
```diff
@@ -9,7 +9,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('learningresources', '0001_initial'),
+    #    ('learningresources', '0001_initial'),
```
noisecapella: Looks like an accidental commit, feel free to leave it if it's intentional

ShawnMilo: Thanks! I had to comment it out temporarily to recreate the `learningobjects` migration and failed to fix it.

learningresources/models.py
```diff
@@ -15,10 +15,14 @@ class Course(models.Model):
     repository = models.ForeignKey('Repository')
     org = models.TextField()
     course_number = models.TextField()
-    semester = models.TextField()
+    run = models.TextField()
     import_date = models.DateField(auto_now_add=True)
     imported_by = models.ForeignKey(User)
 
+    class meta:
```
noisecapella: These classes should be Meta with a capital M

ShawnMilo: Thanks, good catch.

carsongee:
This will need a rebase due to the UI changes

ShawnMilo:
@cg: I know I need to rebase anyway, and I merged from `master` a few minutes ago. Or is something else landing in `master` today after 323502c509b2ef133cdf7b65ad655ed45c4b1dd9?

I'm just waiting to rebase until after @noisecapella gives me the all-clear.

importer/views.py
carsongee: With 1.7 we have the awesome: https://docs.djangoproject.com/en/1.7/ref/forms/api/#django.forms.Form.add_error, so best of both I think, though clean is still probably the "right":tm: way to do that

importer/forms.py
```diff
@@ -6,48 +6,60 @@
...
+        raise ValidationError("Unsupported file type.")
 
-    def save(self, user):
+    def save(self, user_id, repo_id):
         """
         Receives the request.FILES from the view.
 
         Args:
-            user (auth.User): request.user
+            user_id (int): primary key of the user uploading the course.
```
noisecapella, line 48: Docstring needs updating for repo_id

learningresources/api.py
carsongee, line 167: Here is the SQL for `Course.objects.all().order_by('run').distinct('run')`:

```
[{u'sql': u'SELECT DISTINCT ON ("learningresources_course"."run") "learningresources_course"."id", "learningresources_course"."repository_id", "learningresources_course"."org", "learningresources_course"."course_number", "learningresources_course"."run", "learningresources_course"."import_date", "learningresources_course"."imported_by_id" FROM "learningresources_course" ORDER BY "learningresources_course"."semester" ASC LIMIT 21'
```

which is pretty close to what you have.

ui/templates/listing.html
```diff
@@ -0,0 +1,83 @@
...
+<div>
+    <a href="{% url 'upload' repo.id %}">import</a>
+</div>
+
+<div id="nav" style="float: left; height: 100%;">
+<p>
+<h2>Repository {{ repo.name }} </h2>
+</p>
+
+{% if courses.count > 0 %}
```
carsongee, line 16: Just a suggestion, but it might be nice to roll all this into an include file so it is easier to see what is going on here with the logic.

ShawnMilo:
Pushed rebase.

ShawnMilo:
Oh no, minus 0.68%! _hangs head in shame_

importer/tests/test_import.py
```diff
@@ -57,19 +86,44 @@ def test_import_toy(self):
...
     def test_invalid_file(self):
         """Invalid zip file"""
-        self.assertTrue(LearningResource.objects.count() == 0)
-        with self.assertRaises(ValueError):
-            import_course_from_file(self.bad_file, self.user.id)
-        self.assertTrue(LearningResource.objects.count() == 0)
+        try:
+            import_course_from_file(self.bad_file, self.repo.id, self.user.id)
+            raise ValueError("shouldn't happen")
+        except ValueError as ex:
```
carsongee, line 108: This is better done as:

``` python
with self.assertRaises(ValueError):
    import_course......
```

part of the coverage problem is due to this pattern

Issue #75, opened 2015-06-03
As a user, I'd like URLs that contain my repo name


pdpinch:
Gathering related items from #73 and #74 
- [x] change the url of the listing page from /lore/listing/3/1 to /repositories/repo_slug
- [x] change /lore/create_repo to repositories/new
- [x] Change "import" url from importer/upload/repo_id to repositories/repo_slug/import or repositories/repo_id/import
- [x] change /lore/welcome  to  /home

noisecapella:
What about the export url? Currently it's export/<resource_id>

noisecapella:
And status?

ShawnMilo:
I believe status is deprecated. It was just there so I could demo that the original import branch did something.

noisecapella:
Ok, I'm moving status to /status and export to /export and we can remove those in a later story

noisecapella:
Remove status I mean, move export somewhere appropriate

Ferdi:
Yes, good point on "export".  

export/resource_id should be changed to   learningresources/resource_id  or     /lr/resource_id

pdpinch:
fixed via d826cfb

PR #79, opened 2015-06-03
Move most URLs to ui/urls.py, minor URL reformatting

noisecapella:
For #75


importer/views.py
```diff
@@ -34,12 +34,14 @@ def status(request):
...
-def upload(request, repo_id):
+def upload(request, repo_slug):
     """
     Upload a OLX archive.
     """
-    if int(repo_id) not in set([x.id for x in get_repos(request.user.id)]):
+    if repo_slug not in set([x.slug for x in get_repos(request.user.id)]):
         return HttpResponseForbidden("unauthorized")
+    repo = [x for x in get_repos(request.user.id) if x.slug == repo_slug][0]
+    repo_id = repo.id
```
ShawnMilo: Nit. No need to make `repo_id` a variable, since `repo` is in scope, so `repo.id` can just be used in the one place it's needed.

learningresources/tests/test_views.py
```diff
@@ -15,9 +15,16 @@
...
-        """Status page."""
-        resp = self.client.get("/lore/welcome", follow=True)
+    def test_get_home(self):
+        """Welcome page."""
+        resp = self.client.get("/home", follow=True)
+        self.assertTrue(resp.status_code == HTTP_OK)
+        body = resp.content.decode("utf-8")
+        self.assertTrue("Welcome" in body)
+
+    def test_get_root(self):
```
ShawnMilo: It seems that this and the previous test should be combined to a single test, asserting that they both work and that they arrive at the same place.

Arriving at the same place _could_ be checked by checking that the body (or a hash of the body) is identical, but it's probably optional; just continuing to check for the 'Welcome' string should be enough. But combining the two tests should be done for clarity.

learningresources/views.py
```diff
@@ -51,16 +51,18 @@ def create_repo(request):
...
-    repo_id = int(repo_id)
+
+    # may work better as a database query, not likely a bottleneck though
     repos = get_repos(request.user.id)
-    if repo_id not in set([x.id for x in repos]):
+    if repo_slug not in set([x.slug for x in repos]):
         return HttpResponseForbidden("unauthorized")
-    repo = [x for x in repos if x.id == repo_id][0]
+    repo = [x for x in repos if x.slug == repo_slug][0]
+    repo_id = repo.id
```
ShawnMilo: Nit: `repo_id` is used many times, but is still unnecessary here.

ShawnMilo:
Looks good. Will merge when tests pass.

ShawnMilo:
@noisecapella Do tests pass for you locally?

noisecapella:
Yes, I'll investigate why Travis failed

noisecapella:
It looks like master was updated which broke a test, I'll fix and resquash

pdpinch:
I'll merge despite the too-low-to-measure drop in coverage.

PR #81, opened 2015-06-03
Fix model constraints in migrations.

### LINKED ISSUE #80: fix uniqueness constraints
It appears that #39 is being caused by constraints in the models that aren't in the migrations, probably caused by typos reported by @noisecapella, where `Meta` was spelled `meta`, and therefore didn't result in the migration file containing the appropriate constraints.

Deleting and recreating the migration file results in this diff:

``` diff
@@ -76,4 +74,12 @@ class Migration(migrations.Migration):
             name='repository',
             field=models.ForeignKey(to='learningresources.Repository'),
         ),
+        migrations.AlterUniqueTogether(
+            name='learningresource',
+            unique_together=set([('course', 'uuid')]),
+        ),
+        migrations.AlterUniqueTogether(
+            name='course',
+            unique_together=set([('repository', 'org', 'course_number', 'run')]),
+        ),
     ]
```

---

This requires a decision. Right now, the courses do not have UUIDs in the LearningResources, which means we have to create them. However, creating a _universally unique_ value is already going to guarantee uniqueness, so the database constraint seems unnecessary. 

So do we remove the `unique_together` or do we leave it and generate UUIDs?

If we decide to leave the constraint in, we will have to regenerate the initial migration again, so dev and Heroku databases will need to be deleted again to prevent errors.

ShawnMilo:
Fixes #80.

This will require deleting any local databases again, as well as the one on Heroku.

### DIFF
--- a/learningresources/migrations/0001_initial.py
+++ b/learningresources/migrations/0001_initial.py
@@ -29,7 +29,7 @@ class Migration(migrations.Migration):
             name='LearningResource',
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
-                ('uuid', models.TextField()),
+                ('uuid', models.UUIDField()),
                 ('title', models.TextField()),
                 ('description', models.TextField()),
                 ('content_xml', models.TextField()),
@@ -75,5 +75,13 @@ class Migration(migrations.Migration):
             model_name='course',
             name='repository',
             field=models.ForeignKey(to='learningresources.Repository'),
+        ),
+        migrations.AlterUniqueTogether(
+            name='learningresource',
+            unique_together=set([('course', 'uuid')]),
+        ),
+        migrations.AlterUniqueTogether(
+            name='course',
+            unique_together=set([('repository', 'org', 'course_number', 'run')]),
         ),
     ]
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -4,6 +4,8 @@ Learning resources data model
 
 from __future__ import unicode_literals
 
+from uuid import uuid4
+
 from django.db import models
 from django.db import transaction
 from django.contrib.auth.models import User
@@ -35,7 +37,7 @@ class LearningResource(models.Model):
     """
     course = models.ForeignKey(Course)
     learning_resource_type = models.ForeignKey('LearningResourceType')
-    uuid = models.TextField()
+    uuid = models.UUIDField()
     title = models.TextField()
     description = models.TextField()
     content_xml = models.TextField()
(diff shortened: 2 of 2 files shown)


carsongee:
Why not just use: https://github.com/dcramer/django-uuidfield/ and also, we will want a new migration. I also don't think we want every save to be a transaction, just creation right? Or am I missing something?

ShawnMilo:
@carsongee: Yes, I copy/pasted the `save()` below along with the decorator, but it's not really necessary, since the uuid4 won't repeat.

I'm not familiar with the `django-uuidfield` project, so I'll check it out. Why that and not the [built-in](https://docs.djangoproject.com/en/1.8/ref/models/fields/#uuidfield) one?

ShawnMilo:
Funny, I remember reading [this ticket](https://code.djangoproject.com/ticket/4682) years ago, back when Jacob had officially marked it `wontfix`. Evidently they did a 180.

carsongee:
The reason to use the project is avoid pinning us to 1.8.  It appears Django Guardian may require us to drop to 1.7

noisecapella:
Why don't we want to use Django 1.8?

ShawnMilo:
@noisecapella: See above. Django Guardian may not work properly with 1.8.

Otherwise, I'm sure'd we'd want the latest.

noisecapella:
Where does Django Guardian require 1.7? Their website says 1.3 and above

I would argue we want to pin ourselves to a particular Django version so I have certainty of what features we can use, although I'm not against changing the requirements if necessary

ShawnMilo:
@noisecapella: I don't know. I believe it was causing some kind of problem that @carsongee and @giocalitri encountered. I hope we don't have to drop back to 1.7.

ShawnMilo:
@pdpinch I don't know how the Coveralls gods decide, but we just got back that 0.0% we lost last time. ^_^

ShawnMilo:
Moved back to "needs review" after updating from master and using Django's UUIDField, based on our conversation this morning about not having to revert to an older version of Django.

This _will_ require wiping out the Heroku database again.

---

# BUT

Do we want to just remove the UUID field entirely? We've talked a few times about adding a subclass of `django.db.Model` and putting in `uuid`, `date_created`, and `date_modified` fields so the automatically propagate to all models. So we could remove this field and do that.

ShawnMilo:
Changing status due to decision needed.

Options:
- Add the base class, giving UUIDs (and automatic timestamps) to all models
- Remove this field entirely for now (there is no data to put in it unless we generate it upon import)
- Move forward with something like this pull request

I think the second option (remove the field for now) is best, unless we are officially going forward with moving all primary keys from auto-incrementing integers to UUIDs, in which case we should create the base class ASAP.

Ferdi:
I think we should :
- Add automatic timestamps
- Add UUIDs (at least for leanrningresource)   -- I'm not sure if they have to be the primary key or a second field.  I like the idea of having them as primary keys but it feels a bit experimental to me.

ShawnMilo:
@Ferdi @carsongee 

Someone (I think Carson) suggested keeping ints as primary keys, but using the UUIDs for the foreign key mappings. That would maintain the relationships outside of the database without conflict.

Do you like this plan?
- Add base class with auto timestamp for add, auto-timestamp for modified, and (non-required) UUID
- Add migration to backfill all blank UUIDs with a UUID
- Add migration to make UUIDs required
- Add migrations to change foreign keys from pointing to the pk to the UUID

ShawnMilo:
Closing in favor of working on #134.

PR #83, opened 2015-06-03
Add styling to LearningObjects listing page 2

pwilkins:
Add styling to LORE listing page, implementing designer's templates.
This is an initial pass at styling the page since the finished design
requires features that are not yet implemented.

Please review - @ferdi @pdpinch @carsongee

Closes Issue #73

PR #85, opened 2015-06-03
Feature/cg/list styling

carsongee:
This should supersede #83.  @pwilkins can fix things, but due to github issues I couldn't PR his PR.

### DIFF
--- a/ui/templates/listing.html
+++ b/ui/templates/listing.html
@@ -1,105 +1,106 @@
-{% extends "base.html" %}
+{% extends "repo_base.html" %}
+{% block title %}Listing{% endblock %}
 {% block content %}
 <div id="header"></div>
-
-<div>
-    <a href="{% url 'upload' repo.slug %}">import</a>
-</div>
 <div class="col-md-3">
-    <div class="panel-group lore-panel-group" id="nav">
-        <div class="panel panel-default">
-            <div class="panel-heading">
-    {% if courses.count > 0 %}
-                <h4 class="panel-title">
-                    <a class="accordion-toggle" data-toggle="collapse" data-parent="#accordion" href="#collapse-1">
-                        Course
-                    </a>
-                </h4>
-            </div>
-        </div>
-
-    <ul>
-    {% for course in courses %}
-        <li>
-            {{ course.course_number }}
-        </li>
-
-    {% endfor %}
-    </ul>
-
-        <div class="panel panel-default">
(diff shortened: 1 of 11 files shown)


ui/templates/upload.html
```diff
@@ -1,5 +1,6 @@
-{% extends "base.html" %}
+{% extends "repo_base.html" %}
```
pdpinch, line 1: Should this extend repo_base or just base?

I'm seeing the import button from repo_base, which we don't want here. 

screen shot 2015-06-03 at 4 28 09 pm

carsongee, line 1: I think we still want it to extend repo_base, because when we add taxonomy to the navigation, we probably want it there.  We also probably want to let them know what repo they are in:

It also gets a bit hard to slice and dice, but I'll pull that out.

carsongee, line 1: Internet total terrible.  I just pushed a change to swap the nav to go to the listing page instead.  After this, I think @pwilkins  will have to take over

ui/templates/listing.html
```diff
@@ -1,83 +1,106 @@
...
+      {% endfor %}
+    </table>
+    <div class="row">
+      <div class="col-md-12">
+        <div class="lore-pagination">
+          {% if resources.has_other_pages %}
+          {% if resources.has_previous %}
+          <a href="{% url 'listing' repo.slug resources.previous_page_number  %}"><i class="fa fa-angle-left"></i></a>
+          {% endif %}
+          {{ resources }}
```
pdpinch, line 95: For later, we'd like to remove the text "< Page 2 of 35 >" I'm not exactly sure where this comes from. I'll open an issue -- just don't want to lose track of it. 

screen shot 2015-06-03 at 5 05 31 pm

pdpinch:
Ok. I'm going for it.

ui/templates/listing.html
ShawnMilo, line 95: @pdpinch This `{{ resources }}` line is what is showing `<Page 2 of 35>`.

pwilkins, line 95: I thought I could fiddle the css to make the text less obnoxious but it is insidious.  When I changed the font-size of lore-pagination it not only reduced the size of the text (good), but it also shrank the size of the angle-bracket links (not good).  If I hide the text, it changes the way the grid renders for some reason I haven't sussed out.  Harrumph.

PR #88, opened 2015-06-04
Add footer to listing page

pwilkins:
Add footer to listing page based on the markup in designer's template.  A very small code change.

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -86,6 +86,15 @@
     <div class="container">
     {% block content %}
     {% endblock %}
+    <footer>
+        <div class="container">
+            © 2015, Massachusetts Institute of Technology
+            <span class="links-footer">
+                <a href="">About LORE</a>
+                <a href="">Help</a>
+            </span>
+        </div>
+    </footer>
     </div>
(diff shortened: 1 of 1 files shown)


pwilkins:
A very small code change to add the page footer to the base template.  Please review - @Ferdi @pdpinch @carsongee

pdpinch:
I think this would be better outside of the "container" div. i.e. move the `footer` down below the `</div>` in live 98. I think that will left-align the footer as it is in the HTML design template at http://pureui.com/c/mit/lore/build001/

pwilkins:
Actually, that isn't where it's laid out on http://pureui.com/c/mit/lore/build001/  but I'd be happy to change it if you like.

pwilkins:
@pdpinch Do you want the change you proposed? The difference between the two locations is the alignment of the footer on the page.  The designer aligned the footer with the right edge of the page, while your suggestion aligns it with the list block.  It also loses the dotted line separator but I can get that back.

pdpinch:
Turns out I was looking at an older version when the sidebar was still there. 

:+1: I'll merge.

PR #98, opened 2015-06-04
Async processing of course uploads.

### LINKED ISSUE #28: Asynchronous Course Import
As a user, I would like to not have my course imports timeout when I either have a slow Internet connection, a large course, or a bit of both.  Dependent on #4

ShawnMilo:
Closes #28.

### DIFF
--- a/importer/tasks.py
+++ b/importer/tasks.py
@@ -1,0 +1,13 @@
+"""
+Celery tasks for import module.
+"""
+from __future__ import unicode_literals
+
+from lore.celery import async
+from importer.api import import_course_from_file
+
+
+@async.task
+def import_file(path, repo_id, user_id):
+    """Asynchronously import a course."""
+    import_course_from_file(path, repo_id, user_id)
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -48,10 +48,6 @@ class LearningResource(models.Model):
     xa_avg_grade = models.FloatField(default=0)
     xa_histogram_grade = models.FloatField(default=0)
 
-    class Meta:
-        # pylint: disable=invalid-name,missing-docstring,too-few-public-methods
-        unique_together = ("course", "uuid")
-
 
 @python_2_unicode_compatible
 class LearningResourceType(models.Model):
--- a/lore/celery.py
+++ b/lore/celery.py
@@ -1,0 +1,19 @@
+"""
+As described in
+http://celery.readthedocs.org/en/latest/django/first-steps-with-django.html
+"""
+from __future__ import absolute_import
+
+import logging
+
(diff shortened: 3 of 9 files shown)


learningresources/models.py
```diff
@@ -46,10 +46,6 @@ class LearningResource(models.Model):
     xa_avg_grade = models.FloatField(default=0)
     xa_histogram_grade = models.FloatField(default=0)
 
-    class Meta:
```
noisecapella, line 51: Is this removal intentional? If so I'm fine with it, commenting just in case it's an oversight

importer/tasks.py
```diff
@@ -0,0 +1,11 @@
+"""
```
noisecapella, line 1: No from **future** import unicode here? It may be a good idea to add verifying this exists to our list of things to check when tox runs. 

We don't need it here but it would be good to be consistent across the application and to prevent possible oversights for future commits

learningresources/models.py
ShawnMilo, line 51: Yes, it was intentional because things were breaking without it. I could have fixed it with a migration (and did, for a while), but since #80 is out there pending, I figured I'd just wait for it to be merged instead of fixing it twice.

noisecapella, line 51: I figured, makes sense

importer/tasks.py
ShawnMilo, line 1: It's a good idea to have in general, and it's only not there because I forgot it. There are no strings in this file, though, so it doesn't matter.

lore/celery.py
```diff
@@ -0,0 +1,25 @@
...
+
+async = Celery('lore')  # pylint: disable=invalid-name
+
+# Using a string here means the worker will not have to
+# pickle the object when using Windows.
+async.config_from_object('django.conf:settings')
+async.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
+
+
+@async.task(bind=True)
```
noisecapella: Was this added for local testing? If it's not needed in production we should remove it

ShawnMilo: Yep, it was copy/paste out of the sample. But I've tested actual imports, so it should go. Good catch.

importer/tasks.py
ShawnMilo, line 1: Added (it fell below this comment).

noisecapella:
Looks good, I'll merge when Travis finishes

ShawnMilo:
:exclamation: ^_^

noisecapella:
The decrease in code coverage is from the tests doing a direct upload instead of using celery to do it async. Seems reasonable to me

PR #99, opened 2015-06-04
Changes pagination to use querystring.

### LINKED ISSUE #97: Update listing page pagination to use the querystring instead of the URL
Instead of `example.com/repo-name/2` for page 2, it should be `example.com/repo-name?page=2`.

There will be no `1` for page 1 -- that should be the default.

ShawnMilo:
Closes #97.


noisecapella:
Everything here seems straightforward, I'll merge this

Issue #100, opened 2015-06-04
Add REST API


noisecapella:
Acceptance Criteria:
- [x] Create a Django app called  rest
- [x] Integration with Django rest framework app
- [x] Github style (https://developer.github.com/v3/#pagination) pagination ( http://www.django-rest-framework.org/api-guide/pagination/)
- [x] Authentication ( session based for now ?)

Optionally (looks like @noisecapella has started):
- [x] vocabulary rest API

noisecapella:
Are we using HATEOAS @Ferdi?

Ferdi:
I would keep it very simple for now. E.g. Just enough for us to implement the admin forms.

Ferdi:
@noisecapella Can you split this into 2 PRs as described at #138  ?

noisecapella:
@ferdi Can you be more specific? What needs to be moved out of the PR linked above?

Ferdi:
@noisecapella I missed your description of this issue. Looks like you have defined a clear scope.

Ferdi:
@noisecapella I have changed the description of this issue to add some more details about what to expect. @noisecapella  good ?

noisecapella:
Yes these tasks all look good

pdpinch:
closed via #106

PR #106, opened 2015-06-05
Add rest app

noisecapella:


noisecapella:
For issue #100

rest/tests/test_rest.py
```diff
@@ -0,0 +1,150 @@
...
+        """Test REST access for term"""
+        vocab = create_vocabulary(
+            self.repo.id,
+            "name",
+            "description",
+            True,
+            Vocabulary.FREE_TAGGING,
+            1000
+        )
+        resp = self.client.get('/api/v1/terms/')
```
Ferdi: I thought it would be /api/v1/vocabs/1/terms ?

rest/tests/test_rest.py
```diff
@@ -0,0 +1,150 @@
...
+
+        input_dict = {
+            'name': 'name',
+            'description': 'description',
+            'required': True,
+            'weight': 1000,
+            'vocabulary_type': 'f',
+            'repository': self.repo.id,
+        }
+        resp = self.client.post('/api/v1/vocabularies/', input_dict)
```
Ferdi: Why '/' at the end of all urls ?

noisecapella: There was a discussion about trailing slashes in the taxonomy URLs issue. Basically Django will redirect URLs without the trailing slash to include the trailing slash if only that URL is defined, but not vice versa, so it's the more consistent option.

rest/tests/test_rest.py
noisecapella: This will be changed. I'm actually looking at /api/v1/repositories/slug/vocabularies/slug/terms now

Ferdi: yes, what you said, I forgot about repositories.  Just curious, from development perspective, which one is easier to do ,  "ids" or "slugs" ?

rest/tests/test_rest.py
Ferdi: So there is no way to have urls without trailing slashes in Django?

rest/tests/test_rest.py
noisecapella: They're pretty much the same, you just set lookup_field = 'name_of_column' in the serializer

Ferdi: ok, do we expect them (slugs) to change and if so do you see any problem with that ?

noisecapella: They will change if the user changes the name of the vocabulary or repository, or the label on the term. I think this kind of access is restricted to curators so it may not be a big deal. Ideally URLs [never change](http://www.w3.org/Provider/Style/URI.html) but the alternative is slugs that may not match up with the name at all if the name changes, which may be confusing.

lore/settings.py
```diff
@@ -104,7 +106,7 @@ def get_var(name, default):
     'guardian.backends.ObjectPermissionBackend',
 )
 
-ROOT_URLCONF = 'ui.urls'
+ROOT_URLCONF = 'lore.urls'
```
giocalitri: Why do you need this?

noisecapella: The rest app has its own urls so it made sense to move the base for urls back into the lore app and then include ui and rest urls from there.

giocalitri: I do not understand this division: why admin or as should be in lore.url? Plus I guess there was a reason why the urls have been moved to ui

rest/serializers.py
```diff
@@ -0,0 +1,90 @@
...
+            'id',
+            'slug',
+            'name',
+            'description',
+            'create_date',
+            'created_by',
+        )
+        read_only_fields = (
+            'slug',
+        )
```
giocalitri, line 42: should id be read only too?

giocalitri, line 42: do we want people to be able to modify the created_by?

lore/settings.py
noisecapella: Would you rather have these listed in ui.urls, then include the rest urls from there? I am fine either way

giocalitri: I guess it makes sense to move the lore.urls back to ui. Not sure if the rest urls should be in the rest app: I do not like having everything in ui, but if this is the path we chose to follow, I guess we have to have everything there.

rest/serializers.py
```diff
@@ -0,0 +1,90 @@
...
+            'name',
+            'description',
+            'vocabulary_type',
+            'required',
+            'weight',
+            'repository',
+        )
+        read_only_fields = (
+            'slug',
+        )
```
giocalitri, line 70: like before

rest/serializers.py
```diff
@@ -0,0 +1,90 @@
...
+        fields = (
+            'id',
+            'slug',
+            'label',
+            'weight',
+            'vocabulary',
+        )
+        read_only_fields = (
+            'slug',
+        )
```
giocalitri, line 95: like before

rest/serializers.py
noisecapella, line 42: created_by is a HiddenField which cannot be overridden by the user. But I should add a test to verify that. 'id' should be marked read-only, probably 'create_date' too

rest/tests/test_rest.py
```diff
@@ -0,0 +1,471 @@
...
+from taxonomy.models import Vocabulary
+from rest.serializers import VocabularySerializer
+
+HTTP_OK = 200
+HTTP_CREATED = 201
+HTTP_NO_CONTENT = 204
+HTTP_BAD = 400
+HTTP_UNAUTHORIZED = 403
+HTTP_NOT_FOUND = 404
+HTTP_METHOD_NOT_ALLOWED = 405
```
giocalitri: I know I did the same in my tests, but we should probably use [httplib](https://docs.python.org/2/library/httplib.html#httplib.HTTPS_PORT) for this

rest/tests/test_rest.py
```diff
@@ -0,0 +1,471 @@
...
+    """
+    REST test
+    """
+
+    def test_repositories(self):
+        """
+        Test for Repository
+        """
+        resp = self.client.get('/api/v1/repositories/')
+        self.assertEqual(HTTP_OK, resp.status_code)
```
giocalitri: in your tests you are doing this a lot: probably it makes sense to write a helper function that checks the status code (and maybe returns the resp or directly the JSON)

rest/tests/test_rest.py
```diff
@@ -0,0 +1,471 @@
...
+        repositories = as_json(resp)
+        self.assertEqual(1, repositories['count'])
```
giocalitri, line 41: Super nit, but shouldn't this be `self.assertEqual(repositories['count'], 1)`

noisecapella, line 41: I don't think there's a defined preference, the function arguments just say 'first' and 'second' for this

rest/tests/test_rest.py
noisecapella: Will do

rest/tests/test_rest.py
noisecapella: It looks like httplib is Python 2 only

rest/tests/test_rest.py
```diff
@@ -0,0 +1,742 @@
...
+
+
+class TestRest(LoreTestCase):
+    """
+    REST test
+    """
+    # pylint: disable=too-many-public-methods, too-many-arguments
+    def get_repositories(self, expected_status=OK):
+        """Get list of repositories"""
+        resp = self.client.get('/api/v1/repositories/')
```
giocalitri: since you use this base string a lot of times, probably it's better to declare it in the setUp of this test class and reuse it

rest/tests/test_rest.py
```diff
@@ -0,0 +1,742 @@
...
+                "description",
+                True,
+                Vocabulary.FREE_TAGGING,
+                1000
+            ) for i in range(100)]
+
+        resp = self.client.get(
+            '/api/v1/repositories/{repo_slug}/vocabularies/'.format(
+                repo_slug=self.repo.slug,
+            ))
```
giocalitri, line 428: why don't you use your function `get_vocabularies(self, repo_slug, expected_status=OK)` ?

rest/tests/test_rest.py
```diff
@@ -0,0 +1,742 @@
...
+        self.assertEqual(OK, resp.status_code)
+        vocabularies = as_json(resp)
+        self.assertEqual(100, vocabularies['count'])
+        self.assertEqual([VocabularySerializer(x).data
+                          for x in expected[20:40]], vocabularies['results'])
+
+    def test_authentication(self):
+        """Make sure unauthenticated and unauthorized users can't get access"""
+
+        anonymous = Client()
```
giocalitri: you can use the base. LoreTestCase.logout/login functions

rest/tests/test_rest.py
```diff
@@ -0,0 +1,742 @@
...
+                slug=self.repo.slug
+            )
+        )
+        self.assertEqual(FORBIDDEN, resp.status_code)
+
+        client = Client()
+        client.login(
+            username=self.USERNAME_NO_REPO,
+            password=self.PASSWORD
+        )
```
giocalitri, line 456: oh... you are using it... so why are you creating another client?

rest/tests/test_rest.py
```diff
@@ -0,0 +1,742 @@
...
+        resp = client.post(
+            '/api/v1/repositories/', repo_dict
+        )
+        self.assertEqual(FORBIDDEN, resp.status_code)
+
+        resp = anonymous.post(
+            '/api/v1/repositories/', repo_dict
+        )
+        self.assertEqual(FORBIDDEN, resp.status_code)
+
```
giocalitri, line 482: again, why don't you use the functions you defined at the beginning?

rest/urls.py
```diff
@@ -0,0 +1,42 @@
+"""
+URLs for REST application
+"""
+
+from __future__ import unicode_literals
+
+from django.conf.urls import url, include
+
+from .views import (
```
giocalitri, line 9: no strong opinion here, but I dislike relative imports

rest/views.py
```diff
@@ -0,0 +1,134 @@
...
+from rest_framework.generics import (
+    ListCreateAPIView,
+    RetrieveUpdateDestroyAPIView,
+)
+from rest_framework.permissions import (
+    DjangoObjectPermissions,
+    IsAuthenticated,
+)
+from rest_framework.reverse import reverse
+
```
giocalitri: no space here

rest/views.py
```diff
@@ -0,0 +1,134 @@
...
+from rest.serializers import (
+    VocabularySerializer,
+    RepositorySerializer,
+    TermSerializer,
+)
+from learningresources.models import Repository
```
giocalitri, line 26: a space before this

rest/views.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+# pylint: disable=too-many-ancestors
+class RepositoryList(ListCreateAPIView):
+    """REST list view for Repository"""
+    permission_classes = (DjangoObjectPermissions,)
+    queryset = Repository.objects.all()
```
giocalitri: how do you protect this? (it might be in the rest of the code and I haven't seen it)

rest/tests/test_rest.py
noisecapella, line 428: This tests the implicit page=1 behavior. Technically `get_vocabularies` does the same thing but making it explicit makes it easier to see the intention here

rest/tests/test_rest.py
noisecapella: LoreTestCase.logout/login deals with self.client but I want a different client logged in as a different user, or not logged in at all

rest/views.py
```diff
@@ -0,0 +1,134 @@
...
+    lookup_field = 'slug'
+    lookup_url_kwarg = 'vocab_slug'
+    # WARNING: when implement authorization this should be changed
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        """Filter to vocabularies for a repository"""
+        return Repository.objects.get(
+            slug=self.kwargs['repo_slug']
+        ).vocabulary_set.all()
```
giocalitri: again, how do you protect this?

rest/tests/test_rest.py
noisecapella, line 482: These functions all use self.client but I want to test it using an anonymous user

rest/views.py
noisecapella, line 26: `rest` is the new app being created in this PR so this should properly separate our imports from the django/rest_framework imports

rest/views.py
noisecapella: See `DjangoObjectPermissions` one line above this. It enforces object permissions for this view

rest/views.py
noisecapella: The `IsAuthenticated` class says that any authenticated user is allowed to use this view. There are no object-level taxonomy permissions yet AFAIK, but when they are defined we should change that setting. 

We can also set this globally which is probably what we should do once all models have permissions defined

rest/tests/test_rest.py
giocalitri: what does this client differently from the self one?

rest/views.py
giocalitri, line 26: ok

rest/tests/test_rest.py
noisecapella: It is not logged in, it's an anonymous user

giocalitri: yes, but if you call self.logout() at the beginning of your test you have the same behavior (or am I missing something)?

rest/views.py
giocalitri: we need to talk about this

taxonomy/tests/test_api.py
```diff
@@ -169,24 +170,19 @@ def test_create_vocabulary(self):
...
-                                                    9))
+        # IntegrityError from duplicate repo_id and name
+        self.assertRaises(
+            IntegrityError,
+            lambda: create_vocabulary(self.repo.id,
+                                      "vocabulary",
+                                      "description",
+                                      "abc",
+                                      Vocabulary.FREE_TAGGING,
+                                      9))
```
giocalitri, line 182: no strong opinions here but I dislike lambda functions (even if here they make more clear what's going on)

ui/urls.py
```diff
@@ -16,18 +16,22 @@
 
 from __future__ import unicode_literals
 
-from django.conf.urls import include, url
+from django.conf.urls import url, include
```
giocalitri: why did you switch?

rest/tests/test_rest.py
```diff
@@ -0,0 +1,770 @@
...
+from taxonomy.models import Vocabulary, Term
+from rest.serializers import VocabularySerializer
+
+OK = 200
+CREATED = 201
+NO_CONTENT = 204
+BAD_REQUEST = 400
+FORBIDDEN = 403
+NOT_FOUND = 404
+METHOD_NOT_ALLOWED = 405
```
giocalitri: I don't think it's necessary going in this direction, but since I didn't know a better solution, I asked on StackOverflow:
http://stackoverflow.com/questions/30846551/how-to-check-http-status-code-in-python-version-agnostic-way

rest/views.py
giocalitri: There is extra work to be done to protect this: George and I talked and he is on it

Ferdi: What's the plan ?

giocalitri: I guess for now he needs to protect the APIs to limit the results to only what's available/visible by the user.
Then, if you want he can add some decorators to show the right http status errors

learningresources/tests/test_api.py
```diff
@@ -55,6 +57,11 @@ def test_create(self):
             create_course(**self.kwargs)
         self.assertTrue(course_count() == after)
 
+        self.assertRaises(
+            IntegrityError,
+            lambda: create_course(None, self.repo.id, 4, 9, self.user.id)
+        )
```
pwilkins: Question: Why is this an integrity error?  It doesn't jump out at me.

ShawnMilo: Probably because a course with the pk of 4 doesn't exist in the database. I think it's not clear because the test is using magic numbers. It should be more explicit. Also, despite the name, `course_number` is an alphanumeric field.

rest/tests/test_rest.py
```diff
@@ -0,0 +1,770 @@
...
+            ),
+            json.dumps(repo_dict),
+            content_type='application/json',
+        )
+        self.assertEqual(expected_status, resp.status_code)
+        if expected_status == OK:
+            result_dict = as_json(resp)
+            if not skip_assert:
+                for key, value in repo_dict.items():
+                    self.assertEqual(value, result_dict[key])
```
pwilkins: Would it be useful to refactor this repeated code?

learningresources/tests/test_api.py
noisecapella: This is actually a NOT NULL constraint error. I added a comment to make it clearer

rest/tests/test_rest.py
noisecapella: I doubt it. There's an assert for the Location header which is difficult to factor out

learningresources/tests/test_api.py
```diff
@@ -55,6 +57,10 @@ def test_create(self):
             create_course(**self.kwargs)
         self.assertTrue(course_count() == after)
 
+        # Assert that passing None causes a NOT NULL constraint error
+        with self.assertRaises(IntegrityError):
```
ShawnMilo: You can use `with self.assertRaises(IntegrityError) as ex` and check for the NOT NULL message in `ex.args` to make it more explicit, and not need the comment.  [example](https://github.com/mitodl/lore/blob/master/importer/tests/test_import.py#L75)

noisecapella:
To clarify on REST permissions:
- view_repo will allow you to GET a repository and see your collection of repositories
- POST new repositories only allowed for user with add_repo permission
- add_edit_metadata will allow you to PUT or PATCH repository information, including changing its name (which changes the slug). Is this correct?
- DELETE currently not allowed for any permission. Is this correct?

noisecapella:
@giocalitri confirmed all except add_edit_metadata which he said he was not sure about. @Ferdi @pdpinch any insight?

Ferdi:
@noisecapella "add_edit_metadata" is about the learningresource, not repositories. ok ?

noisecapella:
So PUT and PATCH should not be allowed for repositories?

pwilkins:
Please add this block to the bottom of ``docs/api.rst``.  This will include the REST api in the LORE API docs:

``` 
REST
========

.. automodule:: rest.api``
    :members:
    :undoc-members:
    :show-inheritance:
```

pwilkins:
Ignore my last comment.  REST doesn't belong in the LORE API docs.

noisecapella:
I am assuming view_repo will allow people to update repository models, let me know if we want to do things differently

giocalitri:
I don't think view_repo should allow to change anything

noisecapella:
It makes sense but it means we have no way to edit a repository name or description. But that can be dealt with down the line, so I'll make the change

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+
+class RepositoryDetail(RetrieveAPIView):
+    """REST detail view for Repository"""
+    serializer_class = RepositorySerializer
+    lookup_field = 'slug'
+    lookup_url_kwarg = 'repo_slug'
+
+    def get_queryset(self):
+        """Filter to this repository"""
+        repo_ids = [repo.id for repo in get_repos(self.request.user.id)]
```
giocalitri: you should call get_repo and except for PermissionDenied or NotFound

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+        return Repository.objects.filter(
+            slug=self.kwargs['repo_slug']
+        ).filter(id__in=repo_ids)
+
+
+class VocabularyList(ListCreateAPIView):
```
giocalitri, line 74: a POST on this with null value in column "required" violates not-null constraint and gives me an `IntegrityError`

giocalitri, line 74: probably the client does not pass the checkbox as False but as Null

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+    lookup_url_kwarg = 'vocab_slug'
+
+    def get_queryset(self):
+        """Filter to vocabularies for a repository"""
+        repo_ids = [repo.id for repo in get_repos(self.request.user.id)]
+        repos = Repository.objects.filter(
+            slug=self.kwargs['repo_slug']
+        ).filter(id__in=repo_ids)
+        if not repos.exists():
+            return []
```
giocalitri: this is wrong:
if the user does not have permissions on the repo he gets a 200 OK with an empty list: instead he should get a 403

giocalitri: the user without permissions on the repo is able to post a new vocabulary here

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+        return {'Location': url}
+
+
+class VocabularyDetail(RetrieveUpdateAPIView):
+    """REST detail view for Vocabulary"""
+    serializer_class = VocabularySerializer
+    lookup_field = 'slug'
+    lookup_url_kwarg = 'vocab_slug'
+
+    def get_queryset(self):
```
giocalitri, line 102: the user without permissions to view the repo receives a 404 and not a 403 here

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class TermList(ListCreateAPIView):
+    """REST list view for Term"""
+    serializer_class = TermSerializer
+    lookup_field = 'slug'
+    lookup_url_kwarg = 'term_slug'
+
+    def get_queryset(self):
+        """Filter to terms within a vocabulary and repository"""
```
giocalitri, line 118: same problem of user without permissions here: gets 200 with no results instead of 403

giocalitri, line 118: and can create a term

rest/views.py
```diff
@@ -0,0 +1,178 @@
...
+        url = reverse('term-detail', kwargs={
+            'vocab_slug': self.kwargs['vocab_slug'],
+            'repo_slug': self.kwargs['repo_slug'],
+            'term_slug': data['slug'],
+        })
+        return {'Location': url}
+
+
+class TermDetail(RetrieveUpdateAPIView):
+    """REST detail view for Term"""
```
giocalitri, line 140: wrong permissions also here

giocalitri:
I am a bit uncomfortable merging without checking at least for the basic "view_repo" permissions in all the views. 
You can easily  accomplish this using the `get_repo`  function in all of them and then we can say that the proper permissions will be implemented in following PRs.

giocalitri:
also coverage is low: what's missing? 
I think the problem is with 
`rest/tests/base                                   139     10     72     17    87%`

rest/views.py
noisecapella, line 118: Term and Vocabulary validation will be handled in #132

learningresources/api.py
```diff
@@ -128,7 +128,10 @@ def get_repos(user_id):
     Returns:
         repos (query set of learningresource.Repository): Repositories
     """
-    user = User.objects.get(id=user_id)
+    try:
+        user = User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        return []
```
ShawnMilo: Since we have `PermissionDenied` declared above, it would be better to raise it here. If a developer has a bug and is passing a bad value here, it'll silently "fail" but return a value that _looks like_ what they expect. That could cause them to waste a lot of time figuring out what went wrong (or not notice at all and ship the code to production).

learningresources/tests/test_api.py
```diff
@@ -6,11 +6,14 @@
 
 import logging
 
-from learningresources.models import Course, LearningResource
 from learningresources.api import (
-    create_course, get_resource, get_repo,
+    get_repo,
     NotFound, PermissionDenied,
 )
+from django.db.utils import IntegrityError
```
ShawnMilo: Minor: The Django import should be above those within the lore project. The order is:
- standard library
- pip-installed packages
- local code

(With a line of whitespace between those sections.)

learningresources/tests/test_api.py
```diff
@@ -61,6 +64,10 @@ def test_create(self):
             create_course(**self.kwargs)
         self.assertTrue(course_count() == after)
 
+        # NOT NULL constraint fails on org
+        with self.assertRaises(IntegrityError):
```
ShawnMilo: If this line:

``` python
with self.assertRaises(IntegrityError):
```

is changed to this:

``` python
with self.assertRaises(IntegrityError) as ex:
```

it would become possible to check that the error was caused by the org field.

noisecapella: Unfortunately the exception string changes when running against SQLite and against postgres

learningresources/tests/test_api.py
```diff
@@ -83,6 +90,11 @@ def test_get_resource(self):
             )
         )
 
+    def test_invalid_create_repo(self):
+        """Create a repository with empty name and description"""
+        with self.assertRaises(IntegrityError):
```
ShawnMilo: I think we should capture integrity errors at the API level, and add a `ValidationError` to the `learningresources.api` custom exceptions. It may be out of scope for this PR. Any opinions on this @pdpinch @carsongee?

learningresources/tests/test_api.py
ShawnMilo: Do they both contain the string "org"? It's better to match on a `contains` than nothing at all.

When is sqlite3 used? I thought it was always PostgreSQL, in all environments?

learningresources/api.py
noisecapella: Good point

learningresources/tests/test_api.py
```diff
@@ -117,6 +129,11 @@ def test_get_repo(self):
             self.user_norepo.id
         )
 
+    def test_invalid_create_repo(self):
+        """Create a repository with empty name and description"""
+        self.assertRaises(IntegrityError,
+                          lambda: create_repo(None, None, self.user.id))
+
```
ShawnMilo, line 147: Same comments on checking the value of the exception, and not having the API raise integrity errors to the caller of the API. I won't repeat this comment again for other occurrences, so I'm not too annoying and also since a decision needs to be made about adding a `ValidationError`.

lore/settings.py
```diff
@@ -289,6 +291,13 @@ def get_var(name, default):
     },
 }
 
+REST_FRAMEWORK = {
```
ShawnMilo, line 294: I have no problem with this, but does it need to get any of this from an environment variable, @carsongee?

rest/permissions.py
```diff
@@ -0,0 +1,46 @@
...
+from learningresources.models import Repository
+from learningresources.api import (
+    get_repo,
+    PermissionDenied,
+    NotFound,
+)
+
+
+# pylint: disable=protected-access
+class AddRepoPermission(BasePermission):
```
ShawnMilo, line 22: As discussed in chat, this name is misleading. When called via `GET` it returns a list of repositories.

If it actually is checking "add repository" permissions, then the "return True" for any user on line 27 is bad (just because they're authenticated doesn't mean they could ever successfully submit an "add repo" request.

learningresources/tests/test_api.py
pdpinch: It would be nice, but it is out of scope for this PR.

rest/serializers.py
```diff
@@ -0,0 +1,94 @@
...
+            'id',
+            'slug',
+            'name',
+            'description',
+            'create_date',
+            'created_by',
+        )
+        read_only_fields = (
+            'id',
+            'slug',
```
ShawnMilo, line 40: `create_date` should probably be in `read_only_fields` also; it's populated automatically when the record is first created, and should never be modified.

learningresources/tests/test_api.py
ShawnMilo: Okay, thanks. Should I open an issue, or is "it would be nice" a gentle way of saying it's not important?

learningresources/tests/test_api.py
noisecapella: When I use tox it uses sqlite3. I'll see if I can add a check for "NULL" in the string

rest/serializers.py
```diff
@@ -0,0 +1,94 @@
...
+        model = Vocabulary
+        fields = (
+            'id',
+            'slug',
+            'name',
+            'description',
+            'vocabulary_type',
+            'required',
+            'weight',
+            'repository',
```
ShawnMilo, line 65: Should `repository` be in `read_only_fields` as well? It should never be edited.

rest/serializers.py
```diff
@@ -0,0 +1,94 @@
...
+
+    class Meta:
+        # pylint: disable=missing-docstring
+        model = Term
+        fields = (
+            'id',
+            'slug',
+            'label',
+            'weight',
+            'vocabulary',
```
ShawnMilo, line 90: Same comment as above; `vocabulary` should not be changeable for a `Term`.

learningresources/tests/test_api.py
pdpinch: Sorry, I should have been more explicit. It would be nice = "open an issue"

learningresources/tests/test_api.py
ShawnMilo: Checking for "org" in both cases would be better. A `NULL` could appear for any number of reasons.

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+    DEFAULT_TERM_DICT = {
+        'label': 'term label',
+        'weight': 1000,
+    }
+
+    def assert_options_head(self, url, expected_status):
+        """Assert OPTIONS and HEAD"""
+        resp = self.client.options(url)
+        self.assertEqual(expected_status, resp.status_code)
+        as_json(resp)
```
ShawnMilo, line 55: Is this left over from something, or is it just there to guarantee the output was able to be parsed to JSON?

If it's the latter, this isn't a bad way to do it, but please add a comment.

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+        self.assertIn("HEAD", resp['ALLOW'])
+        self.assertIn("OPTIONS", resp['ALLOW'])
```
ShawnMilo, line 57: Is the check for `OPTIONS` redundant, since it had to successfully call `OPTIONS` to get here?

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+        self.assertIn("GET", resp['ALLOW'])
+
+        resp = self.client.head(url)
+        self.assertEqual(expected_status, resp.status_code)
+
+    def get_repositories(self, expected_status=HTTP_200_OK,
+                         skip_options_head_test=False):
+        """Get list of repositories"""
+        url = REPO_BASE
```
ShawnMilo, line 66: If `url` is never changed, why not just use `REPO_BASE` throughout the function?

rest/serializers.py
noisecapella, line 65: It's a `HiddenField` which is a write only field that doesn't take input from the user. There are tests for these fields to confirm that the user can't insert input via those fields.

`repository` is filled in using information from the URL so the information is already available to the user and we don't need to return the information in a GET, so it doesn't make sense to have this field be readable.

rest/tests/base.py
noisecapella, line 66: For consistency with the other `get_` methods in this class

rest/serializers.py
ShawnMilo, line 65: Okay, that all makes sense, except did you mean "unreadable" instead of "readable"? Not nitpicking -- just making sure I get it.

noisecapella, line 65: I mean that in a GET we don't want to return the `repository` field in the JSON, we don't want it to be readable in that way.

ShawnMilo, line 65: Okay. So, the `repository` field _should_ be in `fields` because it's useful, and making it hidden tells the serializer to leave it out of the JSON and not accept input values for it on POST?

lore/settings.py
noisecapella, line 294: We can change this in a separate PR worst case, right? The page size came from @Ferdi. We don't really use `DEFAULT_PERMISSION_CLASSES` since we override it always but I figured it was a good precaution

ShawnMilo, line 294: From my perspective it's fine forever. I just wanted to make sure devops was aware in case this could cause a problem in deployment that you and I wouldn't predict, like my use of a hard-coded index name in Haystack did.

rest/serializers.py
noisecapella, line 65: Yes, but necessary instead of just useful because the field is required to be populated for the model on `save()`

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+            return as_json(resp)
+
+    def create_repository(self, repo_dict=DEFAULT_REPO_DICT,
+                          expected_status=HTTP_201_CREATED,
+                          skip_assert=False):
+        """Helper function to create repository"""
+
+        resp = self.client.post(REPO_BASE, repo_dict)
+        self.assertEqual(expected_status, resp.status_code)
+        if expected_status == HTTP_201_CREATED:
```
ShawnMilo: nit:

This would be clearer as:

``` python
if resp.status_code == HTTP_201_CREATED:
```

Since the `if` only cares if it's a 201, not about `expected_status`.

rest/serializers.py
ShawnMilo, line 65: Okay, cool. Thanks for explaining. I should have said "necessary" -- that would have been more clear.

rest/tests/base.py
noisecapella, line 57: In checking that `OPTIONS` was successful this checks the header to make sure `Allow` contains the methods it should. It's not quite redundant for that reason

ShawnMilo, line 57: Okay. It's really a nit, anyway. If the server was misconfigured in a way that responded to `OPTIONS` while disabling `OPTIONS` in the header then this would find it, and it definitely doesn't hurt anything.

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+        self.assertEqual(expected_status, resp.status_code)
+        if expected_status == HTTP_200_OK:
+            result_dict = as_json(resp)
+            if not skip_assert:
+                for key, value in repo_dict.items():
+                    self.assertEqual(value, result_dict[key])
+            return result_dict
+
+    def put_repository(self, repo_slug, repo_dict,
```
ShawnMilo, line 114: It appears that there are no tests that test this by changing a field in an existing repository, either by calling it twice with different values, or doing a `create_repository` and then changing it with PUT.

rest/tests/base.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+        if not skip_options_head_test:
+            self.assert_options_head(url, expected_status=expected_status)
+
+        resp = self.client.get(url)
+        self.assertEqual(expected_status, resp.status_code)
+        if expected_status == HTTP_200_OK:
+            return as_json(resp)
+
+    def delete_repository(self, repo_slug,
```
ShawnMilo, line 149: There doesn't seem to be a test that checks that this works by getting a count of repos before and after, or by deleting a known repo and then trying (and failing) to retrieve it.

rest/permissions.py
noisecapella, line 22: It's difficult to name because permission classes need to handle both read and write methods. If we wanted separate permission classes for different methods then we can override `get_permissions()` on the view to check `request.method` there instead, but then we're just moving the work. The [Django REST framework docs](http://www.django-rest-framework.org/api-guide/permissions/#custom-permissions) recommend doing it in the permission class although at the moment I'm not sure how reusable these permission classes will be.

We could override `check_permissions` in each view and just do the checking there instead, if it's unlikely we can make reusable permission classes.

ShawnMilo, line 22: Could the `ViewRepoPermission` class below be used to validate view access, freeing up `AddRepoPermission` for checking for insert permission?

noisecapella, line 22: `ViewRepoPermission` checks view access for a particular repository but `AddRepoPermission` is checking for permission to create repositories in general, so they can't be used together.

rest/tests/base.py
noisecapella, line 114: I'll add this to the tests. For the record PATCH and PUT are not allowed for repository currently, #191 may add support for this in the future

rest/tests/base.py
noisecapella, line 149: This is tested, but it's complicated by the fact that deletes aren't currently allowed at all for repositories (see #191). TestRest.test_repositories currently tries the delete and verifies that the count hasn't changed since the delete failed

rest/tests/test_rest.py
```diff
@@ -0,0 +1,570 @@
...
+        """
+        Test for Repository
+        """
+        repositories = self.get_repositories()
+        self.assertEqual(1, repositories['count'])
+
+        self.create_repository()
+        repositories = self.get_repositories()
+        self.assertEqual(2, repositories['count'])
+        new_repo_slug = repositories['results'][1]['slug']
```
ShawnMilo: It doesn't seem safe to assume the repositories are coming back in a fixed order. However, it doesn't look like the following parts really care which repo it is. So should the variable be named `repo_slug`?

noisecapella: The repositories do come back in order by `id`, but I'm fine renaming this to `repo_slug`

rest/tests/test_rest.py
```diff
@@ -0,0 +1,579 @@
...
+
+        # missing fields
+        input_dict = {
+            'name': 'name',
+            'description': 'description',
+        }
+        self.create_vocabulary(self.repo.slug, input_dict,
+                               expected_status=HTTP_400_BAD_REQUEST)
+
+        input_dict = {
```
ShawnMilo: Nit: This looks identical to the `input_dict` above; could have been reused. I know one is used to try to create a vocabulary and the other a repository.

rest/tests/test_rest.py
```diff
@@ -0,0 +1,579 @@
...
+        )
+        self.assertEqual(HTTP_405_METHOD_NOT_ALLOWED, resp.status_code)
+
+        vocab1_dict = self.create_vocabulary(self.repo.slug, vocab1_dict)
+
+        # test duplicate vocabulary
+        self.create_vocabulary(self.repo.slug, vocab1_dict,
+                               expected_status=HTTP_400_BAD_REQUEST)
+
+        vocab2_dict = {
```
ShawnMilo: For clarity, it would be better to re-use `vocab1_dict` from above (maybe change the name to `vocab_dict`), and just set `vocab_dict["name"] = "name 2"` so it's obvious that what's being tested here is a unique name.

Or copy it and change just the name field in the copy, if both `1` and `2` are needed below for more lookups.

rest/tests/test_rest.py
```diff
@@ -0,0 +1,579 @@
...
+        output_dict = self.put_term(
+            self.repo.slug, vocab.slug, new_term_slug, input_dict
+        )
+        new_term_slug = output_dict['slug']
+        self.assertEqual(input_dict['label'],
+                         Term.objects.get(slug=new_term_slug).label)
+        # never created a duplicate
+        self.assertEqual(1, Term.objects.count())
+
+        # test missing repository
```
ShawnMilo: nit: copy/paste typo: `repository` should be `term`.

rest/tests/test_rest.py
```diff
@@ -0,0 +1,579 @@
...
+            self.create_repository(
+                {
+                    'name': "name{i}".format(i=i),
+                    "description": "description"
+                }
+            ) for i in range(40)]
+
+        repositories = self.get_repositories()
+
+        # 100 we created + self.repo
```
ShawnMilo: Typo? `100` should be `40`.

Photo I actually took last week:

https://aoeus.com/code_comments.png

rest/tests/test_rest_authorization.py
```diff
@@ -0,0 +1,476 @@
...
+        # with same name to user
+        self.create_repository(expected_status=HTTP_400_BAD_REQUEST)
+
+        # switch to curator_user
+        self.logout()
+        self.login(self.curator_user.username)
+
+        self.create_repository(expected_status=HTTP_403_FORBIDDEN)
+
+        self.create_repository({
```
ShawnMilo, line 73: Is this test redundant, since the first proves a 403 for that call for that user?

rest/tests/test_rest_authorization.py
```diff
@@ -0,0 +1,476 @@
...
+        self.delete_repository(self.repo.slug,
+                               expected_status=HTTP_403_FORBIDDEN)
+
+        self.logout()
+        self.login(self.curator_user.username)
+
+        # curator does have view_repo but nobody is allowed
+        # to delete repositories
+        self.delete_repository(self.repo.slug,
+                               expected_status=HTTP_405_METHOD_NOT_ALLOWED)
```
ShawnMilo, line 147: I understand it's intentional, but why is this a 405 instead of 403, if the action isn't allowed?

rest/tests/test_rest_authorization.py
```diff
@@ -0,0 +1,476 @@
...
+        # add_repo_user shouldn't be able to see any repositories yet
+        self.get_repository(self.repo.slug, expected_status=HTTP_403_FORBIDDEN)
+        repos = self.get_repositories()
+        self.assertEqual(0, repos['count'])
+
+        # add_repo_user creates a repository
+        new_repo_slug = self.create_repository()['slug']
+
+        self.get_repository(new_repo_slug)
+        self.get_repository(self.repo.slug, expected_status=HTTP_403_FORBIDDEN)
```
ShawnMilo: Is this line an accidental copy/paste of line 170?

rest/tests/test_rest_authorization.py
```diff
@@ -0,0 +1,476 @@
...
+        self.assertEqual(1, self.get_vocabularies(self.repo.slug)['count'])
+
+        # login as author which doesn't have manage_taxonomy permissions
+        self.logout()
+        self.login(self.author_user.username)
+
+        vocab_dict = dict(self.DEFAULT_VOCAB_DICT)
+        vocab_dict['name'] = 'other_name'  # prevent name collision
+
+        # author does not have create access, but we haven't disabled this yet
```
ShawnMilo, line 217: nit, I think.

It seems odd to test for incorrect behavior.

rest/tests/test_rest_authorization.py
```diff
@@ -0,0 +1,476 @@
...
+        self.put_vocabulary(self.repo.slug, vocab_slug,
+                            self.DEFAULT_VOCAB_DICT)
+
+        # login as author which doesn't have manage_taxonomy permissions
+        # but we haven't implemented manage_taxonomy permissions yet so this
+        # will go through
+        self.logout()
+        self.login(self.author_user.username)
+
+        self.patch_vocabulary(self.repo.slug, vocab_slug,
```
ShawnMilo, line 252: Same comment about testing incorrect behavior. There's at least one more below, but I'll stop commenting so I don't get too repetitive.

rest/views.py
```diff
@@ -0,0 +1,157 @@
...
+class RepositoryDetail(RetrieveAPIView):
+    """REST detail view for Repository"""
+    serializer_class = RepositorySerializer
+    lookup_field = 'slug'
+    lookup_url_kwarg = 'repo_slug'
+    permission_classes = (ViewRepoPermission, IsAuthenticated)
+
+    def get_queryset(self):
+        """Filter to this repository"""
+        return Repository.objects.filter(slug=self.kwargs['repo_slug'])
```
ShawnMilo, line 71: If this is only even meant to return a single repository, should it return a single value (using `.get`) instead of a list?

ShawnMilo:
I have some comments. Mostly minor. There are still some outstanding comments from Giovanni that might need to be addressed.

rest/tests/test_rest.py
noisecapella: Yeah, that's an oversight

rest/tests/test_rest.py
noisecapella: `vocab1_dict` is used below for its slug, but I'll change the initialization of `vocab2_dict` to be clearer in intention

rest/tests/test_rest_authorization.py
noisecapella, line 147: The user has view_repo permission to see the repository so a 403 wouldn't make sense here

rest/tests/test_rest_authorization.py
noisecapella, line 217: These tests were copied from my taxonomy authorizations branch so I could get a better picture what the state of taxonomy authorization is in this PR. If these tests break it signals a change of behavior which may be important to us even if it's toward a more correct behavior. In any case they will change assertions in the next PR toward correct behavior.

rest/views.py
noisecapella, line 71: It must return a queryset, even of one item

noisecapella:
@giocalitri has confirmed no extra changes needed

rest/tests/base.py
ShawnMilo, line 114: It looks like this is still missing. Am I just not seeing it?

noisecapella, line 114: I added a `put_respository` call on line 77 in test_rest.py

ShawnMilo, line 114: Yes, but it returns a 405. I meant a test that makes a change and ensures the change happened.

ShawnMilo:
Also, test coverage had a drop. I'm running tox locally to see if there are any easy wins for making that up. I know adding so many lines of code is a factor, but if there are any significant gaps in coverage on the newly-added stuff I may request more tests.

rest/tests/base.py
noisecapella, line 114: PATCH and PUT is not supported for repository at the moment due to lack of permissions, #191 may add support for this in the future. So there can be no test that makes this change.

ShawnMilo, line 114: This would also improve test coverage; there are no tests that actually return a 200 for `patch_repository` and `put_repository`.

ShawnMilo:
All the `if not skip_assert:` lines are only partially covered. Can they be removed entirely? Otherwise they'd need more tests. `rest/tests/base.py` seems to be the biggest culprit in the test coverage drop.

rest/tests/base.py
ShawnMilo, line 114: Okay, that's fair enough. We'll have to temporarily accept the coverage drop until those changes are made. Is there an open issue for that?

noisecapella, line 114: Edit: It's #191

ShawnMilo, line 114: Sorry to be a PITA. I don't see anything in #132 about editing or updating repositories. It looks like #132 is a prerequisite to that. In any case, it looks like there's nothing more to be done before merging this, and the lack of test coverage will always be there to remind us.

noisecapella, line 114: It's actually #191, I made an edit but those don't get emailed out

ShawnMilo, line 114: Cool.

PR #116, opened 2015-06-05
Support running celery in Heroku

### LINKED ISSUE #109: make asynchronous upload work in Heroku

carsongee:
This gets the worker going on Heroku, so it partially resolves #109 , but because they do not share storage with the upload, the worker fails to find the tar.gz and isn't usable until we integrate Django storages.  We can merge this here as is, which would be my preference, and create another story to add django-storages support for course uploads (with a default to local /tmp for dev/docker, and configurable to use S3), or I can solve that issue here.

@pdpinch and @Ferdi, thoughts?

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: uwsgi uwsgi.ini
+worker: celery -A lore worker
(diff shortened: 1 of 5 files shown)


Ferdi:
I agree with you - merge as is.
- Sent from my phone

noisecapella:
Did you want to make any changes regarding the celery backend?

PR #117, opened 2015-06-06
Lore authorization / first part of [#94837982]

giocalitri:
This is the first part of  #94837982: Authorization for lore
For now only the "create a new repository" is fully protected. Any authorization for Curators and Authors have still to be implemented.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,9 @@ node_modules/
 codekit-config.json
 .pycharm_helpers/
 
+.project
+.pydevproject
+
 *.DS_Store
 
 # Django static
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,9 +8,7 @@ import sphinx_bootstrap_theme
 sys.path.insert(0, os.path.abspath('..'))
 
 # django-specific configuration
-from django.conf import settings
-settings.configure()
-
+os.environ['DJANGO_SETTINGS_MODULE'] = 'lore.settings'
 # -- General configuration ------------------------------------------------
 
 extensions = [
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -11,6 +11,9 @@ from django.utils.text import slugify
 from django.utils.encoding import python_2_unicode_compatible
 from django.shortcuts import get_object_or_404
 
+from roles.api import roles_init_new_repo, roles_update_repo
+from roles.permissions import RepoPermission
+
 
 class Course(models.Model):
     """
@@ -79,14 +82,36 @@ class Repository(models.Model):
 
     @transaction.atomic
     def save(self, *args, **kwargs):
-        """Handle slugs."""
+        """Handle slugs and groups"""
+        is_create = False
+        is_update = False
         if self.id is None or self.name != get_object_or_404(
                 Repository, id=self.id).name:
+            # if it is an update of the repository, need the old slug
+            if self.id is not None:
+                is_update = True
+                old_slug = get_object_or_404(Repository, id=self.id).slug
             slug = slugify(self.name)
             count = 1
             while Repository.objects.filter(slug=slug).exists():
                 slug = "{0}{1}".format(slugify(self.name), count)
                 count += 1
             self.slug = slug
+            # check if it's necessary to initialize the permissions
+            if self.id is None:
+                is_create = True
+        new_repository = super(Repository, self).save(*args, **kwargs)
+        if is_create:
+            roles_init_new_repo(self)
+        if is_update:
+            roles_update_repo(self, old_slug)
+        return new_repository
 
-        return super(Repository, self).save(*args, **kwargs)
+    class Meta:
+        # pylint: disable=missing-docstring
+        permissions = (
+            RepoPermission.view_repo,
+            RepoPermission.import_course,
+            RepoPermission.manage_taxonomy,
+            RepoPermission.add_edit_metadata,
+        )
--- a/learningresources/tests/test_views.py
+++ b/learningresources/tests/test_views.py
@@ -4,12 +4,16 @@ Test the importer views to make sure they work.
 
 from __future__ import unicode_literals
 
+import logging
+
 from learningresources.models import Repository
 
 from .base import LoreTestCase
 
 HTTP_OK = 200
 UNAUTHORIZED = 403
+
+log = logging.getLogger(__name__)
 
 
 class TestViews(LoreTestCase):
@@ -26,6 +30,23 @@ class TestViews(LoreTestCase):
         self.assertTrue(resp.status_code == HTTP_OK)
         body = resp.content.decode("utf-8")
         self.assertTrue("<title>MIT - LORE </title>" in body)
+        self.assertTrue('>Create repository</a>' in body)
+
+    def test_get_home_norepo(self):
+        """Home Page with no authorization to create repositories"""
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        resp = self.client.get("/home", follow=True)
+        self.assertTrue(resp.status_code == HTTP_OK)
+        body = resp.content.decode("utf-8")
+        self.assertTrue("<title>MIT - LORE </title>" in body)
+
+        resp = self.client.get("/", follow=True)
+        self.assertTrue(resp.status_code == HTTP_OK)
+        body = resp.content.decode("utf-8")
+        self.assertTrue("<title>MIT - LORE </title>" in body)
+        self.assertFalse('<a href="/lore/create_repo/">'
+                         'Create repository</a>' in body)
 
     def test_create_repo_post(self):
         """Create repo."""
@@ -93,3 +114,4 @@ class TestViews(LoreTestCase):
         self.assertTrue(resp.status_code == HTTP_OK)
         body = resp.content.decode("utf-8")
         self.assertTrue("This field is required." in body)
+        self.assertTrue(Repository.objects.count() == 1)
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,10 @@ django-bootstrap3==5.4.0
 celery==3.1.18
 redis==2.10.3
 
+# django guardian branch that fixed bug in master
+git+https://github.com/mitodl/django-guardian@de7ca5c39dfd0870067497f5843581e9cd2b017f#egg=django-guardian-dev
+
+
 # cas application requirements
 
 git+https://github.com/mingchen/django-cas-ng@69463fe7c23f025be7165c2967f4ceb983d60472#egg=django-cas-ng
--- a/roles/api.py
+++ b/roles/api.py
@@ -1,0 +1,105 @@
+"""
+Functions for handling roles
+"""
+
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Group
+from django.db import transaction
+from guardian.shortcuts import assign_perm
+
+from roles.permissions import RepoPermission, GroupTypes
+
+
+def roles_init_new_repo(repo):
+    """
+    Create new groups for the repository
+
+    It assumes that there are only 3 types of users:
+        - administrator
+        - curator
+        - author
+
+    Args:
+        repo (learningresources.models.Repository): repository used to create
+            groups and assign permissions to them
+    Returns:
+        None
+    """
+    with transaction.atomic():
+        administrator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_administrator.format(repo.slug)
+        )
+        curator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_curator.format(repo.slug)
+        )
+        author_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_author.format(repo.slug)
+        )
+
+    with transaction.atomic():
+        # administrator permissions
+        for permission in RepoPermission.administrator_permissions():
+            assign_perm(permission, administrator_group, repo)
+        # curator permissions
+        for permission in RepoPermission.curator_permissions():
+            assign_perm(permission, curator_group, repo)
+        # author permissions
+        for permission in RepoPermission.author_permissions():
+            assign_perm(permission, author_group, repo)
+
+
+def roles_update_repo(repo, old_slug):
+    """
+    Updates the groups names for the repo
+
+    Args:
+        repo (learningresources.models.Repository): repository used to update
+            groups and assign permissions to them
+        old_slug (unicode): old slug string used to retrieve the groups that
+            need to be renamed
+    Returns:
+        None
+    """
+    # if the slug has not changed there is nothing to do
+    if repo.slug == old_slug:
+        return
+    with transaction.atomic():
+        administrator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_administrator.format(old_slug)
+        )
+        curator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_curator.format(old_slug)
+        )
+        author_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.repo_author.format(old_slug)
+        )
+    administrator_group.name = GroupTypes.repo_administrator.format(repo.slug)
+    curator_group.name = GroupTypes.repo_curator.format(repo.slug)
+    author_group.name = GroupTypes.repo_author.format(repo.slug)
+    with transaction.atomic():
+        administrator_group.save()
+        curator_group.save()
+        author_group.save()
+
+
+def assign_user_to_repo_group(
+        user,
+        repo,
+        group_type):
+    """
+    Assigns an user to a repo specific group type
+
+    Args:
+        user (django.contrib.auth.models.User): user
+        repo (learningresources.models.Repository): repository used to extract
+            the right group to use
+        group_type (roles.permissions.GroupTypes): group string to be used to
+            construct the group name
+    Returns:
+        None
+    """
+    with transaction.atomic():
+        repo_group = Group.objects.get(name=group_type.format(repo.slug))
+        user.groups.add(repo_group)
+        user.save()
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ setenv =
 [testenv:docs]
 envlist = py27
 changedir=docs
+
 deps=
     -r{toxinidir}/doc_requirements.txt
 commands=
--- a/ui/templates/welcome.html
+++ b/ui/templates/welcome.html
@@ -2,8 +2,8 @@
 {% block content %}
     <div class="section-header">
         <div class="row">
-            <div class="col-md-2 col-md-offset-5"><a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a></div>
+            <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
+                                                     class="btn btn-primary btn-large">Create repository</a>{% endif %}</div>
         </div>
     </div>
 
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -7,13 +7,17 @@ from __future__ import unicode_literals
 
 import logging
 
+from django.contrib.auth.models import User
 from django.db import transaction
 from django.shortcuts import get_object_or_404
 from django.http.response import HttpResponseForbidden
 
+from guardian.shortcuts import get_objects_for_user
+
 from learningresources.models import (
     Course, Repository, LearningResource, LearningResourceType
 )
+from roles.permissions import RepoPermission
 
 TYPE_LOOKUP = {}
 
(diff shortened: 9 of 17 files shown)


learningresources/views.py
```diff
@@ -42,6 +52,11 @@ def create_repo(request):
         form = RepositoryForm(data=request.POST)
         if form.is_valid():
             repo = form.save(request.user)
+            assign_user_to_repo_group(
```
Ferdi: Is signals https://docs.djangoproject.com/en/1.8/topics/signals/ a  "better" way to do this ?    /cc @carsongee @ShawnMilo

learningresources/models.py
```diff
@@ -88,5 +92,19 @@ def save(self, *args, **kwargs):
...
 
-        return super(Repository, self).save(*args, **kwargs)
+    class Meta:
+        # pylint: disable=missing-docstring
+        # pylint: disable=old-style-class
+        # pylint: disable=no-init
+        # pylint: disable=too-few-public-methods
+        permissions = (
+            RepoPermission.edit_repo,
+            RepoPermission.use_repo,
```
Ferdi: what's the difference between "use" and "view" ?

giocalitri: I wanted to distinguish the authorization given to the Curator and the one given to the Author.
Initially I did not have a clear idea of how the urls and the user experience would look like. Probably now we do not need 2 different permissions.

learningresources/views.py
ShawnMilo: I'm biased against signals because they're "magical." If you're reading the code to figure out how it works, or trying to track down a bug, you're not getting the full story.

They're useful for things like auditing and attaching behavior to models outside your codebase, but if its our own code, I'd prefer to avoid signals.

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+from guardian.shortcuts import assign_perm
+
+from role.permissions import RepoPermission, GroupTypes
+
+
+def roles_init_new_repo(repo):
+    """
+    Create new groups for the repo
+    """
+    administrator_group = Group.objects.get_or_create(
```
carsongee: You'll need to wrap these in a transaction to prevent race conditions where the groups gets made multiple times.

tox.ini
```diff
@@ -14,6 +14,9 @@ setenv =
 [testenv:docs]
 envlist = py27
 changedir=docs
+setenv =
```
carsongee: I don't think this is needed after #116 merges, but we should test that

role/api.py
giocalitri: I don't think groups can be created multiple times because the name must to be unique, but there is no harm to wrap everything in a transaction anyway.

carsongee: With certain SQL backends, _cough_ MySQL _cough_ we have a lot of experience (several times a day) with that constraint not working in a distributed app tier environment, so it is front of mind.

ShawnMilo: I love [this](https://groups.google.com/forum/#!msg/django-users/AKpy31__Nec/vW0eLf42COcJ) `django-users` post. I keep it bookmarked for times like this.

learningresources/views.py
```diff
@@ -33,6 +37,12 @@ def welcome(request):
 
 
 @login_required
+@permission_required_or_403(
```
pdpinch: In the review meeting, we discussed returning a 404 instead of a 403, for better security. django-guardian doesn't appear to have a comparable permission_required_or_404 decorator. I've opened a new issue for resolving this so we don't block this PR.

pdpinch:
Testing notes from this morning's demo:
1. Create a super-user, e.g. superadmin
2. Create a admin user (e.g. admin1) end grant him the add_repository permission (in the UI)
3. As admin1 create a repo, e.g. repo1. Your user should be member of the repo admin group. 
4. Create a user, e.g. curator1
5. Verify that curator1 cannot create a new repo/cannot access the create repository url)
6. Attempt to access the repo1. Confirm 403 response. 
7. Using /admin ui, make curator1 member of the repo1_curators group
8. Confirm that curator1 can now view repo1 listing

pdpinch:
Aslo, @giocalitri this needs a rebase.

giocalitri:
This afternoon I spent some time talking with Ferdi about this and we realized that I should make some changes to the permission definition.
I also spent some time with Carson analyzing the possible approaches and we found one that should solve the problems with small changes to the code.
This should not be a big problem but I expect to do some work on the tests.

docs/conf.py
```diff
@@ -9,7 +9,9 @@
 
 # django-specific configuration
 from django.conf import settings
-settings.configure()
+
+# Removed for breaking doc building with django-guardian installed.
+# settings.configure()
```
noisecapella: If this is not needed we should remove this line entirely. What error did you get?

ShawnMilo: It blew up when building docs because `ANONYMOUS_USER_ID` did not exist in settings. It _is_ in `lore/settings.py`, but `settings.configure()` blows away the settings with all defaults. I left that line and the comment in just in case someone else knew a better solution to the problem.

giocalitri: This was changed by @ShawnMilo when he fixed the tests problems and I forgot to clean up.

learningresources/models.py
```diff
@@ -88,5 +92,20 @@ def save(self, *args, **kwargs):
...
+            is_create = True
+        new_repository = super(Repository, self).save(*args, **kwargs)
+        if is_create:
+            roles_init_new_repo(self)
+        return new_repository
 
-        return super(Repository, self).save(*args, **kwargs)
+    class Meta:
+        # pylint: disable=missing-docstring
+        # pylint: disable=old-style-class
```
noisecapella: old-style-class, no-init, and too-few-public-methods are now disabled globally so you can remove these lines

docs/conf.py
ShawnMilo: Additional note:

This issue is why [these lines](https://github.com/mitodl/lore/pull/117/files#r31921279) (the environment variables) were added to `tox.ini`.

learningresources/models.py
```diff
@@ -88,5 +92,20 @@ def save(self, *args, **kwargs):
                 slug = "{0}{1}".format(slugify(self.name), count)
                 count += 1
             self.slug = slug
+            is_create = True
```
noisecapella: This isn't strictly true, I think you want to just check self.id is None for this value. The rest of the expression (self.name != get_obj... ) is there to deal with updating slugs if the name changes for an updated repository

learningresources/models.py
giocalitri: ok

learningresources/models.py
giocalitri: You are right.
And with this comment you made me realize that the update of the group name is missing in case the name has changed

learningresources/views.py
```diff
@@ -33,6 +37,13 @@ def welcome(request):
...
+    '{}.add_{}'.format(
```
noisecapella: Is using internal app or model names good practice here? It may be better to just write in the name here since I think it only matters if it's consistent with other permission references

requirements.txt
```diff
@@ -14,6 +14,8 @@ django-appconf==1.0.1
 django-bootstrap3==5.4.0
 celery==3.1.18
 redis==2.10.3
+git+https://github.com/mitodl/django-guardian@app_label_none#egg=django-guardian-dev
```
noisecapella: Use git hash here

noisecapella: Nevermind, that's what app_label_none is

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+
+from django.contrib.auth.models import Group
+from guardian.shortcuts import assign_perm
+
+from role.permissions import RepoPermission, GroupTypes
+
+
+def roles_init_new_repo(repo):
+    """
+    Create new groups for the repo
```
noisecapella: We should specify arguments and return values in these docstrings according to the [napoleon](https://pypi.python.org/pypi/sphinxcontrib-napoleon) style guide

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    administrator_group = Group.objects.get_or_create(
+        name=GroupTypes.repo_administrator.format(repo.slug)
+    )[0]
+    curator_group = Group.objects.get_or_create(
+        name=GroupTypes.repo_curator.format(repo.slug)
+    )[0]
+    author_group = Group.objects.get_or_create(
```
noisecapella: Maybe we can do something like author_group, _ = Group.obj... instead? Tuple unpacking looks cleaner here, no strong opinion on my part though

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+    for permission in RepoPermission.administrator_permissions():
+        assign_perm(permission, administrator_group, repo)
+    # curator permissions
+    for permission in RepoPermission.curator_permissions():
+        assign_perm(permission, curator_group, repo)
+    # author permissions
+    for permission in RepoPermission.author_permissions():
+        assign_perm(permission, author_group, repo)
+
+    return True
```
noisecapella: We should remove this if there's no reason for the caller to use the return value. It looks like this function will either succeed or throw an exception

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def assign_user_to_repo_group(
+        user,
+        repo,
+        group_type=GroupTypes.repo_administrator):
+    """
+    Assigns an user to a repo specific group type
```
noisecapella: Arguments and return value in docstring

learningresources/views.py
giocalitri: In theory you are right, but I hate to put in the code hardcoded strings that are dependent on other code.

role/api.py
giocalitri: I agree

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+    for permission in RepoPermission.author_permissions():
+        assign_perm(permission, author_group, repo)
+
+    return True
+
+
+def assign_user_to_repo_group(
+        user,
+        repo,
+        group_type=GroupTypes.repo_administrator):
```
noisecapella: We may want to make `group_type` not take a default value since a missed parameter by the caller would give a user admin permissions accidentally

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Assigns an user to a repo specific group type
+    """
+    repo_group = Group.objects.get(name=group_type.format(repo.slug))
```
noisecapella: I think we should have a transaction here because of the `save`. Thoughts @cg?

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+    user.groups.add(repo_group)
+    user.save()
+    return True
```
noisecapella: This return value is unnecessary (sorry for the nitpicking, just trying to make everything easier to spot)

role/permissions.py
```diff
@@ -0,0 +1,62 @@
+"""
+Definition of custom permissions
+"""
+
+from __future__ import unicode_literals
+
+
+class GroupTypes(object):  # pylint: disable=too-few-public-methods
```
noisecapella: This pylint comment can be removed now since it's in pylintrc

role/api.py
giocalitri: I already changed

role/permissions.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        return [
+            cls.view_repo[0],
+            cls.import_course[0],
+            cls.manage_taxonomy[0],
+            cls.add_edit_metadata[0],
+        ]
+
+    @classmethod
+    def curator_permissions(cls):
```
noisecapella: Do curators and administrators have the same permissions at the moment? Just making sure

role/api.py
giocalitri: Right: I will remove the default value

learningresources/models.py
noisecapella: Can you add tests for these cases too?

learningresources/tests/test_views.py
```diff
@@ -26,6 +30,23 @@ def test_get_home(self):
         self.assertTrue(resp.status_code == HTTP_OK)
         body = resp.content.decode("utf-8")
         self.assertTrue("<title>MIT - LORE </title>" in body)
+        self.assertTrue('>Create repository</a>' in body)
+
+    def test_get_home_norepo(self):
+        """Home Page with no athorization to create repositories"""
```
noisecapella: Typo in docstring

role/permissions.py
giocalitri: Yes, for now

role/permissions.py
```diff
@@ -0,0 +1,62 @@
...
+    repo_curator = '{}_repo_curators'
+    repo_author = '{}_repo_authors'
+
+
+class RepoPermission(object):
+    """
+    Permissions for the repo objects
+    Django permissions are defined as a tuple of (name, description)
+    """
+    view_repo = ('view_repo', 'Permission to view the repository')
```
noisecapella: Is localization required for the description?

giocalitri: what do you mean?

noisecapella: Are Django permission descriptions ever seen by the outside world? If so we should localize them with ugettext

ui/templates/welcome.html
```diff
@@ -2,8 +2,8 @@
 {% block content %}
     <div class="section-header">
         <div class="row">
-            <div class="col-md-2 col-md-offset-5"><a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a></div>
+            <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
```
noisecapella, line 5: Is the variable `perms` provided globally to templates?

role/api.py
```diff
@@ -0,0 +1,50 @@
...
+
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Group
+from guardian.shortcuts import assign_perm
+
+from role.permissions import RepoPermission, GroupTypes
+
+
+def roles_init_new_repo(repo):
```
Ferdi: I think this function should more to learningresources app

Ferdi:
Should the app be named "role" or "roles" ?

docs/conf.py
pwilkins: @ShawnMilo I added configuration to tox.ini to build the Sphinx documentation as a tox task.  Does that answer the question, or are you asking whether all of these lines are still necessary?

ShawnMilo: @pwilkins I think the only question is whether to delete `settings.configure()` (currently commented out) instead of leaving it commented out with my note. I think we should delete it now that people are aware of it. I just left it there with the comment so others would understand why I cut out some Sphinx boilerplate.

I don't think anyone is asking for anything to be changed, so if so, please mention it.

Ferdi:
Can you document about how to (as a deveoper) use "roles" (sorry if you already have done). E.g. how do I protect a view to be visible only to users with "manage taxonomy" permission ?   How do I do that in a template ?  etc.

docs/conf.py
noisecapella: I'm fine with leaving the comment, usually I just comment on those things to make sure it was an intentional change

pwilkins: Have you run `docker-compose run web tox -e docs` to see if the Sphinx docs still build after this change?

ShawnMilo: @pwilkins; Yes, this was manually tested and shown to fix the error, and also passed Travis.

noisecapella: Just double checked it myself

pwilkins: then simply remove it, if Sphinx builds without that line.  It has no other purpose.

role/permissions.py
pdpinch: localization is a requirement for work on edX, but it's not a high priority for us. We can open an issue and address it later.

docs/conf.py
giocalitri: will remove then

learningresources/models.py
giocalitri: sure

ui/templates/welcome.html
giocalitri, line 5: yes see [here](https://docs.djangoproject.com/en/1.8/topics/auth/default/#permissions)

noisecapella, line 5: Ah, that's convenient

.gitignore
```diff
@@ -66,6 +66,9 @@ node_modules/
 codekit-config.json
 .pycharm_helpers/
 
+.project
+.pydevproject
```
pwilkins, line 70: Check with Carson, but I thought that we were keeping IDE specific gitignore settings in our local global gitignore configurations.

giocalitri, line 70: I just assumed I could add these give that there were some for pycharm as well, but I can remove them

tox.ini
carsongee: Confirmed that this is solved, please remove

.gitignore
pwilkins, line 70: That's just what I thought when I joined the team, too.  Ask Carson whether we have a preference as a team.

carsongee, line 70: That is a suggestion from the git team/authors: http://git-scm.com/docs/gitignore brought up by @bdero to ignore those files globally.

giocalitri, line 70: ok, I'll revert that commit and than we can talk about the rest of the stuff in this gitignore

carsongee, line 70: The default one from github adds tons of that stuff, and then it looks like we snuck in some eclipse stuff.  I say go ahead and leave this in and we can address it all later as an extremely low priority issue lol

requirements.txt
carsongee: I still would prefer the hash of HEAD of that branch so we can still update that branch as ended without breaking existing deployments, but a comment of what branch it is at, would also be helpful

learningresources/models.py
```diff
@@ -79,14 +82,38 @@ def has_resources(self):
...
     def save(self, *args, **kwargs):
-        """Handle slugs."""
+        """Handle slugs and groups"""
+        is_create = False
+        is_update = False
         if self.id is None or self.name != get_object_or_404(
                 Repository, id=self.id).name:
+            # if it is an update of the repository, need the old slug
+            if self.id is not None and \
+                    self.name != get_object_or_404(Repository,
```
noisecapella: Why not just `self.id is not None` here?

carsongee: nit: I actually though \ ticks for line continuation was a pep8 violation: I would prefer

``` python
if (
            self.id is not None and
            self.name != get_object_or_404(
                Repository,
                id=self.id
            ).name
):
```

carsongee: see http://legacy.python.org/dev/peps/pep-0008/#multiline-if-statements for other options

giocalitri: @noisecapella <del>Because I want to update the group names only when there is an update of the repo name and consequently of the slug</del>
Got it

noisecapella: But you've already checked `self.name != get_object_or_404(Repository, id=self.id).name` a few lines above

roles/api.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Create new groups for the repository
+
+    It assumes that there are only 3 types of users:
+        - administrator
+        - curator
+        - author
+
+    Args:
+        repo (learningresources.Models.Repository): repository used to create
```
noisecapella: Models should be models

roles/api.py
```diff
@@ -0,0 +1,105 @@
...
+        for permission in RepoPermission.author_permissions():
+            assign_perm(permission, author_group, repo)
+
+
+def roles_update_repo(repo, old_slug):
+    """
+    Updates the groups names for the repo
+
+    Args:
+        repo (learningresources.Models.Repository): repository used to update
```
noisecapella: Models should be models

roles/api.py
```diff
@@ -0,0 +1,105 @@
...
+def assign_user_to_repo_group(
+        user,
+        repo,
+        group_type):
+    """
+    Assigns an user to a repo specific group type
+
+    Args:
+        user (auth.User): user
+        repo (learningresources.Models.Repository): repository used to extract
```
noisecapella: Models should be models

roles/api.py
```diff
@@ -0,0 +1,105 @@
...
+
+def assign_user_to_repo_group(
+        user,
+        repo,
+        group_type):
+    """
+    Assigns an user to a repo specific group type
+
+    Args:
+        user (auth.User): user
```
noisecapella: Can you change this to either `django.contrib.auth.models.User` or just `User`?

learningresources/models.py
```diff
@@ -79,14 +82,36 @@ def has_resources(self):
...
-        """Handle slugs."""
+        """Handle slugs and groups"""
+        is_create = False
+        is_update = False
         if self.id is None or self.name != get_object_or_404(
                 Repository, id=self.id).name:
+            # if it is an update of the repository, need the old slug
+            if self.id is not None:
+                is_update = True
+                old_slug = self.slug
```
noisecapella: Should probably change this to `old_slug = get_object_or_404(Repository, id=self.id).slug` in case caller has set slug value to something other than what it was in the database

noisecapella:
Looks good, will merge once Travis is ready

pwilkins:
My integration testing of the features was successful.

Issue #118, opened 2015-06-08
Document REST api


Ferdi:
- Documentation using  https://github.com/apiaryio/api-blueprint
- Integrate with https://apiary.io/ ( we don't have to)

http://docs.lore.apiary.io/

pdpinch:
- [X] Merge this draft to master
- [x] Change the branch that's tied to apiary

pdpinch:
@Ferdi how do we change the branch on the apiary docs?

We could set it to master or, in a few days we should have a release branch.

pdpinch:
Ok. Apiary is working off master now. It should update whenever there is a commit.

PR #119, opened 2015-06-08
(WIP) Initial draft for the rest api doc

Ferdi:
#118

See http://docs.lore.apiary.io/

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -6,18 +6,22 @@ HOST: http://lore.mit.edu/
 Simple API on top of MIT LORE application
 
 # URL
-URL Structure is `https://{domain}/api/v1/{resource}/{resourceId}`
+URL Structure is `https://{domain}/api/v1/{resource}/{resource_id}`
 
 | Code   | Description |
 | ---    | --- |
 |`{domain}`    | The domain of the LORE service. `private-b0103d-lore.apiary-mock.com/` is the Apiary mock server. |
 |`/api/v1/`    | The service endpoint includes the current version (v1) |
 |`{resource}`  | The resource endpoint for a specific set of objects in the system. Supported endpoints include repositories, vocabularies, learningresources, members|
-|`{resource_idd}`  |The `resource_id` sets the ID for a specific object to reference |
+|`{resource_id}`  |The `resource_id` sets the unique identifier (name or numerical) for a specific object to reference |
 
 *Example: `http://lore.mit.edu/api/v1/repos` will return a json representation of all repositories.*
 
-*Example: `http://lore.mit.edu/api/v1/repos/123` will return a json representation of the repository with id=123.*
+*Example: `http://lore.mit.edu/api/v1/repos/physics1` will return a json representation of the repository with name "physics1".*
+
+*Example: `http://lore.mit.edu/api/v1/repos/physics1/learningresources/1007` will return a json representation
+of the learning resource with ID= 1007.*
+
 
 # Operations
 LORE API uses the following HTTP verbs
@@ -72,10 +76,10 @@ Page numbering is 1-based. Omitting the ?page parameter will return the first pa
 
 Repository resource
 
-## Repositories [/repos/{repo_id}]
+## Repositories [/repos/{repo_name}]
 
 + Parameters
-    + repo_id: `17` (number, required) - ID of the Repository
+    + repo_name: `physics-intro` (string, required) - name of the repository
 
 ### Retrieve a repository [GET]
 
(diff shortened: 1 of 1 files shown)


apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+
+# URL
+URL Structure is `https://{domain}/api/v1/{resource}/{resourceId}`
+
+| Code   | Description |
+| ---    | --- |
+|`{domain}`    | The domain of the LORE service. `private-b0103d-lore.apiary-mock.com/` is the Apiary mock server. |
+|`/api/v1/`    | The service endpoint includes the current version (v1) |
+|`{resource}`  | The resource endpoint for a specific set of objects in the system. Supported endpoints include repositories, vocabularies, learningresources, members|
+|`{resource_idd}`  |The `resource_id` sets the ID for a specific object to reference |
```
pdpinch: typo: `{resource_id}` instead of `{resource_idd}`

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+### Create a New Repository [POST]
+
+Create a repository by providing its JSON representation
+
++ Request (application/json)
+
+        {
+            "name": "Physics 1",
+            "description": "All intro-to-physics courses"
+        }
```
pdpinch: should we return the slug also?

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+    + Headers
+
+            Location: /repos/123
+
+    + Body
+
+            {
+                "name": "Physics 1",
+                "description": "All intro-to-physics courses"
+            }
```
pdpinch, line 120: Should we return the slug also?

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+|`500`| Server error. |
+
+
+# Authentication
+
+tbd.
+
+# Pagination
+
+Requests that return multiple items all accept a `?page` parameter
```
noisecapella, line 57: How many items per page? Default is 10 for django-rest-framework

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+Page numbering is 1-based. Omitting the ?page parameter will return the first page.
+
+# LORE API Root [/]
+
+## Retrieve the Entry Point [GET]
+
++ Response 200 (application/json)
+
+        {
+            "repositories_url": "/repos"
```
noisecapella, line 70: Providing links like this is a property of HATEOAS, is that what we're going for?

apiary.apib
noisecapella, line 120: I'm not sure it's relevant for a REST api since we're using id numbers in the URL

apiary.apib
Ferdi, line 57: I suggest  ~ 20  for our case

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
++ Response 201 (application/json)
+
+    + Headers
+
+            Location: /repos/123
+
+    + Body
+
+            {
+                "name": "Physics 1",
```
noisecapella: Should include an id number to reference it in the future. I'm not sure that the Location header is useful here

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+
++ Parameters
+    + repo_id: `17` (number, required) - ID of the Repository
+
+### Retrieve a repository [GET]
+
++ Response 200 (application/json)
+
+        {
+            "tbd": "tbd",
```
noisecapella, line 89: This should be the same as the return from 'Create a new repository'

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+            "repositories_url": "/repos"
+        }
+
+
+
+## Group Repositories
+
+Repository resource
+
+## Repositories [/repos/{repo_id}]
```
noisecapella: Just to confirm, we're using /repos/ here instead of /repositories/?

apiary.apib
```diff
@@ -0,0 +1,333 @@
...
+                {
+                    "tbd": "tbd",
+                }
+            ]
+
+## Group Learningobjects
+
+Learning Object resource
+
+## Learningobjects [/repos/{repo_id}/learningobjects/{learningobject_id}]
```
noisecapella, line 145: Should be learningresources

Ferdi:
2 questions for discussions so far:
- Should we use `{id}` or `{slug}`
- Should the name in the url be "repos" or "repositories"

I dont have a strong opinion on either.

carsongee:
I'm a slug fan ala github API (since it is my favorite so far), also github uses repos, but there docs say repositories, but I don't care much on that front

carsongee:
https://developer.github.com/v3/repos/#list-your-repositories

apiary.apib
Ferdi: I think Location header is a sort of requirement on "POST"  http://goo.gl/87Mti0 .  I agree that we should return ID or the full representation of the newly created resource.

Ferdi:
Good idea on following github design when we can. It's one of my favorite apis too.

pdpinch:
@noisecapella what else needs to be addressed before we merge this?

noisecapella:
@pdpinch I haven't done any work on this yet, I've changed it back to work in progress

PR #127, opened 2015-06-08
Use `CELERY_ALWAYS_EAGER` to control use of asynchronous tasks

### LINKED ISSUE #110: refactor upload settings to always use celery
We can use celery for synchronous and asynchronous uploads. If we use celery everywhere, we'll have less code to support.  

For tests, we can set `always_eager = True`

carsongee:
Closes #110
[finished #96309378]


importer/tests/test_views.py
```diff
@@ -52,9 +52,8 @@ def test_upload_duplicate(self):
         self.assertTrue(Course.objects.count() == 0)
         self.upload_test_file()
         self.assertTrue(Course.objects.count() == 1)
-        body = self.upload_test_file()
+        self.upload_test_file()
         self.assertTrue(Course.objects.count() == 1)
```
ShawnMilo, line 56: Why remove the testing for the error message?

carsongee, line 56: Because it no longer is there, even with it running in eager mode, it isn't triggering error messages because it still runs "asynchronously" by returning immediately and the exception is raised outside the form save.  It prioritizes some need for status updates to the user, but that should get cleaned up with the stories coming up in a bit I hope, and it should simplify writing that so we don't have to worry about how to return errors to the user.

carsongee, line 56: I'll go ahead and remove that exception handler as well now too, but it might be worthing talking more about this if you'd like.  It certainly isn't better for the UX until that code is completed

ShawnMilo:
Cool. I'll merge after Travis finishes. I saw briefly that the coverage dropped, but that's a false indicator since more lines of code were removed than added, so the percentage suffers slightly.

carsongee:
The evil build #666 https://travis-ci.org/mitodl/lore/builds/65961399

ShawnMilo:
It's only evil if it fails. :game_die:

carsongee:
Nice work coveralls, "coverage decreased (-0.0%)

ShawnMilo:
Yeah, because of rounding. I bet it was like `-0.02` or something.

PR #141, opened 2015-06-09
Groundwork for search functionality.

### LINKED ISSUE #139: Search: Install search engine and integration with django-haystack
See #120.

ShawnMilo:
Adds Haystack and Elasticsearch to stack.

Adds first search index to LearningResource model.

closes #139

### DIFF
--- a/learningresources/search_indexes.py
+++ b/learningresources/search_indexes.py
@@ -3,19 +3,19 @@ Index information for Haystack.
 
 http://django-haystack.readthedocs.org/en/latest/tutorial.html
 """
-from haystack.indexes import (
-    SearchIndex, Indexable, CharField
-)  # pragma: no cover
 
-from learningresources.models import LearningResource  # pragma: no cover
+from haystack import indexes
+
+from learningresources.models import LearningResource
 
 
-class LearningResourceIndex(SearchIndex, Indexable):  # pragma: no cover
+class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     """
     Index configuration for the LearningResource model.
     """
-    text = CharField(
+    text = indexes.CharField(
         document=True, model_attr="content_xml", use_template=True)
+    resource_type = indexes.CharField(model_attr="learning_resource_type")
 
     def get_model(self):
         """Return the model for which this configures indexing."""
--- a/learningresources/search_indexes.py
+++ b/learningresources/search_indexes.py
@@ -1,0 +1,26 @@
+"""
+Index information for Haystack.
+
+http://django-haystack.readthedocs.org/en/latest/tutorial.html
+"""
+from haystack.indexes import (
+    SearchIndex, Indexable, CharField
+)  # pragma: no cover
+
+from learningresources.models import LearningResource  # pragma: no cover
+
+
+class LearningResourceIndex(SearchIndex, Indexable):  # pragma: no cover
+    """
+    Index configuration for the LearningResource model.
+    """
+    text = CharField(
+        document=True, model_attr="content_xml", use_template=True)
+
+    def get_model(self):
+        """Return the model for which this configures indexing."""
+        return LearningResource
+
+    def index_queryset(self, using=None):
+        """Records to check when updating entire index."""
+        return self.get_model().objects.all()
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -85,6 +85,7 @@ INSTALLED_APPS = (
     'importer',
     'ui',
     'taxonomy',
+    'haystack',
 )
 
 MIDDLEWARE_CLASSES = (
@@ -267,3 +268,15 @@ CELERY_ALWAYS_EAGER = get_var("CELERY_ALWAYS_EAGER", True)
 # guardian specific settings
 ANONYMOUS_USER_ID = None
 GUARDIAN_RAISE_403 = True
+
+# Haystack
+HAYSTACK_CONNECTIONS = {
+    'default': {
+        'ENGINE': (
+            'haystack.backends.elasticsearch_backend'
+            '.ElasticsearchSearchEngine'
+        ),
+        'URL': get_var('HAYSTACK_URL', 'http://127.0.0.1:9200'),
+        'INDEX_NAME': 'haystack'
+    }
+}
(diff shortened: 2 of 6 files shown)


lore/settings.py
```diff
@@ -254,8 +255,21 @@ def get_var(name, default):
 }
 
 # Celery
+USE_CELERY = get_var("USE_CELERY", "True")
```
carsongee: Looks like this snuck back in

learningresources/search_indexes.py
```diff
@@ -0,0 +1,25 @@
...
+from haystack import indexes
+
+from learningresources.models import LearningResource
+
+
+class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
+    """
+    Index configuration for the LearningResource model.
+    """
+    text = indexes.CharField(
```
carsongee, line 16: Is there  a way to include related sets like the vocabularies and such?

lore/settings.py
ShawnMilo: Thanks, fixed.

learningresources/search_indexes.py
ShawnMilo, line 16: There must be, but I don't know how yet. In the next branch (for #142), I've added the `LearningResource.learning_resource_type` to the config, which is a foreign key, and it indexes it fine, despite it being a foreign key. I'm going through the documentation and will figure it out.

The docs seem good content-wise, but the progression is pretty bad. It brings you from zero to 25, then jumps right to 60.

ui/urls.py
```diff
@@ -43,5 +43,6 @@
         r'vocabularies/(?P<vocab_slug>[-\w]+)/$',
         edit_vocabulary,
         name="edit_vocabulary"),
+    url(r'^search/', include('haystack.urls')),
```
Ferdi: so , can we have it be `/repositories/<reposlug>` ?

Ferdi:
I just noticed. I think search, should be a separate app

ui/urls.py
ShawnMilo: The URL is now just the "repositories/slug" URL. We're no longer importing any URLs from Haystack.

(WIP, not yet pushed.)

pdpinch:
Hmm. That `# pragma: no cover` didn't seem to take. 

Aside from the coverage, @carsongee @Ferdi do you guys have any other issues? Can this be merged?

ui/templates/search/indexes/learningresources/learningresource_text.txt
```diff
@@ -0,0 +1,3 @@
+{{ object.title }}
```
Ferdi, line 1: You probably dont need this file to go in this PR

ShawnMilo, line 1: Yeah, it's non-obvious, but it's required by the [Haystack config](https://github.com/mitodl/lore/pull/141/files#diff-1b36f1f0c5b2773e5fe7a4cfdb525190R20).

Ferdi, line 1: Sorry, it was my misunderstanding.  I thought it was the template to show the search result.

learningresources/search_indexes.py
```diff
@@ -0,0 +1,26 @@
+"""
+Index information for Haystack.
+
+http://django-haystack.readthedocs.org/en/latest/tutorial.html
+"""
+from haystack.indexes \
+    import SearchIndex, Indexable, CharField  # pragma: no cover
```
carsongee: I would prefer to remove all these no coverage comments and just let the coverage score go down.  These will all likely get covered as soon as we have more code that exercises them anyway and it will be worth knowing that

carsongee: Also, a small style preference, feel free to ignore, but I like

``` python
from haystack.indexes import (
    SearchIndex, Indexable, CharField
)
```

or

``` python
from haystack.indexes import (
    SearchIndex,
    Indexable,
    CharField
)
```

learningresources/search_indexes.py
```diff
@@ -0,0 +1,26 @@
...
+
+from learningresources.models import LearningResource  # pragma: no cover
+
+
+class LearningResourceIndex(SearchIndex, Indexable):  # pragma: no cover
+    """
+    Index configuration for the LearningResource model.
+    """
+    # This is a config file for Haystack; nothing to test.
```
carsongee: nit: I don't think this comment is necessary. The statement here is also true for model classes as long as they don't have methods, but we don't explicitly call that out, and these classes typically get "integration" testing coverage when views or apis are tested.

learningresources/search_indexes.py
ShawnMilo: I agree with you on these things. I did them this way because:
- The coverage score is zero. As a config file it doesn't get covered at all.
- Writing the imports as "one line" was done because the `pragma` tag has no scope larger than one line.

I'll fix the imports and put the `pragma` on the first or last line and see if coverage respects it.

learningresources/search_indexes.py
ShawnMilo: Yep, you're right. I had a `skip` line in there which has since been removed, and the comment was to cover it. I'll fix that.

learningresources/search_indexes.py
carsongee: >  As a config file it doesn't get covered at all.

I'm confused by this, If the config file is never imported by python, than it isn't being used, and shouldn't be included... For example both `cas/conf.py` and `lore/settings.py` are covered almost entirely simply by being there, since config python files still show as covered when the project starts up...  Won't this show as covered as soon as we have a view or something that uses haystack? Or a test for a signal or something?  What happens if you start up `coverage run ipython` and import and do some stuff with Haystack?  It should show this as covered I would think.

ShawnMilo: I don't know, but the tests do use Haystack and it gets 0% coverage. I'll remove the `pragma` statements and check again to be sure, though.

ShawnMilo:
@cg: I pushed a version with the `pragma` statements removed. It appears that you're right -- the coverage is fine in the next branch which has a view that uses Haystack, so the coverage may suffer with this commit but it will be temporary. I was confused because of working with each of these branches separately yesterday. Sorry for confusing you along with myself.

carsongee:
looks good :rooster:

PR #145, opened 2015-06-09
Switch to using django storage instead of raw os calls

### LINKED ISSUE #109: make asynchronous upload work in Heroku

carsongee:
Add S3 support for storing files
Closes #109
[Finishes #96309384]


docker-compose.yml
```diff
@@ -34,7 +34,7 @@ web:
     - db
     - redis
 celery:
-  build: .
+  image: lore_web
```
pwilkins, line 39: Question: I'm curious about what this change does.  Did Celery just build everything before, and now it's using an existing image?

importer/tests/test_import.py
```diff
@@ -7,15 +7,44 @@
...
+    else:
+        _, ext = os.path.splitext(path)
+
+    return default_storage.save(
+        '{prefix}/{random}{ext}'.format(
+            prefix=settings.IMPORT_PATH_PREFIX,
+            random=str(uuid.uuid4()),
+            ext=ext
+        ),
+        open(path, 'rb')
```
pwilkins: Question: I'm surprised that `open()` doesn't need a `try/except` block.  Must be a Python thing.

importer/tests/test_import.py
```diff
@@ -7,15 +7,44 @@
...
+
 from importer.api import import_course_from_file
 from importer.tasks import import_file
 from learningresources.models import LearningResource, Course
 from learningresources.tests.base import LoreTestCase
 
 
+def copy_file(path):
+    """
+    Copy given file into django default_storage
```
pwilkins: nit: missing sentence period.

docker-compose.yml
carsongee, line 39: Yeah, it just cuts the build time in half

pwilkins:
This is outside the scope of this PR, but `__init__.py/import_course_from_path()` needs a `Returns:` in its docstring.

importer/tests/test_import.py
carsongee: Well we get to be a bit lazy in tests since we are using input from our own repo and want uncaught exceptions if the file isn't there so the tests fail.

noisecapella:
@pwilkins That docstring is updated in master

lore/settings.py
```diff
@@ -169,6 +171,32 @@ def get_var(name, default):
...
+AWS_STORAGE_BUCKET_NAME = get_var('AWS_STORAGE_BUCKET_NAME', False)
+# Provide nice validation of the configuration
+if(
+        LORE_USE_S3 and
+        (not AWS_ACCESS_KEY_ID or
+         not AWS_SECRET_ACCESS_KEY or
+         not AWS_STORAGE_BUCKET_NAME)
+):
+    raise ImproperlyConfigured(
+        'You have enabled S3 support, but are missing either '
```
pwilkins: Nitty Nit: Common English usage has "either" used with multiple choices, but for people who have nothing better to do with their time than argue about such things, "either" is a choice between two options.  Absolutely no one will care either way.

ui/forms.py
```diff
@@ -59,14 +60,15 @@ def save(self, user_id, repo_id):
...
         # Save the uploaded file into a temp file.
-        handle, filename = mkstemp(suffix=uploaded_file.ext)
-        os.close(handle)
-
-        with open(filename, 'wb') as temp:
-            for chunk in uploaded_file:
-                temp.write(chunk)
-
-        import_file.delay(filename, repo_id, user_id)
+        path = default_storage.save(
```
pwilkins, line 63: Is there a limit on the size of the uploaded file?

importer/tests/test_import.py
pwilkins: Ah.  I'm still thinking from a Java perspective where leaving out a try/catch block gets you a dope slap from the compiler.  Now I see that in the tests what you did makes perfect sense.

pwilkins:
I was able to successfully run the tests with S3 storage enabled as well as disabled.

ui/forms.py
carsongee, line 63: Not imposed by us, just by disk space and server timeout

carsongee:
I have added two new commits that should stay as separate commits.  They are new issues discovered as a result of these changes.  We were having a lot of leftover test files in our bucket after running tests.  So the first commit refactors the way we create test zip files into the test base class, and the second adds a try finally construct to the upload function to ensure no matter what happens the uploaded file and temporary files are deleted to prevent filling up disks with old archive files.

PR #148, opened 2015-06-11
Import course protected [#96516342]

### LINKED ISSUE #131: Add authorizations to the import form
Repository admins and curators can import courses. 

Authors cannot

giocalitri:
New APIs, tests for APIs and tests for protected import course

Created new apis for roles and learningresources: the idea is that to protect other sections we need to always check if the current object the user is working on belongs to a repo he can have access

closes #131

### DIFF
--- a/importer/tests/test_views.py
+++ b/importer/tests/test_views.py
@@ -8,11 +8,14 @@ import logging
 
 from learningresources.models import LearningResource, Course
 from learningresources.tests.base import LoreTestCase
+from roles.api import assign_user_to_repo_group, remove_user_from_repo_group
+from roles.api import GroupTypes
 
 from .test_import import get_course_zip
 
 HTTP_OK = 200
 UNAUTHORIZED = 403
+NOT_FOUND = 404
 
 log = logging.getLogger(__name__)
 
@@ -20,17 +23,106 @@ log = logging.getLogger(__name__)
 class TestViews(LoreTestCase):
     """Hit each view."""
 
+    def setUp(self):
+        super(TestViews, self).setUp()
+        self.import_url = "/repositories/{0}/import/"
+        self.import_url_slug = self.import_url.format(self.repo.slug)
+        self.repository_url = "/repositories/{0}/"
+        self.repository_url_slug = self.repository_url.format(self.repo.slug)
+
     def test_upload_get(self):
         """GET upload page."""
-        resp = self.client.get("/repositories/{0}/import".format(
-            self.repo.slug), follow=True)
+        resp = self.client.get(self.import_url_slug, follow=True)
         body = resp.content.decode("utf-8")
         self.assertTrue('enctype="multipart/form-data"' in body)
 
     def test_upload_get_bad_repo(self):
         """GET upload page."""
-        resp = self.client.get("/repositories/999/import", follow=True)
-        self.assertTrue(resp.status_code == UNAUTHORIZED)
+        self.assert_status_code(self.import_url.format("999"), NOT_FOUND)
+
+    def test_upload_with_permissions(self):
+        """GET upload page with different user permissions"""
+        def check_user_no_permission():
+            """
+            Helper function to check that the user has
+            no permission on the repo
+            """
+            # user cannot see the repository page
+            self.assert_status_code(
+                self.repository_url_slug,
+                UNAUTHORIZED
+            )
+            # and cannot see the import page
+            self.assert_status_code(
+                self.import_url_slug,
+                UNAUTHORIZED
+            )
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        # user has no permissions at all
+        check_user_no_permission()
+        # user has author permissions
+        assign_user_to_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_AUTHOR
+        )
+        # user can see the repository page
+        self.assert_status_code(
+            self.repository_url_slug,
+            HTTP_OK
+        )
+        # but cannot see the import for the repo
+        self.assert_status_code(
+            self.import_url_slug,
+            UNAUTHORIZED
+        )
+        # user has no permissions
+        remove_user_from_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_AUTHOR
+        )
+        check_user_no_permission()
+        # user has curator permissions
+        assign_user_to_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_CURATOR
+        )
+        # user can see the repository page
+        self.assert_status_code(
+            self.repository_url_slug,
+            HTTP_OK
+        )
+        # and can see the the import for the repo
+        self.assert_status_code(
+            self.import_url_slug,
+            HTTP_OK
+        )
+        # remove curator permissions
+        remove_user_from_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_CURATOR
+        )
+        check_user_no_permission()
+        # user has admin permissions
+        assign_user_to_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_ADMINISTRATOR
+        )
+        # user can see the repository page
+        self.assert_status_code(
+            self.repository_url_slug,
+            HTTP_OK
+        )
+        # and can see the the import for the repo
+        self.assert_status_code(
+            self.import_url_slug,
+            HTTP_OK
+        )
 
     def test_upload_post(self):
         """POST upload page."""
@@ -52,7 +144,7 @@ class TestViews(LoreTestCase):
         """Used multiple times in tests"""
         with open(get_course_zip(), "rb") as post_file:
             resp = self.client.post(
-                "/repositories/{0}/import/".format(self.repo.slug),
+                self.import_url_slug,
                 {"course_file": post_file, "repository": self.repo.id},
                 follow=True
             )
@@ -61,7 +153,7 @@ class TestViews(LoreTestCase):
     def test_invalid_form(self):
         """Upload invalid form"""
         resp = self.client.post(
-            "/repositories/{0}/import/".format(self.repo.slug),
+            self.import_url_slug,
             {}, follow=True
         )
         self.assertTrue(resp.status_code == HTTP_OK)
--- a/learningresources/tests/base.py
+++ b/learningresources/tests/base.py
@@ -35,6 +35,13 @@ class LoreTestCase(TestCase):
         """just a client logout"""
         self.client.logout()
 
+    def assert_status_code(self, url, code, return_body=False):
+        """Asserts the status code of GET attempt"""
+        resp = self.client.get(url, follow=True)
+        self.assertTrue(resp.status_code == code)
+        if return_body:
+            return resp.content.decode("utf-8")
+
     def setUp(self):
         """set up"""
         super(LoreTestCase, self).setUp()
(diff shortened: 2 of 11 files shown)


importer/tests/test_views.py
```diff
@@ -30,7 +33,53 @@ def test_upload_get(self):
...
         """GET upload page."""
         resp = self.client.get("/repositories/999/import", follow=True)
+        self.assertTrue(resp.status_code == NOT_FOUND)
+
+    def test_upload_with_permissions(self):
+        """GET upload page with different user permissions"""
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        # user has no permissions at all
+        resp = self.client.get("/repositories/{0}/import".format(
```
ShawnMilo: Since the same URL is used multiple times in the test, maybe make it `import_url` and re-use it for clarity.

importer/tests/test_views.py
```diff
@@ -30,7 +33,53 @@ def test_upload_get(self):
...
         self.assertTrue(resp.status_code == UNAUTHORIZED)
+        # user has author permissions and cannot see the import for the repo
+        assign_user_to_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.repo_author
+        )
+        resp = self.client.get("/repositories/{0}/import".format(
+            self.repo.slug), follow=True)
+        self.assertTrue(resp.status_code == UNAUTHORIZED)
```
ShawnMilo: I know this is a test of imports and not the repository listing, but it would add clarity to test for a 403 on the repo view page, then test for a 200 (but still a 403 on the import), just to show that the permissions are working properly.

importer/tests/test_views.py
```diff
@@ -30,7 +33,53 @@ def test_upload_get(self):
...
+            self.user_norepo,
+            self.repo,
+            GroupTypes.repo_curator
+        )
+        assign_user_to_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.repo_administrator
+        )
+        resp = self.client.get("/repositories/{0}/import".format(
```
ShawnMilo: It would be nice to see it 403 again between `HTTP_OK` as curator and then `HTTP_OK` as admin, just so we know something actually changed.

Since that's being done so many times already in this test, maybe create a local function to do it.

``` python
def assert_status_code(url, code):
    """Asserts the status code of GET attempt"""
    resp = self.client.get(url, follow=True)
    self.assertTrue(resp.status_code == code)
```

importer/tests/test_views.py
giocalitri: I added this test in the repo listing

giocalitri: But I see your point

ShawnMilo: Yeah, it's just more explicit, even if it is a bit of duplication.

learningresources/tests/test_api.py
```diff
@@ -78,6 +83,28 @@ def test_get_resource(self):
...
 
 
+class TestRepoAPI(LoreTestCase):
+    """
+    Tests repo getters
+    """
+    def test_get_repo_or_error(self):
+        """repo does not exist"""
+        # this should not fail
+        _ = get_repo_or_error(self.repo.slug, self.user.id)
```
ShawnMilo: I think the `_ =` is not necessary here.

learningresources/tests/test_api.py
```diff
@@ -78,6 +83,28 @@ def test_get_resource(self):
...
+        self.assertRaises(
+            Http404,
+            get_repo_or_error,
+            "unexistent_repo",
```
ShawnMilo: Microscopic nit; "nonexistent" is a slightly more correct.

learningresources/tests/test_views.py
```diff
@@ -68,6 +70,58 @@ def test_listing_unauthorized(self):
         self.assertTrue(resp.status_code == UNAUTHORIZED)
         self.assertTrue("unauthorized" in body)
 
+    def test_listing_importcourse_perms(self):
+        """
```
ShawnMilo, line 72: This test might be easier to follow with a similar function to the one mentioned above. Maybe even make it part of the class.

ui/views.py
```diff
@@ -117,13 +118,21 @@ def edit_vocabulary(request, repo_slug, vocab_slug):
 
 
 @login_required
+@permission_required_or_403(
```
ShawnMilo, line 121: What happens if `@login_required` is removed? Would the behavior be the same?

ui/views.py
```diff
@@ -117,13 +118,21 @@ def edit_vocabulary(request, repo_slug, vocab_slug):
...
+    # pylint: disable=protected-access
+    # the following string is "learningresources.import_course"
+    # (unless the Repository model has been moved)
+    '{}.{}'.format(
+        Repository._meta.app_label,
+        RepoPermission.import_course[0]
+    ),
+    (Repository, 'slug', 'repo_slug')
+)
```
ShawnMilo, line 130: Is there a way to make these new decorators into one-liners? Maybe define them in another file?

ui/views.py
giocalitri, line 121: I haven't tried, but I think that the only difference is that the user will not be redirected to the login page but s/he will receive a 403 all the times because the anonymous user does not have any permission

ui/views.py
giocalitri, line 130: meaning pre-initialize them in another view-specific decorator?

Ferdi, line 130: @giocalitri we talked briefly about this and you have a good reason, but it would be great to find a way to make this a bit more readable.

giocalitri, line 130: @Ferdi: we agreed on a comment that describes what it means for the permission string

For the rest I don't think there is another or clearer way to call the api

ShawnMilo, line 130: @giocalitri: I didn't know they had to be initialized per-view. That makes it harder.

ui/views.py
ShawnMilo, line 121: Okay, just an idle thought since if they're not logged in they can't have any permissions. Don't worry about it. If it'll return different status codes it's probably not worth it.

ui/views.py
Ferdi, line 130: @giocalitri ok

importer/tests/test_views.py
```diff
@@ -30,7 +33,53 @@ def test_upload_get(self):
...
+            GroupTypes.repo_author
+        )
+        resp = self.client.get("/repositories/{0}/import".format(
+            self.repo.slug), follow=True)
+        self.assertTrue(resp.status_code == UNAUTHORIZED)
+        # user has curator permissions and can see the the import for the repo
+        remove_user_from_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.repo_author
```
ShawnMilo: Should the `GroupTypes` `repo_author`, etc., be `REPO_AUTHOR`, etc.? I believe most package constants are like that, such as `logging.DEBUG`.

importer/tests/test_views.py
```diff
@@ -20,17 +23,113 @@
 class TestViews(LoreTestCase):
     """Hit each view."""
 
+    import_url = "/repositories/{0}/import/"
+    repository_url = "/repositories/{0}/"
```
ShawnMilo: @carsongee: Based on our previous conversation about me nit-picking this, do you like these hanging out here in the class namespace instead of a `setUp` override, since it's less code? If so, I'll ignore it.

ShawnMilo: Actually, I take that back. @giocalitri, I think this should definitely be in a `setUp` override, mostly because then `self.repo` is in scope, and the URL can have `self.repo.slug` only added once, rather than passed in many times  during the test.

giocalitri: I can move it to the setUp: I actually intended to do it and forgot

giocalitri: the strings are used with at least 2 different slugs, so I guess I still need to format it

importer/tests/test_views.py
```diff
@@ -20,17 +23,113 @@
...
+            self.import_url.format(self.repo.slug),
+            HTTP_OK
+        )
+        # remove curator permissions
+        remove_user_from_repo_group(
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_CURATOR
+        )
+        # User cannot see anything
```
ShawnMilo: This check for the user unable to see anything is in this test three times. I think it would be cleaner if it was declared as a function _within_ `test_upload_with_permissions`, and just called three times.

learningresources/tests/test_views.py
```diff
@@ -19,16 +21,14 @@
 class TestViews(LoreTestCase):
     """Hit each view."""
 
+    repository_url = "/repositories/{0}/"
```
ShawnMilo: Same comment here as above, about putting it in setup and having `self.repo.slug` in there once.

learningresources/tests/base.py
```diff
@@ -35,6 +35,13 @@ def logout(self):
         """just a client logout"""
         self.client.logout()
 
+    def assert_status_code(self, url, code, return_body=False):
+        """Asserts the status code of GET attempt"""
+        resp = self.client.get(url, follow=True)
+        self.assertTrue(resp.status_code == code)
+        if return_body:
```
ShawnMilo, line 42: This isn't necessary; it's okay to always return the body and let the caller decide whether to ignore it. That doesn't break pep8 or pylint, right?

giocalitri, line 42: I did not want to do it because the function is called `assert_status_code` and the body is not about that.
In this way it makes more clear when I want the body or not

learningresources/tests/test_views.py
giocalitri: here I'll do it

learningresources/tests/test_views.py
```diff
@@ -63,11 +59,82 @@ def test_create_repo_post(self):
...
+            self.user_norepo,
+            self.repo,
+            GroupTypes.REPO_CURATOR
+        )
+        body = self.assert_status_code(
+            self.repository_url.format(self.repo.slug),
+            HTTP_OK,
+            return_body=True
+        )
+        self.assertTrue("Import Course</a>" in body)
```
ShawnMilo, line 116: For tests which check for the ability to see the import page, a better check is to hit the import page and get denied, rather than hit the view page and look for text.

I know the display of that link is dependent upon the same permission setting that allows view of the page, but it still seems better to test the actual thing.

importer/tests/test_views.py
giocalitri: added an extra `self.import_url_slug`

learningresources/tests/test_views.py
giocalitri, line 116: I hate to look for text in the page, but since the way I check permissions in the template is slightly different, I cannot assume this test works if the user cannot check the url.

I did this in the past using a library to load the body and then looking for the tag via id.
I guess I can look into this if you think it's worth

giocalitri, line 116: And with library I mean [Beautiful Soup](http://www.crummy.com/software/BeautifulSoup/bs4/doc/)

ShawnMilo, line 116: Okay, never mind. I see that `test_upload_with_permissions` _does_ test for a 403 being returned when an invalid attempt is made. I understand now that _this_ test is _meant_ to check for the visibility of the button.

learningresources/tests/base.py
ShawnMilo, line 42: True. I'd rather get rid of the `if` and always return it, but not badly enough to figure out the perfect renaming now. Maybe in a future PR if I happen to be messing with the tests.

ShawnMilo:
Will merge after Travis passes.

PR #154, opened 2015-06-12
try adding app config for search

ShawnMilo:
@carsongee: Help request, since you've done this before.

I'm trying to move the Haystack and Elasticsearch settings out of `lore/settings.py`, but it seems to break initializing the app:

```
  File "/src/.tox/py34/lib/python3.4/site-packages/haystack/__init__.py", line 34, in <module>
    raise ImproperlyConfigured('The HAYSTACK_CONNECTIONS setting is required.')
django.core.exceptions.ImproperlyConfigured: The HAYSTACK_CONNECTIONS setting is required.
```

I tried moving `search` to the top of `INSTALLED_APPS` (under `cas`), but the same thing happens.

Am I doing something wrong, or do we need to leave Haystack settings in `lore/settings.py`?


carsongee:
You have to add a models.py that calls it, see cas for how to do it and the hack comment in there

carsongee:
You shouldn't have to add it high in the list because it doesn't have other app dependencies, CAS has that because it needs to load before auth and admin.  Haystack shouldn't have that same requirement

search/conf.py
```diff
@@ -0,0 +1,29 @@
+"""search application configuration"""
+from appconf import AppConf
+from django.conf import settings
+
+
+class SearchAppConf(AppConf):
+    """Configuration class for a configurable CAS authentication system"""
+
+    # ENABLED = False
```
carsongee: I think you probably want to call this HAYSTACK_URL, that way you can still set it with get_var in settings.py via the `SEARCH_HAYSTACK_URL` setting (that prefix gets added by AppConf). You can then just use its value in the connection dictionary below

ShawnMilo: I don't understand what "this" means here (the "this" I want to call `HAYSTACK_URL`).

carsongee: The configuration variable, of which you haven't added one yet, but commented out one: http://django-appconf.readthedocs.org/en/latest/ for docs.  That may help you understand, if it doesn't feel free to hit me up on chat or something.

ShawnMilo:
@carsongee: I'm stuck. Every time I try something, I end up with this:

``` bash
  File "/src/.tox/py34/lib/python3.4/site-packages/haystack/__init__.py", line 34, in <module>
    raise ImproperlyConfigured('The HAYSTACK_CONNECTIONS setting is required.')
django.core.exceptions.ImproperlyConfigured: The HAYSTACK_CONNECTIONS setting is required.
```

I've tried:
- duplicating the conf.py from cas
- using Django's own [AppConfig](https://docs.djangoproject.com/en/1.8/ref/applications/#application-configuration) class and changing `lore/settings.py` to have that in `INSTALLED_APPS`
- putting the `django-appconf` into `models.py` directly
- importing the config in `search/__init__.py`
- moving those same pieces around in different ways
- using the 'Meta' class from `django-appconf'
- spelling the settings with and without the `HAYSTACK_` prefix
- importing `django.conf.settings` and attempting to assign the values to `settings`
- using and not using the `prefix` setting in the appconf `Meta`.

However, as you can see from the `1/0` lines, the config (neither Django's nor appconf's) are run before Haystack tries to load, despite the fact that I've moved `search` above `haystack` in `INSTALLED_APPS`.

carsongee:
Can you rebase this on master?  There are all sorts of extra commits showing in the github diff

carsongee:
Have you tried mixing it with Django's AppConfig: https://docs.djangoproject.com/en/dev/ref/applications/#configuring-applications

it has some startup signals and such that should allow you to configure potentially before haystack is imported.  It is really tough that they are doing their settings check on `__init__.py` though.

ShawnMilo:
Sure, I'll rebase it. Yes, I tried using Django's AppConfig, as described [above](https://github.com/mitodl/lore/pull/154#issuecomment-111629780).

carsongee:
Sorry, missed that

carsongee:
Now the one commit has all the changes in it...

ShawnMilo:
Yeah, I'm struggling with the rebase. I may just create a new PR based on `master`, since this is on a long-ago merged branch.

ShawnMilo:
I'll get back to this in a bit; I want to finish reviewing #183 first because that's higher priority.

ShawnMilo:
The branch has become hopelessly conflicted. Closing; will open a new PR with a fresh branch against `master`.

PR #171, opened 2015-06-15
Make ElasticSearch Index Configurable

carsongee:
This will allow us to reuse the same elastic search cluster for multiple deployments by changing the index name.

### DIFF
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -308,7 +308,7 @@ HAYSTACK_CONNECTIONS = {
             '.ElasticsearchSearchEngine'
         ),
         'URL': get_var('HAYSTACK_URL', 'http://127.0.0.1:9200'),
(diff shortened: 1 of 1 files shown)


ShawnMilo:
Will merge when Travis succeeds.

lore/settings.py
```diff
@@ -308,7 +308,7 @@ def get_var(name, default):
             '.ElasticsearchSearchEngine'
         ),
         'URL': get_var('HAYSTACK_URL', 'http://127.0.0.1:9200'),
-        'INDEX_NAME': 'haystack'
+        'INDEX_NAME': get_var('HAYSTACK_INDEX', 'haystack'),
```
Ferdi, line 311: should the index name be something like "lore_production"  ?

PR #173, opened 2015-06-15
Properly handle export when things go wrong.

### LINKED ISSUE #172: Export failing
@ShawnMilo or @giocalitri ?

```
Internal Server Error: /learningresources/1/
Traceback (most recent call last):
  File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "./ui/views.py", line 244, in export
    get_resource(resource_id, request.user.id).content_xml,
AttributeError: type object 'HttpResponseForbidden' has no attribute 'content_xml'
```

ShawnMilo:
If the user has no permission or the resource doesn't exist, return
a 404.

Note: Although the `export` view previously appeared to have no
permissions checking, it did because the `get_resource` API function
would raise a 403 response for bad permissions.

This commit improves the codebase by moving HTTP response raising from
the API to the views, as is proper.

closes #172 

Reading order suggestion:
1. The primary work was done in `learningresources/tests/test_views.py`, so that should be read first
2. Objects for testing were added to `learningresources/tests/base.py`
3. The actual fix is in `learningresources/api.py` and `ui/views.py`
4. The rest are updates to the tests after adding the testing objects to `base.py`

### DIFF
--- a/learningresources/tests/test_views.py
+++ b/learningresources/tests/test_views.py
@@ -6,6 +6,8 @@ from __future__ import unicode_literals
 
 import logging
 
+from django.core.urlresolvers import reverse
+
 from learningresources.models import Repository
 from roles.api import assign_user_to_repo_group, remove_user_from_repo_group
 from roles.permissions import GroupTypes
@@ -14,6 +16,7 @@ from .base import LoreTestCase
 
 HTTP_OK = 200
 UNAUTHORIZED = 403
+NOT_FOUND = 404
 
 log = logging.getLogger(__name__)
 
@@ -184,3 +187,25 @@ class TestViews(LoreTestCase):
         body = resp.content.decode("utf-8")
         self.assertTrue("This field is required." in body)
         self.assertTrue(Repository.objects.count() == 1)
+
+    def test_export_good(self):
+        """Get raw XML of something I should be allowed to see."""
+        url = reverse("export", args=(self.resource.id,))
+        resp = self.client.get(url, follow=True)
+        body = resp.content.decode("utf-8")
+        self.assertTrue(resp.status_code == HTTP_OK)
+        self.assertTrue(self.resource.content_xml in body)
+
+    def test_export_no_permission(self):
+        """Get raw XML of something I should not be allowed to see."""
+        url = reverse("export", args=(self.resource.id,))
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        resp = self.client.get(url, follow=True)
+        self.assertTrue(resp.status_code == UNAUTHORIZED)
+
+    def test_export_nonexistent(self):
+        """Get raw XML of something than does not exist."""
+        url = reverse("export", args=(999999,))
+        resp = self.client.get(url, follow=True)
+        self.assertTrue(resp.status_code == NOT_FOUND)
--- a/ui/views.py
+++ b/ui/views.py
@@ -6,8 +6,9 @@ from __future__ import unicode_literals
 import logging
 
 from django.contrib.auth.decorators import login_required
+from django.core.exceptions import PermissionDenied as DjangoPermissionDenied
 from django.core.urlresolvers import reverse
-from django.http import HttpResponse
+from django.http import HttpResponse, Http404
 from django.http.response import HttpResponseForbidden
 from django.shortcuts import (
     render,
@@ -19,8 +20,9 @@ from guardian.decorators import permission_required_or_403
 from guardian.shortcuts import get_perms
 
 from learningresources.api import (
-    get_repo_courses, get_repo_or_error, get_repos,
+    get_repo_courses, get_repo, get_repos,
     get_resource, get_runs, get_user_tags,
+    NotFound, PermissionDenied,
 )
 from learningresources.models import Repository
 from roles.api import assign_user_to_repo_group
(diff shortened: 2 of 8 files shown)


ShawnMilo:
@cg FYI.

learningresources/tests/test_views.py
```diff
@@ -184,3 +187,25 @@ def test_invalid_repo_form(self):
...
+        self.assertTrue(resp.status_code == HTTP_OK)
+        self.assertTrue(self.resource.content_xml in body)
+
+    def test_export_no_permission(self):
+        """Get raw XML of something I should not be allowed to see."""
+        url = reverse("export", args=(self.resource.id,))
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        resp = self.client.get(url, follow=True)
+        self.assertTrue(resp.status_code == NOT_FOUND)
```
giocalitri: I'm not sold about the 404 in case the resource exists and the user does not have permissions: I still think it's a 403

giocalitri:
Other than this conceptual objection I do not see anything else weird

learningresources/tests/test_views.py
ShawnMilo: I'm happy to change it. What was the circumstance being discussed last week about wanting to return 404s instead of 403s so we don't leak info?

giocalitri: I guess the concern is that we do not want to disclose informations about the existence of the resource.
I understand it, but I guess there is not consensus on that.
cc: @Ferdi and @carsongee

ShawnMilo: As long as the URL only contains the primary key, it seems that replacing 403s with 404s can only prevent the user from figuring out _how many_ resources are in the database, but nothing at all about the actual resources.

giocalitri: yeah, like I said it's more conceptual that a real issue

Ferdi: If it was up to me, I would still go with 403

pdpinch:
I'm OK with a 403 in this case.

giocalitri:
if we go for 403, it's enough to change the function `has_repo` in `learning_resources.api` to call `get_repo_or_error`

ShawnMilo:
I think we should completely eliminate Django HTTP stuff from the API, since the API should be consumable by command lines scripts and anything else.

As is, my current fix isn't good enough because it just returns `None` for permission denied _and_ a missing resource. A couple of options are raising a custom exception (`learningresources.api.PermissionDenied` or `learningresources.api.NotFound`, perhaps), or returning two values -- the item and the error code.

It seems that custom exceptions are more reasonable, and what would be expected from a developer consuming the API. Neither solution will take long.

Thoughts? @giocalitri @Ferdi @carsongee

giocalitri:
I am for the custom exceptions. 
Still I think there should be a single function to check the view permission on the repo, since it's going to be used by all the other APIs (even if they still do not do it).

ShawnMilo:
Okay. I'll start on that unless @Ferdi or @carsongee have an issue with it.
- custom exceptions
- eliminate `has_repo` (it's only used in one place) in favor of `get_repo_or_error`

ui/views.py
```diff
@@ -240,7 +241,12 @@ def extra_context(self):
...
-    )
+    try:
+        return HttpResponse(
+            get_resource(resource_id, request.user.id).content_xml,
+            content_type='text/xml'
+        )
+    except NotFound:
+        raise Http404()
+    except PermissionDenied:
+        return HttpResponseForbidden("unauthorized")
```
giocalitri: why you don't raise a `PermissionDenied` instead and let Django handle the rendering. The plus is that then we could hook some custom templates for these errors.

ShawnMilo: That's fine, except I'd have to import `PermissionDenied` from both the API and Django, so I'd have to rename one. I don't mind doing that. Is it objectively better to raise Django's `PermissionDenied` here?

ShawnMilo: On second look, I don't think using Django's `PermissionDenied` would work. It's not a response.

giocalitri: Is it not intercepted by Django and rendered as a response?
(I was pretty sure it is handled in this way)

ShawnMilo: You are correct. I didn't know it did that. I'm running tests again and then I'll push that update.

ShawnMilo: @giocalitri: pushed

ui/views.py
```diff
@@ -6,8 +6,9 @@
 import logging
 
 from django.contrib.auth.decorators import login_required
+from django.core.exceptions import PermissionDenied as DjangoDenied
```
giocalitri: I guess DjangoPermissionDenied is more clear

ui/views.py
```diff
@@ -240,7 +242,12 @@ def extra_context(self):
...
-    )
+    try:
+        return HttpResponse(
+            get_resource(resource_id, request.user.id).content_xml,
+            content_type='text/xml'
+        )
+    except NotFound:
+        raise Http404()
+    except PermissionDenied:
+        raise DjangoDenied
```
giocalitri: isn't it `raise DjangoDenied()` ?

giocalitri:
small comments then we are ready to go

ui/views.py
ShawnMilo: Evidently it passes tests either way. I added the `()` for consistency. Stand by for another push when tests pass.

PR #175, opened 2015-06-16
minor cleanup

ShawnMilo:
Rename a couple of tests.


giocalitri:
:ship:

giocalitri:
Only one note, I hope that for changes like this we can avoid going through a pull request: it seems excessive to me.

PR #176, opened 2015-06-16
Export view protected with decorator

### LINKED ISSUE #151: Add authorizations to the export view
User must have permission on a repository in order to export a learningresource

This view is in the ui app

giocalitri:
The real protection and all the unit tests have been already implemented: this is only an extra step to protect the URL.
fixes #151

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 import logging
 
 from django.contrib.auth.decorators import login_required
-from django.core.exceptions import PermissionDenied as DjangoPermissionDenied
 from django.core.urlresolvers import reverse
 from django.http import HttpResponse, Http404
 from django.http.response import HttpResponseForbidden
@@ -22,7 +21,7 @@ from guardian.shortcuts import get_perms
 from learningresources.api import (
     get_repo_courses, get_repo, get_repos,
     get_resource, get_runs, get_user_tags,
-    NotFound, PermissionDenied,
+    NotFound
 )
 from learningresources.models import Repository
 from roles.api import assign_user_to_repo_group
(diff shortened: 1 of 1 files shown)


ui/views.py
```diff
@@ -242,6 +242,16 @@ def extra_context(self):
 # pylint: disable=unused-argument
 # repo_slug argument will be used by the decorator to protect the view
 @login_required
+@permission_required_or_403(
```
ShawnMilo, line 244: Since this is here, it's impossible to hit the `PermissionDenied` in the view. Since it's duplicating the same check and the decorator is a good place to do it, I think we should remove the second `except` to prevent confusion, duplication, and reduced test coverage (it should be impossible to cover that line in tests).

giocalitri, line 244: We could leave the lines and disable coverage for that (just saying)

ShawnMilo, line 244: We could, but that could still lead to confusion for someone (even me) reading the code later. Removing code is almost always a good thing. :trophy: 

Added bonus: This lets us remove that ugly `as DjangoPermissionDenied` import as well. And tests still pass. :joy: 

``` diff
--- ui/views.py
+++ ui/views.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 import logging

 from django.contrib.auth.decorators import login_required
-from django.core.exceptions import PermissionDenied as DjangoPermissionDenied
 from django.core.urlresolvers import reverse
 from django.http import HttpResponse, Http404
 from django.http.response import HttpResponseForbidden
@@ -22,7 +21,7 @@ from guardian.shortcuts import get_perms
 from learningresources.api import (
     get_repo_courses, get_repo, get_repos,
     get_resource, get_runs, get_user_tags,
-    NotFound, PermissionDenied,
+    NotFound
 )
 from learningresources.models import Repository
 from roles.api import assign_user_to_repo_group
@@ -261,5 +260,3 @@ def export(request, repo_slug, resource_id):
         )
     except NotFound:
         raise Http404()
-    except PermissionDenied:
-        raise DjangoPermissionDenied()
```

giocalitri, line 244: ok, will do and push: wait there with the mouse on the merge button

ShawnMilo:
:ok: Will merge after Travis is done. :checkered_flag:

PR #177, opened 2015-06-16
Update tox version & version lock for travis

bdero:
While using `tox 1.9.2`, the `envlist` override was not working, so the `testenv:docs` tox environment failed to build because `alabaster` (sphinx dependency) requires Python 2.7, but Python 3.4 was being used for the virtual environment instead.

The build [has not been failing on travis](https://travis-ci.org/mitodl/lore/builds/66964703) because tox was not version locked for travis builds. This change just updates the version of tox to fix local doc builds, as well as lock the version of tox running on travis to prevent build failure inconsistency in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 env:
   - DATABASE_URL=postgres://postgres@localhost:5432/postgres
 install:
-  - pip install tox
+  - pip install tox==2.0.2
   - pip install coveralls
 script: 
   - tox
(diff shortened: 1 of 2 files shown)


carsongee:
Can you change this to >=2.0.2? We should also fix the tox.ini so that it uses python2.7 as the interpreter i.e:

``` ini
[testenv:docs]
basepython = python2.7
...
```

instead of the env.  I'm not entirely sure that env statement works anyway from another project I've tried using that with.

bdero:
``` yaml
[testenv:docs]
envlist = py27
```

 is working in 2.0.2 as-is (the venv is created using 2.7)

I'll PR the use of >=

PR #179, opened 2015-06-16
adds Slack notifications for Travis CI

ShawnMilo:


bdero:
Is there another discussion going on around this?

ShawnMilo:
@giocalitri and I were just discussing whether this is better in doof. I'd like notifications when Travis finishes so PRs don't sit around neglected.

bdero:
Hipchat already [integrates with travis](https://www.hipchat.com/integrations) - I think the reason no one turned it on was because it would be too noisy to have all the build statuses in the ODL Engineering room, but I would think that we could turn it on in hipchat for a Lore room or something. Thoughts @pdpinch @carsongee ?

pdpinch:
Do reviews need to wait until Travis is done, or are you just looking for an alert about new PRs/status changes? I thought Doof was doing that now. 

I share @bdero 's concern about notification noise. If it's not too obscure, we could try a LORE room for notifications.

bdero:
I can't create new API keys since I'm not an admin on hipchat, but here's the equivalent config for hipchat using this repo, doof's API key, and the LORE room I just created:

``` yaml
notifications:
  hipchat:
    rooms:
      secure: aPUw0QpBDwWacAuI/j0I8+BGekatcrwcB1DEDy+d6AaWbUhrDgKl1no1UwHeiMkjS9slREHzgehPR3wE1oRInvhB1FNO0Rp2qCHoScgH8i+in5yvXgukYnMat1Hc74ml9CiN7KctMt2424kXT504mahe0+xlGxODjWXJMXZFw6o=
```

Not sure if we would want to use the same API key as doof, though, since the calls are throttled by hipchat and I'm not sure if we're close to the limit (I also don't know if it's throttled by key or account ;) ).

ShawnMilo:
This was meant to be a quickie so I could be notified when Travis completed, allowing prompt merging of another developer's pull request I was reviewing. Since it's turned into a time-sink I'm just closing it; it was never a big deal.

carsongee:
Just to add, you can do this with coveralls, since that is the real indicator the build is done, it can e-mail, hipchat, slack, campfire, etc, with no code required: https://coveralls.io/r/mitodl/lore/notifications/email

carsongee:
And by real, I mean final, when the build was successful..if you want notifications for failed builds too, then this would be the way to go

carsongee:
the this here being this PR as written, but preferably with hipchat: http://docs.travis-ci.com/user/notifications/#HipChat-notification

PR #183, opened 2015-06-17
Implement authorization for taxonomy pieces of REST API

noisecapella:


taxonomy/api.py
```diff
@@ -1,252 +1,54 @@
...
-        vocabulary_id (int): Vocabulary id
+        repo_id (int): Repository id
+        user_id (int): User id
+        vocab_slug (unicode): Vocabulary slug
 
     Returns:
         Vocabulary (Vocabulary): The vocabulary from the database
     """
-    return get_object_or_404(Vocabulary, id=vocabulary_id)
+    repo = get_repo(repo_slug, user_id)
```
ShawnMilo, line 26: This could break if the slug is bad. It should be possible to put it in the `try/except` below and replace `Vocabulary.DoesNotExist` with `models.DoesNotExist`.

taxonomy/api.py
```diff
@@ -1,252 +1,54 @@
...
-    """
-    vocabulary = get_object_or_404(Vocabulary, id=vocabulary_id)
-    learning_resource_type = get_object_or_404(
-        LearningResourceType,
-        name=learning_resource_type_name
-    )
-    vocabulary.learning_resource_types.remove(learning_resource_type)
+    repo = get_repo(repo_slug, user_id)
+    try:
+        return repo.vocabulary_set.get(
```
ShawnMilo, line 46: Some comment here about the `DoesNotExist` thing. I won't repeat it if I see it later on in the code.

taxonomy/tests/test_api.py
```diff
@@ -73,252 +60,42 @@ def setUp(self):
...
-        self.assertRaises(Http404,
-                          lambda: get_vocabularies_for_type("xyz"))
+        actual_term = get_term(
+            self.repo.slug,
+            self.user.id,
+            self.vocabulary.slug,
+            self.term.slug
+        )
+        self.assertEquals(self.term, actual_term)
+        with self.assertRaises(NotFound):
```
ShawnMilo, line 74: nit for future reference:

When reading this list of tests, it would be nice to use the names of the kwargs to save the reader from constantly looking up to figure out which positional argument is missing. Don't bother to update this, though.

taxonomy/api.py
```diff
@@ -1,252 +1,54 @@
...
+    repo = get_repo(repo_slug, user_id)
+    try:
+        return repo.vocabulary_set.get(slug=vocab_slug)
+    except Vocabulary.DoesNotExist:
+        raise NotFound()
 
 
-# pylint: disable=too-many-arguments
-@transaction.atomic
-def create_vocabulary(
```
ShawnMilo, line 32: Why was `create_vocabulary` removed? I know it wasn't doing anything other than `objects.create` for now, but if we ever change that (for example, set a default `vocabulary_type` or something, all the `objects.create` lines in the app will have to be hunted down and replaced.

I know this sounds like YAGNI, and it may be, but since it was already there it seems better to have left it.  But there are already four Vocabulary.objects.create lines in the code, and I only expect there to be more eventually.

noisecapella, line 32: Yeah, my rational is mostly YAGNI. The `Vocabulary.objects.create` lines are all in the tests. These API functions were created for the REST API to make use of so they were removed when they were left unused. If we end up needing this functionality later on we can add it back with support for authorization.

rest/permissions.py
```diff
@@ -44,3 +49,60 @@ def has_permission(self, request, view):
...
+            return False
+        return True
+
+
+class ManageTaxonomyPermission(BasePermission):
+    """Checks manage_taxonomy permission"""
+
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
```
ShawnMilo, line 104: I still don't think this should return `True` for any user. I know that this permission is only ever checked with other permissions, but the [documentation](http://www.django-rest-framework.org/api-guide/permissions/#custom-permissions) seems to recommend doing that check and then still checking for permissions based on whether it's a "destructive" method.

There _is_ one section in [the examples](http://www.django-rest-framework.org/api-guide/permissions/#examples) which shows a blind `return True` for `SAFE_METHODS`, but that example is clearly meant for objects the world can view and only owners can edit, like blog posts, not things like taxonomies which are strictly tied to the repository ownership of the user.

noisecapella, line 104: Fair enough, I'll add a check for the view_repo permission in this permission

taxonomy/api.py
ShawnMilo, line 32: Yep, I agree. I was mostly commenting on it being deleted when it was already there; didn't seem to be worth the trouble. But It's not going to kill us if we have to re-add it later.

rest/tests/test_rest.py
```diff
@@ -321,24 +315,25 @@ def test_term(self):
...
+            {
+                "name": "other name",
+                "description": "description",
+                "required": True,
+                "vocabulary_type": Vocabulary.FREE_TAGGING,
+                "weight": 1000,
+            }
+        )['slug']
+
+        self.delete_term(self.repo.slug, vocab_slug, new_term_slug)
```
ShawnMilo, line 336: Should there be a check here that the term is actually gone, such as deleting it again (or getting it again) and receiving `NOT_FOUND`?

noisecapella, line 336: Yes, I'll add it

taxonomy/api.py
noisecapella, line 26: `get_repo` should only raise one of the two exceptions defined in `learningresources.api`, so there should be no need to catch the exception there. If the slug is bad `NotFound` would be raised instead

ShawnMilo, line 26: Yeah, good point.

PR #184, opened 2015-06-17
Convert exceptions to Django exceptions

noisecapella:


ShawnMilo:
This is unnecessary; the exact code was just removed in #176 when the decorator was added. Discussion in that PR led to removal because it leads to redundant code which could never be hit, and also reduce test coverage.

noisecapella:
`get_repo` raises `PermissionDenied` so `get_resource` will also raise `PermissionDenied` since it calls `get_repo`

ShawnMilo:
@giocalitri, I believe that the decorator prevents any of these from ever happening. Is that correct?

giocalitri:
in that specific case yes  and in general in any case where the user does not have view permissions on the repo

noisecapella:
Ok, I've removed the except clause

ui/views.py
```diff
@@ -137,7 +137,10 @@ def upload(request, repo_slug):
     """
     Upload a OLX archive.
     """
-    repo = get_repo(repo_slug, request.user.id)
+    try:
+        repo = get_repo(repo_slug, request.user.id)
```
ShawnMilo: Everything looks good except these additions -- the decorator means it can't happen.

PR #185, opened 2015-06-17
Unlock tox for >= 2.0.2

bdero:
Addressing https://github.com/mitodl/lore/pull/177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 env:
   - DATABASE_URL=postgres://postgres@localhost:5432/postgres
 install:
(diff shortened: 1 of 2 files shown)


ShawnMilo:
Should this still be merged, given the recent conversation in HipChat about local tests failing for @giocalitri?

bdero:
This needs more work since 2.0.2 does environment variable isolation now (which is why it's failing) & envlist has no effect as @carsongee mentioned (it just happened to begin using python 2.7 instead of 3.4 after switching to tox 2  - I think @carsongee will be making a new PR.

PR #194, opened 2015-06-18
Make SSL configurable and enable deploying on release branch

carsongee:
This has been deployed to https://lore-release.herokuapp.com and all major functionality tested (DB, ES, Redis, workers, etc).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 python:
   - '2.7'
+services:
+  - elasticsearch
+before_script:
+  - sleep 10
 env:
   - DATABASE_URL=postgres://postgres@localhost:5432/postgres
 install:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: python
 python:
   - '2.7'
+services:
+  - elasticsearch
+before_script:
+  - sleep 10
```
bdero, line 7: So this is to wait for elasticsearch to start? Kind of wish heroku handled this automatically if it is.

lore/settings.py
```diff
@@ -135,12 +135,20 @@ def get_var(name, default):
...
-            os.path.join(BASE_DIR, 'db.sqlite3')
-        ))
+DEFAULT_DATABASE_CONFIG = dj_database_url.parse(
+    get_var(
+        'DATABASE_URL',
+        'sqlite:///{0}'.format(os.path.join(BASE_DIR, 'db.sqlite3'))
     )
+)
+# Enable SSL
+if get_var('LORE_USE_DB_SSL', False):
```
bdero: Do you think it would be slightly safer if we default this to `True` and override this as `False` when running the docker container (or whenever debug is on)?

.travis.yml
ShawnMilo, line 7: @bdero Yes, it's to wait for it to start. It's straight from the [Travis docs](http://docs.travis-ci.com/user/database-setup/). They explicitly recommend it for Elasticsearch and MongoDB.

bdero, line 7: Whoops, mistyped there, I totally meant travis :fearful: Thanks

lore/settings.py
carsongee: Fair point, secure by default and all.  I'll fix that.

bdero:
:cherry_blossom:

PR #195, opened 2015-06-18
members api #1

giocalitri:
this is the first part to just add a new permission

it also handle migrations in the right way and adds a Django Command for syncing permissions

### DIFF
--- a/roles/api.py
+++ b/roles/api.py
@@ -4,9 +4,9 @@ Functions for handling roles
 
 from __future__ import unicode_literals
 
-from django.contrib.auth.models import Group
+from django.contrib.auth.models import Group, Permission
 from django.db import transaction
-from guardian.shortcuts import assign_perm
+from guardian.shortcuts import assign_perm, get_perms, remove_perm
 
 from roles.permissions import RepoPermission, GroupTypes
 
@@ -40,13 +40,50 @@ def roles_init_new_repo(repo):
     with transaction.atomic():
         # administrator permissions
         for permission in RepoPermission.administrator_permissions():
-            assign_perm(permission, administrator_group, repo)
+            try:
+                Permission.objects.get(codename=permission)
+                assign_perm(permission, administrator_group, repo)
+            except Permission.DoesNotExist:
+                pass
         # curator permissions
         for permission in RepoPermission.curator_permissions():
-            assign_perm(permission, curator_group, repo)
+            try:
+                Permission.objects.get(codename=permission)
+                assign_perm(permission, curator_group, repo)
+            except Permission.DoesNotExist:
+                pass
         # author permissions
         for permission in RepoPermission.author_permissions():
-            assign_perm(permission, author_group, repo)
+            try:
+                Permission.objects.get(codename=permission)
+                assign_perm(permission, author_group, repo)
+            except Permission.DoesNotExist:
+                pass
+
+
+def roles_clear_repo_permissions(repo):
+    """
+    Removes all the permissions a group has on a repo
+    Args:
+        repo (learningresources.models.Repository): repository
+    Returns:
+        None
+    """
+    with transaction.atomic():
+        administrator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.REPO_ADMINISTRATOR.format(repo.slug)
+        )
+        curator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.REPO_CURATOR.format(repo.slug)
+        )
+        author_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.REPO_AUTHOR.format(repo.slug)
+        )
+        all_groups = [administrator_group, curator_group, author_group]
+        for group in all_groups:
+            perms = get_perms(group, repo)
+            for perm in perms:
+                remove_perm(perm, group, repo)
 
 
 def roles_update_repo(repo, old_slug):
--- a/roles/management/commands/sync_permissions.py
+++ b/roles/management/commands/sync_permissions.py
@@ -1,0 +1,20 @@
+"""
+Shell command to synchronize permissions and apply the latest to all groups
+"""
+
+from __future__ import unicode_literals
+
+from django.core.management.base import BaseCommand
+
+from roles.utils import sync_groups_permissions
+
+
+class Command(BaseCommand):
+    """
+    Command for sync_permissions
+    """
+    help = "Synchronizes and updates permissions of the repository groups"
+
+    def handle(self, *args, **options):
+        """Command handler"""
+        sync_groups_permissions()
--- a/roles/signals.py
+++ b/roles/signals.py
@@ -1,0 +1,22 @@
+"""
+Signals handlers
+"""
+from __future__ import unicode_literals
+
+from django.db.models.signals import post_migrate
+from django.dispatch import receiver
+
+from learningresources.models import Repository
+from roles.utils import sync_groups_permissions
+
+
+# pylint: disable=unused-argument
+@receiver(post_migrate)
+def my_callback(sender, **kwargs):
+    """
+    Signal handler to apply new permissions to repositories
+    only in case the sender is learningresources.
+    """
+    # pylint: disable=protected-access
+    if sender.name == Repository._meta.app_label:
+        sync_groups_permissions()
--- a/roles/tests/test_api.py
+++ b/roles/tests/test_api.py
@@ -7,7 +7,10 @@ from __future__ import unicode_literals
 from django.contrib.auth.models import Group
 from django.utils.text import slugify
 from guardian.core import ObjectPermissionChecker
+from guardian.shortcuts import get_perms
+from mock import patch
 
+from learningresources.api import create_repo
 from learningresources.models import Repository
 from learningresources.tests.base import LoreTestCase
 from roles import api
@@ -20,7 +23,8 @@ class TestRoleApi(LoreTestCase):
     roles_init_new_repo is tested indirectly from
     the repository model where it is called
     """
-    # pylint: disable=too-many-instance-attributes
+    # there are too many self assignment in the init: disabling the check
+    # pylint: disable=too-many-instance-attributes, invalid-name
     def __init__(self, *args, **kwargs):
         super(TestRoleApi, self).__init__(*args, **kwargs)
         self.repo_name = 'my little+test repo'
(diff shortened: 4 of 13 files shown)


roles/api.py
```diff
@@ -40,13 +40,51 @@ def roles_init_new_repo(repo):
...
         for permission in RepoPermission.author_permissions():
-            assign_perm(permission, author_group, repo)
+            try:
+                Permission.objects.get(codename=permission)
+                assign_perm(permission, author_group, repo)
+            except Permission.DoesNotExist:
+                pass
+
+
+def roles_whipe_repo_permissions(repo):
```
carsongee: s/whipe/wipe, or clear or remove

roles/api.py
```diff
@@ -40,13 +40,51 @@ def roles_init_new_repo(repo):
...
+            name=GroupTypes.REPO_ADMINISTRATOR.format(repo.slug)
+        )
+        curator_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.REPO_CURATOR.format(repo.slug)
+        )
+        author_group, _ = Group.objects.get_or_create(
+            name=GroupTypes.REPO_AUTHOR.format(repo.slug)
+        )
+        all_groups = [administrator_group, curator_group, author_group]
+        with transaction.atomic():
```
carsongee: you are already in a transaction, no need to start a new one

roles/management/commands/sync_permissions.py
```diff
@@ -0,0 +1,22 @@
...
+class Command(BaseCommand):
+    """
+    Command for sync_permissions
+    """
+    help = "Synchronizes and updates permissions of the repository groups"
+
+    def handle(self, *args, **options):
+        """Command handler"""
+        sync_groups_permissions()
+        sys.stdout.write('Permissions synchronized \n')
```
carsongee: I  prefer no output on success.  It makes cron easier

roles/signals.py
```diff
@@ -0,0 +1,18 @@
...
+
+from django.db.models.signals import post_migrate
+from django.dispatch import receiver
+
+from roles.utils import sync_groups_permissions
+
+
+# pylint: disable=unused-argument
+@receiver(post_migrate)
+def my_callback(sender, **kwargs):
```
carsongee, line 15: Is it possible to use this to determine what app is sending the migration?  If so, it might be nice to only run this for those migrations instead of churning the DB on every single one.  If you decide that isn't worth it, then just change the args to **kwargs since there is no real reason to handle `sender` (or *args, **kwargs if one of them is a pos arg).

roles/tests/test_api.py
```diff
@@ -76,6 +79,60 @@ def test_roles_init_new_repo_2(self):
             sorted(RepoPermission.author_permissions())
         )
 
+    def test_roles_init_new_repo_3(self):
```
carsongee: It is already a bit late, but instead of using numbers for all of these, it would be nice if you described what was under test.  So that if they fail it is easier to understand what broke. If Pylint complains about the method length, feel free to disable the Pylint check for the class

roles/tests/test_api.py
```diff
@@ -208,3 +265,17 @@ def test_remove_user_from_group(self):
             group_type=GroupTypes.REPO_ADMINISTRATOR
         )
         self.assertNotIn(self.user, admin.user_set.all())
+
+    def test_whipe_permissions(self):
```
carsongee: s/whipe/wipe, or remove, or clean

roles/tests/test_permissions.py
```diff
@@ -57,10 +59,14 @@ def test_repo_permissions(self):
             RepoPermission.add_edit_metadata,
             add_edit_metadata
         )
+        self.assertEqual(
+            RepoPermission.manage_repo_users,
+            manage_repo_users
+        )
         self.assertListEqual(
             RepoPermission.administrator_permissions(),
             [view_repo[0], import_course[0],
```
carsongee: nit: This is getting a bit hard to read, maybe switch to:

``` python
[
    view_repo[0],
    import_course[0],
    manage_taxonomy[0],
    add_edit_metadata[0]]
    manage_taxonomy[0],
    add_edit_metadata[0],
    manage_repo_users[0],
]
```

it will also have the side effect of make adding a new permission easier in the future

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+from learningresources.tests.base import LoreTestCase
+from roles.permissions import RepoPermission, GroupTypes
+from roles import utils
+
+
+class TestRoleUtils(LoreTestCase):
+    """
+    Tests for the roles Utils
+    """
+    # pylint: disable=too-many-instance-attributes
```
carsongee: I'm not familiar with this one, can you please add a comment above describing what this is, and why it is disabled?

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            mock_method.return_value = None
+            utils.sync_groups_permissions()
+        # permissions after wiping
+        self.check_group_permission(
+            repo,
+            admin,
+            []
+        )
+        # restore permissions the first call of sync_groups_permissions
+        # will not do anything because the permissions have been alredy wiped
```
carsongee: nits: s/restore/Restore s/alredy/already and a period at the end

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            sorted(expected_permissions)
+        )
+
+    def test_sync_permissions(self):
+        """
+        test the syncpermission
+        """
+        repo = create_repo(self.repo_name, self.repo_desc, self.user.id)
+
+        # permissions before wiping
```
carsongee: nit: Get default permissions before wiping.

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        admin = Group.objects.get(name=self.group_admin)
+        self.check_group_permission(
+            repo,
+            admin,
+            RepoPermission.administrator_permissions()
+        )
+        with patch.object(utils, 'roles_init_new_repo') as mock_method:
+            mock_method.return_value = None
+            utils.sync_groups_permissions()
+        # permissions after wiping
```
carsongee: nit: Get permissions after wiping

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            []
+        )
+        # restore permissions the first call of sync_groups_permissions
+        # will not do anything because the permissions have been alredy wiped
+        utils.sync_groups_permissions()
+        self.check_group_permission(
+            repo,
+            admin,
+            RepoPermission.administrator_permissions()
+        )
```
carsongee, line 71: There aren't any asserts in here. we shouldn't use our own methods to verify our own code unless I'm missing something

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_sync_permissions_command(self):
+        """
+        tests the syncpermission via manage.py command
+        """
+        repo = create_repo(self.repo_name, self.repo_desc, self.user.id)
+
+        # permissions before wiping
```
carsongee: same nits for the same comments

roles/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            []
+        )
+        # restore permissions the first call of sync_groups_permissions
+        # will not do anything because the permissions have been alredy wiped
+        call_command('sync_permissions')
+        self.check_group_permission(
+            repo,
+            admin,
+            RepoPermission.administrator_permissions()
+        )
```
carsongee: I would actually just mock your sync permissions function in the management command for this test and just assert it has been called to limit the scope and duplicated test code, though not a big deal, just a bit cleaner

roles/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    existing group repositories
+
+    Args:
+        None
+    Returns:
+        None
+    """
+    for repo in Repository.objects.all():
+        # this needs to be idempotent
+        roles_whipe_repo_permissions(repo)
```
carsongee: whipe again

roles/signals.py
giocalitri, line 15: I guess it is technically possible, but I need more time to investigare and I think I spent enough on this. If you have an example working with migrations I can try.

roles/tests/test_api.py
giocalitri: I'll cage for the tests about this PR

giocalitri: On the other hand, I guess it's PEP8 complaining, not Pylint: is there a way to disable these checks?

giocalitri: I take it back: it's pylint

roles/tests/test_utils.py
giocalitri: here it's not necessary: it was a leftover of code I copied, but in roles/tests/test_api.py is necessary:
pylint is complain about too many `self` assignment in the `__init__` function

roles/tests/test_utils.py
giocalitri, line 71: `check_group_permission` is a helper method of the test class: in there I run an assert

```
self.assertListEqual(
        sorted(get_perms(group, repo)),
        sorted(expected_permissions)
)
```

roles/tests/test_utils.py
giocalitri: But I would like to test that the result is the same, not that it just gets called

roles/tests/test_utils.py
carsongee, line 71: so, yeah, definitely missed that ;)

roles/tests/test_utils.py
carsongee: You've already tested that with `test_sync_permissions` above, so you really just need a unit test, not an integration test here as it can speed tests up a little bit, but not a big deal.  Happy to merge without changing this.

roles/tests/test_utils.py
giocalitri, line 71: You are right: I misread the comment.

roles/signals.py
carsongee, line 15: It should be easy.  `sender` is an `AppConfig` instance, so you just need to add:

``` python
if sender.name == 'learningresources':
...
```

wrapper around it: https://docs.djangoproject.com/en/1.8/ref/signals/#post-migrate and https://docs.djangoproject.com/en/1.8/ref/applications/#configurable-attributes for reference.

roles/tests/test_api.py
```diff
@@ -76,6 +80,63 @@ def test_roles_init_new_repo_2(self):
...
+                description=self.repo_desc,
+                user_id=self.user.id,
+            )
+            curator = Group.objects.get(name=self.group_curator)
+            self.assertListEqual(
+                get_perms(curator, repo),
+                []
+            )
+
+    # pylint: disable=invalid-name
```
carsongee: Just add one of these right below the class declaration instead of repeating them.

roles/tests/test_api.py
```diff
@@ -208,3 +269,17 @@ def test_remove_user_from_group(self):
             group_type=GroupTypes.REPO_ADMINISTRATOR
         )
         self.assertNotIn(self.user, admin.user_set.all())
+
+    def test_wipe_permissions(self):
+        """
+        Test for roles_whipe_repo_permissions
```
carsongee: s/whipe/wipe

roles/tests/test_permissions.py
```diff
@@ -57,17 +59,27 @@ def test_repo_permissions(self):
...
+        )
         self.assertListEqual(
             RepoPermission.administrator_permissions(),
-            [view_repo[0], import_course[0],
-             manage_taxonomy[0], add_edit_metadata[0]]
+            [view_repo[0],
+             import_course[0],
+             manage_taxonomy[0],
+             add_edit_metadata[0],
+             manage_repo_users[0]]
```
carsongee: nit: still using `[<item>`, instead of

```
[
    <item>,
```

so it isn't quite as easy to add a new item...again, total nit.

carsongee:
Very close, just the conditional signal check and a few nits I can take or leave.  Also will need a squash

giocalitri:
Updated, squashed and pushed

roles/tests/test_api.py
```diff
@@ -20,7 +23,8 @@ class TestRoleApi(LoreTestCase):
     roles_init_new_repo is tested indirectly from
     the repository model where it is called
     """
-    # pylint: disable=too-many-instance-attributes
+    # there are too many self assignment in the init: disabling the check
+    # pylint: disable=too-many-instance-attributes, invalid-name
```
carsongee, line 27: If the too many instance attributes only happens in init, you should scope the check to that function, as disables should be as close to the problem as you can be.  The invalid name, can stay at the class scope though.

roles/tests/test_api.py
```diff
@@ -208,3 +266,17 @@ def test_remove_user_from_group(self):
             group_type=GroupTypes.REPO_ADMINISTRATOR
         )
         self.assertNotIn(self.user, admin.user_set.all())
+
+    def test_wipe_permissions(self):
```
carsongee: nit: change to clear since we renamed the function under test.

carsongee:
Just those last two tiny things

roles/tests/test_api.py
giocalitri, line 27: If I move it to the init, it does not disable the pylint check: I guess it refers to the class and not to the init

giocalitri:
fixed the test and left the pylint as it is

carsongee:
:japanese_goblin:

PR #198, opened 2015-06-19
Fixes indexing so it's possible to search on content_xml.

### LINKED ISSUE #197: search not returning results from `content_xml` field
When searching on content from the `content_xml` field, results only appear if the same content appears in the content's title. Search should return matches from within the `content_xml` field.

ShawnMilo:
The `model_attr` kwarg is ignored when `use_template` is used. If using a
template, all fields must be specified in the template.

Upon review, it seems that the complication of using a template was unnecessary,
because the template was so simple that having it separated the indexing
definitions unnecessarily into multiple locations.

Moves definition of `content_xml` into search_indexes.py with the definition
for all other fields.

Adds `content_xml` to the fields being indexed.

Removes repository name and ID from indexing (never should have been there).

resolves #197

### DIFF
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -4,6 +4,8 @@ Index information for Haystack.
 http://django-haystack.readthedocs.org/en/latest/tutorial.html
 """
 
+from __future__ import unicode_literals
+
 from haystack import indexes
 
 from learningresources.models import LearningResource
@@ -13,8 +15,7 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     """
     Index configuration for the LearningResource model.
     """
-    text = indexes.CharField(
-        document=True, model_attr="content_xml", use_template=True)
+    text = indexes.CharField(document=True)
     resource_type = indexes.CharField(
         model_attr="learning_resource_type",
         faceted=True,
@@ -29,6 +30,12 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     def index_queryset(self, using=None):
         """Records to check when updating entire index."""
         return self.get_model().objects.all()
+
+    def prepare_text(self, obj):  # pylint: disable=no-self-use
+        """Indexing of the primary content of a LearningResource."""
+        return "{0} {1} {2}".format(
+            obj.title, obj.description, obj.content_xml,
+        )
 
     def prepare_run(self, obj):  # pylint: disable=no-self-use
         """Define what goes into the "run" index."""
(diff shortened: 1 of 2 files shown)


ShawnMilo:
@pdpinch: FYI

@giocalitri: Please review; it'll probably take you 30 seconds since you know what you're looking at.

search/search_indexes.py
```diff
@@ -30,6 +31,12 @@ def index_queryset(self, using=None):
         """Records to check when updating entire index."""
         return self.get_model().objects.all()
 
+    def prepare_text(self, obj):  # pylint: disable=no-self-use
```
Ferdi, line 34: Can we index these as separate fields ?

Ferdi:
It doesn't have to be part of this PR, but are we stripping out xml tags on content_xml field ?

search/search_indexes.py
carsongee, line 34: why not remove self and add `@staticmethod` instead of disabling the pylint check?

ShawnMilo, line 34: @Ferdi: Yes. If we do that, then only what is described in `prepare_text` will be searched when someone types in the search box. Everything else will be a separate filter.

@carsongee: This is a feature of Haystack, just like the `clean_fieldname` form override in Django. However, unlike Django's "clean" methods, it does not use something like `self.cleaned_data`, so there's just no "self" use.

Ferdi, line 34: ah, is that because you cant do "OR" in haystack ?

ShawnMilo, line 34: @Ferdi It's slightly different. If a `LearningResource` has `content_xml` and `title`, and you want to type in "protein" and match any `LearningResource` that has "protein" in either `content_xml` or `title`, then you have to index them together.

This is because:

> Every SearchIndex requires there be one (and only one) field with document=True. This indicates to both     Haystack and the search engine about which field is the primary field for searching within.

From [this](http://django-haystack.readthedocs.org/en/latest/tutorial.html#configuration) documentation.

Everything else we index apart from the one marked `document=True` is going to have to be a separate filter. In the case of a `LearningResource`, the `content_xml` is the "meat" of the object, so it would be the "document."

Ferdi, line 34: @ShawnMilo ok, btw, is this related/helpful ? http://stackoverflow.com/a/2741778/4992607

ShawnMilo, line 34: @Ferdi Yes, it's directly related. Putting two values in the "prepare" method and creating a template are two ways to do exactly the same thing. In fact, in this PR I've removed the template that was doing exactly that in order to keep all the logic in one place.

giocalitri:
LGTM. I tested the fix and it works. No comments on the code changes. Merging

PR #199, opened 2015-06-19
Added importing of static assets during course import

carsongee:
I'm still working through tests and such @pwilkins, but if you wanted the model, it is here and shouldn't change.


importer/api/__init__.py
```diff
@@ -112,6 +122,7 @@ def import_course(bundle, repo_id, user_id):
         user_id=user_id,
     )
     import_children(course, src, None)
+    return course
 
 
 def import_children(course, element, parent):
```
pwilkins, line 128: I expected that parsing the static asset XML would be part of the import. Where in this pipeline did you plan for the parsing to take place?

importer/api/__init__.py
```diff
@@ -101,7 +111,7 @@ def import_course(bundle, repo_id, user_id):
         repo_id (int): Primary key of repository course belongs to
         user_id (int): Primary key of Django user doing the import
     Returns:
-        None
+        learningresources.Course
```
noisecapella: Should be `learningresources.models.Course` or just `Course`

importer/api/__init__.py
```diff
@@ -136,3 +147,24 @@ def import_children(course, element, parent):
...
+
+
+def import_static_assets(path, course):
+    """
+    Upload all assets and create model records of them for a given
+    course and path.
+
+    Args:
+        path (string_type): course specific path to extracted OLX tree.
+        course (learningresources.Course): Course to add assets to.
```
noisecapella: Same comment about docstring type

importer/api/__init__.py
```diff
@@ -136,3 +147,24 @@ def import_children(course, element, parent):
...
             import_children(course, child, resource)
+
+
+def import_static_assets(path, course):
+    """
+    Upload all assets and create model records of them for a given
+    course and path.
+
+    Args:
+        path (string_type): course specific path to extracted OLX tree.
```
noisecapella: I think we're using `unicode` for strings in docstrings

importer/api/__init__.py
```diff
@@ -136,3 +147,24 @@ def import_children(course, element, parent):
...
+
+    Returns:
+        None
+    """
+    for root, _, files in walk(path):
+        for name in files:
+            with open(join(root, name), 'r') as open_file:
+                django_file = File(open_file)
+                # Remove base path from file name
+                django_file.name = join(root, name).replace(path + sep, '')
```
noisecapella: I'm a bit uneasy about using `replace` when we clearly only want to chop off the prefix. Can you change this to check the prefix and chop it off if it matches? (or error if we don't find it since it seems we should always see it here)

importer/tests/test_import.py
```diff
@@ -109,3 +115,106 @@ def test_import_task(self):
             Course.objects.count(),
             original_count + 1,
         )
+
+    @staticmethod
+    def test_import_course_from_path():
+        """
+        Validate parameters and returns fir course from path.
```
noisecapella: Is `fir` a typo?

importer/api/__init__.py
ShawnMilo: For what it's worth, `splitext` may help here. It looks like it puts the pathsep on the "wrong side" for what you want, but it's a bit more explicit than `replace`.

``` python
from os.path import splitext
filename = "something.txt"
result = splitext(filename)
print result
```

Output: `('something', '.txt')`

importer/tests/test_import.py
```diff
@@ -109,3 +115,106 @@ def test_import_task(self):
...
+        self.addCleanup(default_storage.delete, asset.asset)
+
+    def test_static_import_integration(self):
+        """
+        Do integration test to validate course import with static assets.
+        """
+        original_count = Course.objects.count()
+        import_file(
+            self.get_course_static_tarball(), self.repo.id, self.user.id)
+        self.assertEqual(
```
noisecapella, line 209: It looks like `import_file` also deletes the file at `self.get_course_static_tarball()`. Can you add an assertion for this?

learningresources/api.py
```diff
@@ -220,3 +220,29 @@ def get_resource(resource_id, user_id):
...
+    Create a static asset for a given course.
+
+    Warning:
+        This takes and open file handle and does not close it. You
+        will need to handle the opening and closing the
+        ``file_handle`` outside of this method.
+
+    Args:
+        course_id (int): Primary key of the Course to add the asset to.
+        file_handle (django.core.files.file):
```
noisecapella: uppercase for `file`

learningresources/api.py
```diff
@@ -220,3 +220,29 @@ def get_resource(resource_id, user_id):
...
+            An open file handle to save to the model.
+
+    Returns:
+        learningresources.StaticAsset
```
noisecapella: Change to `learningresources.models.StaticAsset` or just `StaticAsset`

learningresources/models.py
```diff
@@ -19,6 +19,18 @@
 log = logging.getLogger(__name__)
 
 
+def static_asset_basepath(asset, filename):
+    """
+    Creates folder base path for given asset.
```
noisecapella, line 24: Since this is used outside this file, can you add the args and return value for this?

learningresources/models.py
```diff
@@ -19,6 +19,18 @@
...
+    """
+    return 'assets/{org}/{course_number}/{run}/{filename}'.format(
+        org=asset.course.org,
+        course_number=asset.course.course_number,
```
noisecapella, line 37: I don't think we currently don't verify that these are free of slashes. Is it possible to do a check and raise an exception if a slash is found? There might be a better place to do this, not sure

carsongee, line 37: That is a requirement of OLX, that our OLX parser xbundle should be validating.

carsongee, line 37: And even if it did, that would still be ok actually, just creates an extra folder

importer/api/__init__.py
carsongee: How about lstrip?

importer/tests/test_import.py
carsongee: yeah, should be `fir``course_from_path``

importer/tests/test_import.py
carsongee, line 209: That should probably get tested in https://github.com/mitodl/lore/blob/feature/upload_assets/importer/tests/test_import.py#L95, but I can add it both places.

learningresources/models.py
carsongee, line 24: Good catch

learningresources/tests/base.py
```diff
@@ -157,3 +157,14 @@ def get_course_single_tarball(self):
...
         path = join(abspath(dirname(__file__)), "testdata", "courses")
         return self.copy_file(join(path, "single.tgz"))
+
+    def get_course_static_tarball(self):
+        """
+        Get the path to a course with course.xml in the root of the
+        archive and a static folder with two assets (one of which is
+        in a subfolder.
+
+        Returns: path (unicode): absolute path to tarball.
```
noisecapella: Can you change this to:

```
Returns:
    unicode: absolute path to tarball
```

It's nitpicky but if we ever need automatic docstring inspection for something it'll be useful for these to conform strictly to the spec

importer/tests/test_import.py
```diff
@@ -109,3 +115,106 @@ def test_import_task(self):
...
+                        import_course_from_path(
+                            test_path, test_repo_id, test_user_id
+                        )
+                        mock_import.assert_called_with(
+                            mock_bundle(), test_repo_id, test_user_id
+                        )
+                        mock_static.assert_called_with(
+                            os.path.join(test_path, 'static'), True
+                        )
+
```
noisecapella, line 147: Can you add a test or assertions where `import_course_from_path` is passed a missing or invalid path? Or permissions for the user and repo (do we currently check for this?)

importer/tests/test_import.py
```diff
@@ -109,3 +115,106 @@ def test_import_task(self):
...
+        self.assertEqual(assets.count(), 1)
+        asset = assets[0]
+        dummy = mock.MagicMock()
+        dummy.course = self.course
+        self.assertEqual(
+            asset.asset.name,
+            static_asset_basepath(dummy, basename)
+        )
+        self.addCleanup(default_storage.delete, asset.asset)
+
```
noisecapella, line 170: Can you add a test or assertions for invalid path or course `import_static_assets`?

importer/tests/test_import.py
```diff
@@ -109,3 +115,106 @@ def test_import_task(self):
...
+            static_asset_basepath(dummy, os.path.join(subdir_name, basename))
+        )
+        self.addCleanup(default_storage.delete, asset.asset)
+
+    def test_static_import_integration(self):
+        """
+        Do integration test to validate course import with static assets.
+        """
+        original_count = Course.objects.count()
+        import_file(
```
noisecapella, line 207: Can you add a test or assertions for an invalid path or permissions for repo and user for `import_file`?

learningresources/models.py
noisecapella, line 37: Sounds good

learningresources/tests/test_api.py
```diff
@@ -146,6 +157,30 @@ def test_invalid_create_repo(self):
...
+    def test_create_static_asset(self):
+        """
+        Validate that we can create assets.
+        """
+        basename = 'blah.txt'
+        file_contents = b'hello\n'
+        with tempfile.TemporaryFile() as temp:
+            temp.write(file_contents)
+            test_file = File(temp, name=basename)
+            asset = create_static_asset(self.course.id, test_file)
```
noisecapella, line 173: Can you add a test or assertions where a closed file is passed in?

noisecapella:
What changes were made to the two test tarballs, just for my information?

learningresources/tests/base.py
carsongee: Yeah, that was inherited from all these already.  I'll change them all.

importer/tests/test_import.py
carsongee, line 147: That is all done in https://github.com/mitodl/lore/blob/feature/upload_assets/importer/api/__init__.py#L28.  The only caller of import_course_from_path I'm aware is that.  We could _ this method in fact

importer/tests/test_import.py
carsongee, line 170: That is all validated earlier, anything that slips through I would actually want an uncaught exception to be raised.

importer/tests/test_import.py
carsongee, line 207: That would be a test of the view right (https://github.com/mitodl/lore/blob/feature/upload_assets/ui/views.py#L134)?  That is outside of this area (and also out of scope for this PR).

carsongee:
Yeah, the tarball transparency here is why I made #222. I just added an third folder that is empty for two courses to increase branch coverage of import while I was in here, and the new tar ball has a static fodler which I assert the contents of in the integration test here.

importer/api/__init__.py
carsongee: lol, I mean replace(path, 1)

noisecapella:
Looks good but some tests are failing. I'll merge after that completes

PR #203, opened 2015-06-22
Adds faceted search for taxonomy items.

### LINKED ISSUE #155: Update search index when taxonomy items are changed.
Changes made to taxonomy items for a LearningResource should take effect in the search immediately.

### LINKED ISSUE #158: Search: Facets for dynamic metadata
See also #120 
Continue to built on #144 by adding facets for dynamic metadata ((course, run and learningresource type)

loredynfacets

### LINKED ISSUE #202: Are we stripping xml markup when indexing content_xml field ?
cc @ShawnMilo

ShawnMilo:
User-created vocabularies and terms will be available for filtering
on the repository listing page.

closes #158 [finishes #96959502]
closes #155 [finishes #96912500]
closes #202 [finishes #97469022]

### DIFF
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -2,13 +2,23 @@
 Index information for Haystack.
 
 http://django-haystack.readthedocs.org/en/latest/tutorial.html
+
+There are three ways to define what gets indexed for a field in
+the index. They can not be mixed.
+
+* Use the model_attr kwarg: This works when the field is text, or a
+  foreign key to a model whose unicode representation is what we want.
+* Use the use_template kwarg: For complex mixing of fields.
+* Use a prepare_* function: Flexible and easy.
 """
 
 from __future__ import unicode_literals
 
+from lxml import etree
 from haystack import indexes
 
 from learningresources.models import LearningResource
+from taxonomy.models import Vocabulary
 
 
 class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
@@ -20,8 +30,8 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
         model_attr="learning_resource_type",
         faceted=True,
     )
-    course = indexes.CharField(model_attr="course", faceted=True)
-    run = indexes.CharField(model_attr="course", faceted=True)
+    course = indexes.CharField(faceted=True)
+    run = indexes.CharField(faceted=True)
 
     # repository is here for filtering the repository listing page by the
     # repo_slug in the URL. It is not used or needed in the repository listing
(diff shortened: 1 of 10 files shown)


ShawnMilo:
Test failed on 3.3 (but not 2.7 or 3.4) based on the position of a pylint disable flag. I moved it and am testing locally now.

pdpinch:
How do I add user-created vocabularies and terms in order to test this?

ShawnMilo:
@pdpinch 
I used a script I wrote. There is currently no UI for this. (Adding this comment here publicly although already discussed in HipChat.)

ShawnMilo:
The test failure is _again_ due to Python 3.3. I'm in the middle of a code review locally, so I'll try moving the pylint hint again in a bit.

search/search_indexes.py
```diff
@@ -33,14 +43,46 @@ def index_queryset(self, using=None):
 
     def prepare_text(self, obj):  # pylint: disable=no-self-use
         """Indexing of the primary content of a LearningResource."""
+        try:
+            # Strip XML tags from content before indexing.
+            tree = etree.fromstring(obj.content_xml)
+            content = etree.tostring(tree, encoding="utf-8", method="text")
+        except etree.XMLSyntaxError:
+            # For blank/invalid XML.
+            content = obj.content_xml
```
giocalitri, line 57: Should we log somewhere that this is happening?
Maybe could even show the user that this is happening (in another issue, since here it's out of scope)

search/search_indexes.py
```diff
@@ -33,14 +43,46 @@ def index_queryset(self, using=None):
...
+        to the index because they must be dynamic.
+
+        Technically, we could handle the other stuff (run, course, etc.) here
+        as well, but don't because explicit is better than implicit.
+        """
+        prepared = super(LearningResourceIndex, self).prepare(obj)
+        for vocab in Vocabulary.objects.all():
+            # Values with spaces do not work, so these are replaced with
+            # underscores.
+            name = vocab.name.replace(" ", "_")
```
giocalitri: is space the only character that gives problem?

ShawnMilo: Yes, because the backend treats it as multiple tokens. I did experiment with `slugify`, but it just complicated things slightly more and didn't add any benefit.

ui/urls.py
```diff
@@ -31,10 +32,20 @@
 
 
 def get_sqs():
```
giocalitri: what about moving this function to the search app?

giocalitri: aren't tests for it? (not sure they are necessary)

search/search_indexes.py
ShawnMilo, line 57: @Ferdi What do you think? Is it common or valid for the content of a chunk of XML not to have valid XML inside of it?

@giocalitri: It may be that this would be rare or nonexistent in reality, since should normally be dealing with valid XML in OLX format, and that the only reason I encountered this was due to strings being passed during tests.

ui/views.py
```diff
@@ -227,6 +227,10 @@ def dispatch(self, *args, **kwargs):
     def extra_context(self):
         """Add to the context."""
         context = super(RepositoryView, self).extra_context()
+        vocabularies = [
+            x.replace(" ", "_")
```
giocalitri: this is the second place where you do this: maybe it's too much, but what about a function that does this replacement (in case you don't need to do only this in the future)

ui/urls.py
ShawnMilo: I think it could be trivially moved into the search app. I'll take a run at it.

I can add a test to `ui/tests/test_learningresources_views.py` which should fully cover this function.

ui/views.py
ShawnMilo: It's only done here to match the other one.

The trivial solution would be to disallow spaces in vocabulary names.

ui/urls.py
giocalitri: ok, because your drop in coverage is also caused by this function

ui/views.py
giocalitri: I understand, but it's hard(er) to keep track of the logical connection with just the `replace`

ShawnMilo: Yep. Do you have a suggestion? How about adding a method to the `Vocabulary` model such as `get_slug` instead of using the `name` attribute?

giocalitri: good idea: go for it!

search/tests/test_indexing.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_index_vocabulary(self):
+        """
+        Test that LearningResource indexes are updated when a
+        a term is added or removed.
+        """
+        term = self.terms[0]
+        self.assertTrue(self.count_faceted_results(
+            self.vocabulary.name, term.label) == 0)
+        self.resource.terms.add(term)
```
giocalitri, line 25: where do you create self.resource?

ShawnMilo, line 25: It's in `LoreTestCase`, and used in a bunch of different tests.

giocalitri, line 25: Sorry, searched and missed :-(

ShawnMilo:
Demo videos.

[video 1](http://youtu.be/QTIi2pER8_A), filtering on built-in facets
[video 2](https://youtu.be/KJ2Nwq3Ncnc), filtering on user-added facets

ShawnMilo:
Status was "unable to merge." Manually merged master and updated what had broken; namely that the `create_vocabulary` and `create_term` were deleted from `taxonomy.api` but had been in use in tests. Replaced them with `Vocabulary.objects.create` and `Term.objects.create` in the tests.

carsongee:
@giocalitri Freeze is over, feel free to merge.

Issue #129, opened 2015-06-09
Members REST API for managing users in repositories [#96516348]


pdpinch:
See also https://github.com/mitodl/lore/pull/119/files#diff-2625016b50d68d922257f74801cac29cR294

Add/List/remove repository members via REST API. 
- [x] ~~Members can be added by e-mail address (typically kerberos id + mit.edu).~~ (this will be handled in the ui)
- [x] If the user doesn't exist, an error is returned
- [x] Should be limited to users with admin privileges on the repository (could be another issue if necessary)
- [x] update the API docs

pdpinch:
@giocalitri asked some questions:

>  The admin can see any user?

Yes. There's not much point in protecting the user ids, as they're equivalent to kerberos ids and can be easily looked up. 

>    if yes, how?

For now, I think users just enter an id. If it doesn't exist in LORE, they get an error (similar to what edX does). 

We might want to add an invitation feature later. 

>    the user to add needs to login at least once before s/he can be added (to be in the local Django DB)?

I think so. 

> Should we split this in at least 2 stories:
>         api
>         views/templates 

Will do. This issue will be the API and I'll open a new one for the views/templates

giocalitri:
this is blocked by PR #106

giocalitri:
Do we need a new permission for this? Something like "manage_repo_users" to add only to the administrator group repo?
Otherwise what should I check to see who has permission to mange users: only the fact that the user belong to the repo admin group?
cc @Ferdi @pdpinch

Ferdi:
It sound like an implementation detail, no ?  
Yes,  repo admins will have the right to manage users but not sure what's the best way to implement a) adding a "manage users" permission and assigning to "admin" role or b) simply checking if user belong to 'admin' role

giocalitri:
any preference on the URL for the API?

Ferdi:
@giocalitri https://goo.gl/qU2PIy ?

pdpinch:
Functionally, this is done. The API docs #295 still need to be merged.

PR #208, opened 2015-06-22
members api #2 [#96516348]

giocalitri:
This PR is the second part of #129 and includes the roles APIs necessary for the rest views

### DIFF
--- a/roles/api.py
+++ b/roles/api.py
@@ -8,7 +8,10 @@ from django.contrib.auth.models import Group, Permission
 from django.db import transaction
 from guardian.shortcuts import assign_perm, get_perms, remove_perm
 
-from roles.permissions import RepoPermission, GroupTypes
+from roles.permissions import (
+    RepoPermission, GroupTypes, BaseGroupTypes, InvalidGroupType
+)
+from roles.user_models import UserGroup
 
 
 def roles_init_new_repo(repo):
@@ -162,3 +165,62 @@ def remove_user_from_repo_group(
         repo_group = Group.objects.get(name=group_type.format(repo.slug))
         user.groups.remove(repo_group)
         user.save()
+
+
+def list_users_in_repo(repo, base_group_type=None):
+    """
+    Lists all the users in the repository groups
+    If the group type is specified, the list is limited to that group
+
+    Args:
+        repo (learningresources.models.Repository): repository used to extract
+            the right group to use
+        base_group_type (unicode): group type from
+            roles.permissions.BaseGroupTypes
+    Returns:
+        list (list of roles.user_models.UserGroup): list of users in one or
+            all the repository groups
+    """
+    users_groups = []
+    if base_group_type is not None:
+        if not BaseGroupTypes.is_base_group_type(base_group_type):
+            raise InvalidGroupType
+        base_group_types = [base_group_type]
+    else:
+        base_group_types = BaseGroupTypes.all_base_groups()
+    for base_group_type in base_group_types:
+        group = Group.objects.get(
+            name=GroupTypes.get_repo_groupname_by_base(
+                base_group_type
+            ).format(repo.slug)
+        )
+        users_groups += [
+            UserGroup(user.username, base_group_type)
+            for user in group.user_set.all()
+        ]
+    return users_groups
+
+
+def is_last_admin_in_repo(user, repo):
+    """
+    Checks if user is the last administrator in the repository.
+    It does not check if the user is an actual administrator and in that case
+    it will simply return False
+
+    Args:
+        user (django.contrib.auth.models.User): user
+        repo (learningresources.models.Repository): repository used to extract
+            the right group to use
+    Returns:
+        bool
+    """
+    admins_in_repo = list_users_in_repo(
+        repo,
+        base_group_type=BaseGroupTypes.ADMINISTRATORS
+    )
+    return (
+        UserGroup(
+            user.username,
+            BaseGroupTypes.ADMINISTRATORS
+        ) in admins_in_repo
+    ) and len(admins_in_repo) == 1
--- a/roles/permissions.py
+++ b/roles/permissions.py
@@ -5,13 +5,68 @@ Definition of Django custom permissions
 from __future__ import unicode_literals
 
 
+class PermissionException(Exception):
+    """Custom exception for permission"""
+
+
+class InvalidGroupType(PermissionException):
+    """Invalid group type"""
+
+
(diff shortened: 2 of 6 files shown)


roles/api.py
```diff
@@ -162,3 +164,35 @@ def remove_user_from_repo_group(
...
+
+def list_users_in_repo(repo, base_group_type=None):
+    """
+    Lists all the users in the repository groups
+    If the group type is specified, the list is limited to that group
+
+    Args:
+        repo (learningresources.models.Repository): repository used to extract
+            the right group to use
+        base_group_type (roles.permissions.BaseGroupTypes): group string
```
noisecapella: This should be `unicode`

roles/api.py
```diff
@@ -162,3 +164,35 @@ def remove_user_from_repo_group(
...
+    Returns:
+        list (list of User): list of users in one or all the repository groups
```
noisecapella: This should probably be (User, unicode), or something that says it's a tuple and what each item means

roles/api.py
```diff
@@ -162,3 +164,35 @@ def remove_user_from_repo_group(
...
+    """
+    users_groups = []
+    if base_group_type is not None:
+        if not BaseGroupTypes.is_base_group_type(base_group_type):
+            raise InvalidGroupType
+        base_group_types = [base_group_type]
+    else:
+        base_group_types = BaseGroupTypes.all_base_groups()
+    with transaction.atomic():
```
noisecapella: It looks like the transaction is unnecessary here

roles/api.py
```diff
@@ -162,3 +164,35 @@ def remove_user_from_repo_group(
...
+        for base_group_type in base_group_types:
+            group = Group.objects.get(
+                name=GroupTypes.get_repo_groupname_by_base(
+                    base_group_type
+                ).format(repo.slug)
+            )
+            users_groups = users_groups + [
```
noisecapella: Can you use `+=` here? It looks cleaner to me and may be slightly more efficient but feel free to leave it if you disagree

roles/permissions.py
```diff
@@ -5,13 +5,63 @@
 from __future__ import unicode_literals
 
 
+class PermissionException(Exception):
+    """Custom exception for permission"""
+    pass
```
noisecapella: I don't think you need `pass` here or below since you have the docstring

roles/permissions.py
```diff
@@ -5,13 +5,63 @@
...
+        BaseGroupTypes.CURATORS: REPO_CURATOR,
+        BaseGroupTypes.AUTHORS: REPO_AUTHOR
+    }
+
+    @classmethod
+    def get_repo_groupname_by_base(cls, base_group_type):
+        """
+        Returns the basic group name from the repo name
+        """
+        return cls._GROUP_ASSOC.get(base_group_type)
```
noisecapella: Can you use the square brackets instead? That would raise an exception if they make a mistake and get the name wrong

roles/tests/test_api.py
```diff
@@ -280,3 +283,105 @@ def test_clear_permissions(self):
...
+        )
+
+    def test_list_users_in_repo_invalid_group_type_specified(self):
+        """
+        Test for list_users_in_repo
+        """
+        self.assertRaises(
+            api.InvalidGroupType,
+            lambda: api.list_users_in_repo(self.repo, 'foo_type')
+        )
```
noisecapella, line 440: Can you add tests for these cases: 
- user is assigned to same group twice (may be better put in roles/test_api.py)
- create a new repo without any users assigned to it at all and verify no users listed for it

pdpinch:
@giocalitri when you get a chance, can you update the description to describe what this PR does?

roles/tests/test_api.py
giocalitri, line 440: what do you mean by "may be better put in roles/test_api.py"?

noisecapella, line 440: For some reason I thought this was a different file. Yes I meant the same file we're commenting on :)

giocalitri, line 440: adding them as steps of another test

roles/api.py
```diff
@@ -162,3 +165,61 @@ def remove_user_from_repo_group(
...
+
+def list_users_in_repo(repo, base_group_type=None):
+    """
+    Lists all the users in the repository groups
+    If the group type is specified, the list is limited to that group
+
+    Args:
+        repo (learningresources.models.Repository): repository used to extract
+            the right group to use
+        base_group_type (roles.permissions.BaseGroupTypes): group unicode
```
noisecapella: Can you change the type for this to unicode? If we ever use docstring inspection in IDEs or in documentation generation exact types would be useful

roles/user_models.py
```diff
@@ -0,0 +1,29 @@
...
+        self.group_type = group_type
+
+    def __repr__(self):
+        return "<username:{username} | group_type:{group_type}>".format(
+            username=self.username,
+            group_type=self.group_type
+        )
+
+    def __eq__(self, other):
+        try:
```
noisecapella, line 22: Is it possible to just compare type and `__dict__` here? It would avoid the need for catching an exception and would adapt better to changes in the object

roles/user_models.py
```diff
@@ -0,0 +1,29 @@
...
+            return (self.username == other.username and
+                    self.group_type == other.group_type)
+        except AttributeError:
+            return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
noisecapella, line 30: May need hash function

roles/user_models.py
ShawnMilo, line 22: Explicit is better than implicit. There could be other differences unforeseen, and it would no longer be obvious which attributes we care about. I also think this use of `except AttributeError` is good, and not to be avoided.

It _could_ be good to add a check that `other` is also a `UserGroup`, but that could be overkill.

giocalitri, line 22: I have no strong feelings about either solution

roles/user_models.py
giocalitri, line 30: I made some research and my understanding is that I should not need it.

noisecapella:
Looks good, will merge when tests pass

giocalitri:
rebased master and pushed again

giocalitri:
I just found that I missed a test: fixing now

giocalitri:
Fixed and pushed: @noisecapella feel free to merge when the tests pass

PR #209, opened 2015-06-22
Fixed IntegrityError when 'required' attribute is missing from Vocabulary

noisecapella:
Django REST framework doesn't seem to enforce the `required` field correctly for `BooleanField`. This PR overrides the `BooleanField` to skip a check with `rest_framework.utils.html.is_html_input`. That function assumes that if the data passed in is a MultiDict type datastructure, that means the data came from an HTML form, which isn't true for the test below.


giocalitri:
LGTM: Bug functionally tested

PR #222, opened 2015-06-23
Converted to using diffable test courses

### LINKED ISSUE #218: As a developer I would like the test course(s) to be easily diffable
I plan on making them extract and have the methods to get them zip them on demand

carsongee:
Closes #218
[finished #97661472]

@pwilkins to make adding and testing parsers easier and such. If this merges before #199 then that will need a rebase, or vice versa.


carsongee:
Funny on the coverage.  That line is covered in py33 and py34 which ran later.  You would think it would have shown as covered.  It intentionally isn't covered in py27

learningresources/tests/base.py
```diff
@@ -155,5 +177,5 @@ def get_course_single_tarball(self):
         Returns:
             path (unicode): absolute path to tarball.
         """
-        path = join(abspath(dirname(__file__)), "testdata", "courses")
-        return self.copy_file(join(path, "single.tgz"))
+        path = join(abspath(dirname(__file__)), "testdata", "courses", "toy")
+        return self._make_archive(path, False, '.tgz')
```
pwilkins, line 183: I need to read the code more closely, but here the extension is `tgz` and yet in `make_archive()` the default is `tar.gz`.  Should these both be the same, or is there something that make these two unrelated?

ShawnMilo, line 183: `.tar.gz` and `.tgz` are interchangeable. There is no difference. I did a quick Google search and found one claim that `.tgz` came from old filename length limits in DOS. 

Either way, it's fine. You can change the extension of a `.tgz` to a `.tar.gz` (or vice versa) and still decompress it with the same command and get the same files out.

For that matter, `.xlsx` files are just zip files; despite the name, you can `unzip` them.

carsongee, line 183: so the make_archive() doesn't actually have anything to do with the file name that is under test.  It is just done in a temp directory which is saved to the filesystem for use by all code as the filename defined at https://github.com/mitodl/lore/pull/222/files#diff-589006e737018a1af16a1e95dbf5f35bR137.  So the ext parameter (with defaults to .zip and .tar.gz respectively) is what determines the extension under test, and exercises our extension verifying code in the import.  (From the standup, and these comments, I wanted to double check that was true which is why I asked you to hold off on the merge).

learningresources/tests/base.py
```diff
@@ -121,7 +141,7 @@ def copy_file(self, path):
                 random=str(uuid.uuid4()),
                 ext=ext
             ),
-            open(path, 'rb')
+            open(new_archive, 'rb')
```
pwilkins, line 144: Question for my own information: Why open the file here?

carsongee, line 144: Because django.core.storage requires an open file handle to save and open.

PR #225, opened 2015-06-24
Status page for a health-check of server dependencies. [#1342898]

### LINKED ISSUE #196: As an ops person I would like a health check URL
In order to do easy status and health checking for out instance, it would be convenient to have a url that returns some json and a relevant HTTP status code after checking that everything is working as expected (DB, Redis, ES, etc all available to the app).

ShawnMilo:
Currently covers:
- PostgreSQL
- Redis
- Elasticsearch

closes #196 

Sample output:

``` json
{
    "elasticsearch": {
        "response_ms": 44852,
        "status": "up",
        "status_code": 200
    },
    "postgresql": {
        "response_ms": 1580,
        "status": "up"
    },
    "redis": {
        "response_ms": 599,
        "status": "up",
        "uptime_in_seconds": 9,
        "used_memory": 878312,
        "used_memory_peak": 930208
    }
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,14 @@ python:
   - '2.7'
 services:
   - elasticsearch
+  - redis-server
 before_script:
   - sleep 10
 env:
   global:
     - DATABASE_URL=postgres://postgres@localhost:5432/postgres
+    - BROKER_URL=redis://localhost:6379/4
+    - HAYSTACK_URL=127.0.0.1:9200
 install:
   - "pip install 'tox>=2.0.2,<3.0.0'"
   - pip install coveralls
--- a/status/tests/test_status.py
+++ b/status/tests/test_status.py
@@ -1,0 +1,109 @@
+"""
+Tests for the status module.
+"""
+from __future__ import unicode_literals
+
+from copy import deepcopy
+import json
+import logging
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.test import Client
+from django.test.testcases import TestCase
+
+log = logging.getLogger(__name__)
+
+HTTP_OK = 200
+SERVICE_UNAVAILABLE = 503
+
+
+class TestStatus(TestCase):
+    """Test output of status page."""
+
+    def setUp(self):
+        """
+        Set up client and remember settings for PostgreSQL, Redis,
+        and Elasticsearch. Changes made to django.conf.settings during
+        tests persist beyond the test scope, so if they are changed
+        we need to restore them.
+        """
+
+        # Create test client.
+        self.client = Client()
+        self.url = reverse("status")
+        super(TestStatus, self).setUp()
+
+        # Remember settings in case they're mutilated during a test.
+        self.original_settings = (
+            settings.BROKER_URL,
+            deepcopy(settings.DATABASES),
+            deepcopy(settings.HAYSTACK_CONNECTIONS),
+        )
+
+    def tearDown(self):
+        """Restore settings after each test."""
+        (
+            settings.BROKER_URL,
+            settings.DATABASES,
+            settings.HAYSTACK_CONNECTIONS,
+        ) = self.original_settings
+        super(TestStatus, self).tearDown()
+
+    def get(self, expected_status=HTTP_OK):
+        """Get the page."""
+        resp = self.client.get(self.url, data={"token": settings.STATUS_TOKEN})
+        self.assertEqual(resp.status_code, expected_status)
+        jsn = json.loads(resp.content.decode('utf-8'))
+        log.info(jsn)
+        return jsn
+
+    def test_view(self):
+        """Get normally."""
+        resp = self.get()
+        for key in ("postgresql", "redis", "elasticsearch"):
+            self.assertTrue(resp[key]["status"] == "up")
+
+    def test_no_settings(self):
+        """Missing settings."""
+        del settings.BROKER_URL
+        del settings.DATABASES
+        del settings.HAYSTACK_CONNECTIONS
+        resp = self.get()
+        for key in ("postgresql", "redis", "elasticsearch"):
+            self.assertTrue(resp[key]["status"] == "no config found")
+
+    def test_broken_settings(self):
+        """Settings that couldn't possibly work."""
+        junk = " not a chance "
+        settings.BROKER_URL = junk
+        settings.DATABASES["default"] = junk
+        settings.HAYSTACK_CONNECTIONS["default"]["URL"] = junk
+        resp = self.get(SERVICE_UNAVAILABLE)
+        for key in ("postgresql", "redis", "elasticsearch"):
+            self.assertTrue(resp[key]["status"] == "down")
+
+    def test_invalid_settings(self):
+        """
+        Settings that look right, but aren't (if service is actually down).
+        """
+        settings.BROKER_URL = "redis://bogus:6379/4"
+        settings.DATABASES["default"]["HOST"] = "monkey"
+        settings.HAYSTACK_CONNECTIONS["default"]["URL"] = "pizza:2300"
+        resp = self.get(SERVICE_UNAVAILABLE)
+        for key in ("postgresql", "redis", "elasticsearch"):
+            self.assertTrue(resp[key]["status"] == "down")
+
+    def test_token(self):
+        """
+        Caller must have correct token, or no dice. Having a good token
+        is tested in all the other tests.
+        """
+
+        # No token.
+        resp = self.client.get(self.url)
+        self.assertTrue(resp.status_code == 404)
+
+        # Invalid token.
+        resp = self.client.get(self.url, {"token": "gibberish"})
+        self.assertTrue(resp.status_code == 404)
--- a/status/views.py
+++ b/status/views.py
@@ -1,0 +1,130 @@
+"""
+Status checks for LORE.
+
+Notes:
+
+* Useful messages are logged, but NO_CONFIG is returned whether
+  settings are missing or invalid, to prevent information leakage.
+* Different services provide different information, but all should return
+  "up," DOWN, or NO_CONFIG for the "status" key.
+"""
+
+from datetime import datetime
+import logging
+
+from django.conf import settings
+from django.http import JsonResponse, Http404
+from elasticsearch import Elasticsearch, ConnectionError as ESConnectionError
+from kombu.utils.url import _parse_url as parse_redis_url
+from psycopg2 import connect, OperationalError
+from redis import StrictRedis, ConnectionError as RedisConnectionError
+
+log = logging.getLogger(__name__)
+
+UP = "up"
+DOWN = "down"
+NO_CONFIG = "no config found"
+HTTP_OK = 200
+SERVICE_UNAVAILABLE = 503
+TIMEOUT_SECONDS = 5
+
+
+def get_pg_info():
+    """Check PostgreSQL connection."""
+    log.debug("entered get_pg_info")
+    try:
+        conf = settings.DATABASES['default']
+        database = conf["NAME"]
+        user = conf["USER"]
+        host = conf["HOST"]
+        port = conf["PORT"]
+        password = conf["PASSWORD"]
+    except (AttributeError, KeyError):
+        log.error("No PostgreSQL connection info found in settings.")
+        return {"status": NO_CONFIG}
+    except TypeError:
+        return {"status": DOWN}
+    log.debug("got past getting conf")
+    try:
+        start = datetime.now()
+        connection = connect(
+            database=database, user=user, host=host,
+            port=port, password=password, connect_timeout=TIMEOUT_SECONDS,
+        )
+        log.debug("at end of context manager")
+        micro = (datetime.now() - start).microseconds
+        connection.close()
+    except (OperationalError, KeyError):
+        log.error("Invalid PostgreSQL connection info in settings: %s", conf)
+        return {"status": DOWN}
+    log.debug("got to end of postgres check successfully")
+    return {"status": UP, "response_microseconds": micro}
+
+
+def get_redis_info():
+    """Check Redis connection."""
+    try:
+        url = settings.BROKER_URL
+        _, host, port, _, _, db, _ = parse_redis_url(url)
+    except AttributeError:
+        log.error("No valid Redis connection info found in settings.")
+        return {"status": NO_CONFIG}
+
+    start = datetime.now()
+    try:
+        rdb = StrictRedis(
+            host=host, port=port, db=db, socket_timeout=TIMEOUT_SECONDS)
+        info = rdb.info()
+    except (RedisConnectionError, TypeError):
+        return {"status": DOWN}
+    micro = (datetime.now() - start).microseconds
+    del rdb  # the redis package does not support Redis's QUIT.
+    ret = {
+        "status": UP, "response_microseconds": micro,
+    }
+    fields = ("uptime_in_seconds", "used_memory", "used_memory_peak")
+    ret.update({x: info[x] for x in fields})
+    return ret
+
+
+def get_elasticsearch_info():
+    """Check Elasticsearch connection."""
+    try:
+        url = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
+    except (AttributeError, KeyError):
+        log.error("No Elasticsearch connection info in settings.")
+        return {"status": NO_CONFIG}
+    start = datetime.now()
+    try:
+        search = Elasticsearch(url, request_timeout=TIMEOUT_SECONDS)
+        info = search.info()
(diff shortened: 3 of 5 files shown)


ShawnMilo:
@cg: FYI

pdpinch:
@bdero or @carsongee can you review?

.travis.yml
```diff
@@ -3,10 +3,14 @@ python:
   - '2.7'
 services:
   - elasticsearch
+  - redis-server
 before_script:
   - sleep 10
 env:
-  - DATABASE_URL=postgres://postgres@localhost:5432/postgres
+  global:
```
carsongee, line 10: Why did you add global? We aren't in a build matrix, so shouldn't need to specify that.

status/tests/test_status.py
```diff
@@ -0,0 +1,80 @@
...
+from django.test.testcases import TestCase
+
+log = logging.getLogger(__name__)
+
+
+class TestStatus(TestCase):
+    """Test output of status page."""
+
+    def setUp(self):
+        """Set up client."""
```
carsongee: Pleas add a docstring as to what you are doing here, and maybe a note saying, this saves and restores settings x,y,z in the class docstring so people know what is ok to mess with and what isn't later on.

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+log = logging.getLogger(__name__)
+
+
+def get_pg_info():
+    """Check PostgreSQL connection."""
+    try:
+        conf = settings.DATABASES['default']
+    except (AttributeError, KeyError):
+        log.error("No PostgreSQL connection info found in settings.")
+        return {"status": "no config found"}
```
carsongee: Please move all the messages to a shared dictionary or tuple to reduce chance of error and ease the ability to change status messages across services

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+        start = datetime.now()
```
carsongee, line 49: Please use https://docs.python.org/2/library/timeit.html for the time to increase resolution when possible

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+from redis import StrictRedis, ConnectionError as RedisConnectionError
+
+log = logging.getLogger(__name__)
+
+
+def get_pg_info():
+    """Check PostgreSQL connection."""
+    try:
+        conf = settings.DATABASES['default']
+    except (AttributeError, KeyError):
```
carsongee, line 42: I actually want these to raise up.  We shouldn't be catching anything in a health check in general, since the status code is what most health checkers use to determine health (the json is just a nice to have, but the status code is the critical thing)

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+        log.error("No PostgreSQL connection info found in settings.")
+        return {"status": "no config found"}
+    try:
+        start = datetime.now()
+        connect(
+            database=conf["NAME"], user=conf["USER"], host=conf["HOST"],
+            port=conf["PORT"], password=conf["PASSWORD"]
+        )
+        micro = (datetime.now() - start).microseconds
+    except (OperationalError, KeyError):
```
carsongee, line 57: again, 500 preferred, so not catching

status/views.py
carsongee, line 49: I think all these timers would be better done with a context manager, it will reduce repeated code and let us do a `finally` statement for these

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+        return {"status": "down"}
+    return {"status": "up", "response_ms": micro}
+
+
+def get_redis_info():
+    """Check Redis connection."""
+    pat = re.compile(r'redis://([^:]+):(\d+)/(\d+)$')
+    try:
+        url = settings.BROKER_URL
+    except AttributeError:
```
carsongee, line 69: same on the exception advice, but I wonder if all this parsing is needed, surely there must be something in celery that already does this that we can reuse

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+        url = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
+    except (AttributeError, KeyError):
+        log.error("No Elasticsearch connection info in settings.")
+        return {"status": "no config found"}
+    match = pat.match(url)
+    if match is None:
+        log.error("Invalid Elasticsearch connection info: %s.", url)
+        return {"status": "no config found"}
+    host, port = match.groups()
+    search = Elasticsearch(hosts=[{"host": host, "port": port}])
```
carsongee: Elasticsearch can just take the string as it is in haystack so you don't need to parse it at all.  However, I would check the `HAYSTACK_CONNECTIONS` default and make sure it is of the ElasticSearch type above before testing, so that we could change it to solr for example, and this check wouldn't run

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+    start = datetime.now()
+    try:
+        info = rdb.info()
+    except RedisConnectionError:
+        return {"status": "down"}
+    micro = (datetime.now() - start).microseconds
+    ret = {
+        "status": "up", "response_ms": micro,
+    }
+    fields = ("uptime_in_seconds", "used_memory", "used_memory_peak")
```
carsongee, line 85: Don't we need to close the connection?  We are connection limited by purchase and I would hate to have the connection pool go to infinity for health checks

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+        start = datetime.now()
+        connect(
+            database=conf["NAME"], user=conf["USER"], host=conf["HOST"],
+            port=conf["PORT"], password=conf["PASSWORD"]
+        )
+        micro = (datetime.now() - start).microseconds
+    except (OperationalError, KeyError):
+        log.error("Invalid PostgreSQL connection info in settings: %s", conf)
+        return {"status": "down"}
```
carsongee: Same here, I think we need a disconnect in a finally type statement to prevent connection pool exhaustion

status/views.py
```diff
@@ -0,0 +1,107 @@
...
+    micro = (datetime.now() - start).microseconds
+    return {
+        "status": "up", "response_ms": micro, "status_code": info["status"]
+    }
+
+
+def status(request):  # pylint: disable=unused-argument
+    """Status"""
+    info = {}
```
carsongee, line 116: So I was thinking, since connections are a limited resource, and these do "cost" us.  It would be nice to just do a simple query string type token that is defined in settings to protect this page (i.e. just return a 403 or 404).  Health checkers take a URL and all of the ones I've run into honor the `?=` querystring type params

.travis.yml
ShawnMilo, line 10: [Travis docs](http://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-.travis.yml)

Originally I did this:

``` yaml
env:
  - DATABASE_URL=postgres://postgres@localhost:5432/postgres
  - BROKER_URL=redis://localhost:6379/4
  - HAYSTACK_URL=127.0.0.1:9200
```

However, this doesn't work because Travis won't use all three at once. It does three runs using one each. So the first run through, there's no config for Redis or Elasticsearch.

This also works:

``` yaml
env:
  - DATABASE_URL=postgres://postgres@localhost:5432/postgres BROKER_URL=redis://localhost:6379/4 HAYSTACK_URL=127.0.0.1:9200
```

Both have the same effect, but I think the one-liner is uglier.

status/views.py
ShawnMilo: I don't understand what is desired here. The values are being put into a shared dictionary, and I don't see where the potential is to reduce error. Please elaborate.

status/views.py
ShawnMilo, line 49: I agree with the context manager, but would like argue on `timeit`. It moves code into strings to be `eval`ed, which feels dirty for production code. It looks like it's meant for benchmarks, not for production use. Increasing resolution from microseconds is unnecessary anyway, as network latency will be a far bigger factor in variation than anything else.

status/views.py
ShawnMilo, line 42: Are you saying that if everything else is functional, but Redis is down, you want a 500 error, and that you'd rather not know that the problem was Redis (or PostgreSQL or Elasticsearch)?

It seems that "something is broken" is not useful, and using one of the many free monitoring services would serve that purpose nicely. There's no point in this branch at all if that's the case. Knowing that the Elasticsearch server is down would give a much bigger head-start on fixing the problem.

status/views.py
ShawnMilo, line 69: Efforts were made here to directly connect to PostgreSQL, bypassing the ORM, to Redis, bypassing Celery, and to Elasticsearch, bypassing Haystack, for a true check on the status of the backend servers. This is better than using the intermediary, and is also useful for when a server (such as Redis) is being used by more than one plugin.

status/views.py
ShawnMilo: I tried this and it doesn't work. If you have a working example I'll change it.

As for changing to solr, I think that's YAGNI until we do.

status/views.py
ShawnMilo, line 85: Great point. I will close the connections.

status/views.py
ShawnMilo, line 116: Okay, cool. The connection is going to be SSL everywhere, right? Just wanting to make sure the querystring can't be intercepted.

status/views.py
ShawnMilo, line 49: After adding a context manager, it results in more complexity than it's worth. The context manager returns a value on `__enter__`, not `__exit__`, but it must have an `__exit__` method or it's invalid. The timer would be kicked off in the `__enter_`, and that's where the value must be returned from, so it gets ugly and it seems like this isn't actually a good use for a context manager.

If you think I'm missing the approach you had in mind, please send a link to a context manager that works the way you were thinking and I'll give it another go.

status/views.py
carsongee: a `STATUSES` var at the top, with `no config found` (except that one isn't needed I don't think), `up`, and `down`. Then you can just use that everywhere you need to set status.

status/views.py
carsongee, line 49: I was thinking you would use a dictionary with a `start` and `stop` that would register the times, and then that could be used outside the context, i.e.

``` python
@contextmanager
def timer(thing):
     try:
         thing['start'] = 1
         yield thing
     finally:
         thing['stop'] = 2
```

and then called like:

``` python
timer = {}
with timer(timer):
    ...
print(timer['stop'] - timer['start'])
```

but probably could be a lot smarter than that.

status/views.py
carsongee, line 69: I agree on the bypassing, I just mean that they already probably have a function to parse the URL

status/views.py
carsongee: Fair, on typing, you can skip that. You just need to pass it as:

``` python
Elasticsearch([settings.HAYSTACK_URL])
```

from http://elasticsearch-py.readthedocs.org/en/master/api.html#elasticsearch

Real world example: https://github.com/carsongee/archelon/blob/master/archelond/archelond/tests/config/elastic.py#L17 and https://github.com/carsongee/archelon/blob/master/archelond/archelond/data/elastic.py#L39-L41

status/views.py
carsongee, line 49: That is cool on timeit, millis are probably fine, let alone micro.

ShawnMilo:
Thanks for the updates, @carsongee. Is this ready to be moved back to "Waiting on Author"?

carsongee:
Sorry, yeah. I forgot to flip the flag

status/views.py
carsongee, line 42: I want both, so a 500 result returned, and the JSON with the details, that way the Load Balancer knows the service is unhealthy (say only one of 3 instances can't connect to one of these services, it should be removed from the LB), but our more sophisticated monitoring (i.e. me, or a script designed to use this page) will also want the full status JSON to know of a specific instance. So I guess we don't want to raise, but we do need to still send a 500

status/views.py
carsongee, line 69: To be more specific: https://github.com/celery/celery/blob/a7c168d6df82bcd060e02e647a916d61627959dc/celery/backends/redis.py#L59 and use connparams from that instantiated class to do the connection or just use its client property.

.travis.yml
carsongee, line 10: Cool, thanks for the explanation.

status/views.py
ShawnMilo, line 42: As discussed in meatspace, we're going to return a [503](http://www.checkupdown.com/status/E503.html) status code and the JSON.

ShawnMilo, line 42: This is now addressed in the code & tests.

status/views.py
```diff
@@ -0,0 +1,121 @@
...
+    try:
+        url = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
+    except (AttributeError, KeyError):
+        log.error("No Elasticsearch connection info in settings.")
+        return {"status": NO_CONFIG}
+    search = Elasticsearch(url)
+    start = datetime.now()
+    try:
+        info = search.info()
+    except ESConnectionError:
```
carsongee, line 101: With ES turned off in docker-compose, this just causes it to seemingly infinitely retry, so the page never renders (and would hit the load balancer timeout if it was in production)

status/views.py
```diff
@@ -0,0 +1,121 @@
...
+        match = None
+    if match is None:
+        log.error("Invalid Redis connection info found in settings: %s", url)
+        return {"status": NO_CONFIG}
+    host, port, database = match.groups()
+    rdb = StrictRedis(host=host, port=port, db=database)
+    start = datetime.now()
+    try:
+        info = rdb.info()
+    except RedisConnectionError:
```
carsongee: I appear to get the same result with Redis, removing the redis link in docker-compose appears to cause the page to infinitely try.

carsongee: It came back after 2 minutes with the correct response, but that is 1.5 minutes more than the current LB timeout.

status/views.py
ShawnMilo, line 101: I'm also getting a two-minute timeout with `psycopg2` during one test. However, there don't seem to be any timeout kwargs available for any of those libraries.

A quick Google search suggests the best solution is the [signals](https://docs.python.org/2/library/signal.html) package.

pwilkins:
@ShawnMilo Are these test failures significant and you have more work to do? Should I flip the switch back to "Waiting on Author"?

ShawnMilo:
Yeah, it looks like the failure on Python 3 is legit. I'll fix locally and re-push.

ShawnMilo:
Tests are all passing locally now; not to see if Travis agrees.

status/views.py
```diff
@@ -0,0 +1,158 @@
...
+    try:
+        url = settings.BROKER_URL
+        _, host, port, _, _, db, _ = parse_redis_url(url)
+    except AttributeError:
+        log.error("No valid Redis connection info found in settings.")
+        return {"status": NO_CONFIG}
+
+    start = datetime.now()
+    try:
+        with Timeout(2):
```
carsongee: Can we make this configurable with a default, or at least a constant at the top of the file to make it easier to change later if 2 seconds proves to be problematic?

carsongee:
Got this with redis turned off:
image

carsongee:
Running off of uwsgi instead of the development server appears to resolve the issue and works as expected.  If you can't figure out an easy way around that, I think this is still probably good enough after the magic number fix mentioned.

ShawnMilo:
Magic number & merge conflict with `master` addressed. Still needs flattening if it passes review. I don't have any solution for the signal thing since _it_ is the only solution I found for our other problem.

Specifically, since the Redis, Elasticsearch, and PostgreSQL drivers don't have any timeout kwargs available, I don't know how else to handle it other than letting them time out. I'm open to suggestions.

carsongee:
There is one for redis 2.10 and  greater: https://github.com/andymccurdy/redis-py/blob/81429b44acff3f238a212a1651241b49cde76c85/redis/connection.py#L388, and it looks like elasticsearch can do that as well: https://elasticsearch-py.readthedocs.org/en/master/api.html?highlight=timeout#timeout, and for psycopg2 as well: http://stackoverflow.com/questions/27641740/python-psycopg2-timeout

Though that is just from some googling, your mileage may vary.

ShawnMilo:
I Googled those things a couple of weeks ago and came up short. My Google-fu must be wanting.

ShawnMilo:
Timeout args added, `signals` removed. Tests currently passing.

carsongee:
:purse: has an extra merge commit in there, but feel free to merge after commit is clean, and tests pass. Also, thanks.  This is really cool and I really learned a lot, sorry it took so long.

ShawnMilo:
Excellent. Thanks for holding it to your usual high standards. I learned some cool stuff too. Like you can return data with 500 errors. :o)

PR #229, opened 2015-06-24
Adds abstract base model to contain audit fields.

### LINKED ISSUE #134: Add audit fields to all tables.
As a developer who develops and supports LORE, I'd like automatic timestamps added to each record when it is created and when it is modified. This is simple with Django models by creating a subclass of `models.Model` which has [datetime](https://docs.djangoproject.com/en/1.8/ref/models/fields/#datetimefield) fields using the `auto_now` and `auto_now_add` options, then making all of our models a subclass of the "audit class" instead of `models.Model`.

This has been discussed in #81. Initially we just want the timestamp fields, but may add a UUID field. This UUID field will not be the primary key, but foreign keys _could_ switch to using this UUID field from the primary key, making relationships explicit even when exported outside LORE.

ShawnMilo:
Closes #134

Adds datetime fields for created and modified dates.

Removes older `auto_now_add` fields with inconsistent names in favor of the universal ones. Contains schema and data migrations to support this.

### DIFF
--- a/audit/models.py
+++ b/audit/models.py
@@ -1,0 +1,13 @@
+"""A base class for all models with audit fields."""
+
+from django.db import models
+
+
+class BaseModel(models.Model):
+    """A models.Model with audit fields."""
+    date_created = models.DateTimeField(auto_now_add=True)
+    date_modified = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        """Do not create a database table for BaseModel."""
+        abstract = True
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -54,7 +54,7 @@ def create_course(org, repo_id, course_number, run, user_id):
 
     """
     # Check on unique values before attempting a get_or_create, because
-    # items such as import_date will always make it non-unique.
+    # items such as date_created  will always make it non-unique.
     unique = {
         "org": org, "course_number": course_number, "run": run,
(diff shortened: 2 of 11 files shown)


learningresources/migrations/0004_auto_adds_audit_fields.py
```diff
@@ -0,0 +1,67 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('learningresources', '0003_add_permissions'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='course',
+            name='date_created',
```
noisecapella: Is it possible to use `RenameField` for these migrations instead?

noisecapella:
Looks good overall, just the one question

learningresources/migrations/0004_auto_adds_audit_fields.py
ShawnMilo: I don't believe so, only because it's being added by the abstract base class, not by modifying this model directly.

noisecapella: Sounds good, I'll merge this

noisecapella:
FYI I needed to do some manual migration merging with the static assets stuff so it's a bit different now. I'll merge after the tests rerun

PR #233, opened 2015-06-25
Implemented taxonomy model delete cascading [#96563502]

### LINKED ISSUE #136: Cascade deletes on vocabularies and terms
This should remove `on_delete=models.PROTECT` from the models so that deletes will cascade. I talked with @cg about delete functionality and this is what he prefers in order to maintain database integrity. I think the alternative is doing the deletes by hand which may be more error prone. Any opinions @ShawnMilo?

noisecapella:
Fixes #136

### DIFF
--- a/rest/tests/test_rest.py
+++ b/rest/tests/test_rest.py
@@ -241,13 +241,11 @@ class TestRest(RESTTestCase):
             self.repo.slug, vocab1_dict['slug'])
         self.get_vocabulary(
             other_repo_dict['slug'], vocab1_dict['slug'],
-            expected_status=HTTP_404_NOT_FOUND,
-            skip_options_head_test=True  # vocab permission work not yet done
+            expected_status=HTTP_404_NOT_FOUND
         )
         self.get_vocabulary(
             self.repo.slug, vocab2_dict['slug'],
-            expected_status=HTTP_404_NOT_FOUND,
-            skip_options_head_test=True  # vocab permission work not yet done
+            expected_status=HTTP_404_NOT_FOUND
         )
         self.get_vocabulary(
             other_repo_dict['slug'], vocab2_dict['slug']
@@ -255,8 +253,8 @@ class TestRest(RESTTestCase):
 
     def test_term(self):
         """Test REST access for term"""
-        vocab_slug = self.create_vocabulary(self.repo.slug)['slug']
-        terms = self.get_terms(self.repo.slug, vocab_slug)
+        vocab1_slug = self.create_vocabulary(self.repo.slug)['slug']
+        terms = self.get_terms(self.repo.slug, vocab1_slug)
         self.assertEqual(0, terms['count'])
 
         input_dict = {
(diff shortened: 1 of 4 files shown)


rest/tests/test_rest.py
```diff
@@ -255,8 +256,8 @@ def test_other_vocabulary(self):
 
     def test_term(self):
         """Test REST access for term"""
-        vocab_slug = self.create_vocabulary(self.repo.slug)['slug']
-        terms = self.get_terms(self.repo.slug, vocab_slug)
+        vocab1_slug = self.create_vocabulary(self.repo.slug)['slug']
```
carsongee, line 256: Why the numbering?

noisecapella, line 256: To make it clearer what I mean down below when it tests both vocabularies and both terms

rest/tests/test_rest.py
```diff
@@ -364,15 +365,31 @@ def test_term(self):
...
         self.logout()
-        self.get_term(self.repo.slug, vocab_slug, term1['slug'],
+        self.get_term(self.repo.slug, vocab1_slug, term1['slug'],
                       expected_status=HTTP_403_FORBIDDEN)
-        self.get_term(self.repo.slug, vocab_slug, term2['slug'],
+        self.get_term(self.repo.slug, vocab1_slug, term2['slug'],
                       expected_status=HTTP_403_FORBIDDEN)
         self.get_term(self.repo.slug, vocab2_slug, term1['slug'],
                       expected_status=HTTP_403_FORBIDDEN)
         self.get_term(self.repo.slug, vocab2_slug, term2['slug'],
```
carsongee, line 375: oh, because there is a vocab2

rest/tests/test_rest.py
```diff
@@ -364,15 +365,31 @@ def test_term(self):
...
+        # vocab1 was deleted and term1 is now also deleted
+        self.get_term(self.repo.slug, vocab1_slug, term1['slug'],
+                      expected_status=HTTP_404_NOT_FOUND)
+        self.get_term(self.repo.slug, vocab1_slug, term2['slug'],
+                      expected_status=HTTP_404_NOT_FOUND,
+                      skip_options_head_test=True)
+        self.get_term(self.repo.slug, vocab2_slug, term1['slug'],
+                      expected_status=HTTP_404_NOT_FOUND,
+                      skip_options_head_test=True)
+        self.get_term(self.repo.slug, vocab2_slug, term2['slug'])
```
carsongee: Maybe a comment saying that this is asserting the other term was not deleted in the cascade (unless I'm misreading the test).  It might be nice to have more descriptive names than vocab1 and vocab2.  Maybe something to describe what makes them different, or descriptive of what they are being used to test.  The test is a bit hard for me to easily follow as is

rest/views.py
```diff
@@ -137,7 +138,7 @@ def get_queryset(self):
             slug=self.kwargs['vocab_slug']
         )
         if not vocabs.exists():
-            return []
+            raise Http404()
```
carsongee: This seems like it might be out of scope?  Changing a return to a raise seems like a pretty big API change to add in with cascaded deletes.

carsongee: And there don't seem to be any new tests to assert the new behavior

carsongee:
You'll also need to check the migration in here.  Running migrate on this branch results in:

```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

rest/views.py
noisecapella: I'll pull this into a separate PR

rest/tests/test_rest.py
```diff
@@ -27,6 +27,7 @@
 )
 
 
+# pylint: disable=too-many-statements
```
carsongee: This should be inside the class block instead of the file level I think

rest/tests/test_rest.py
carsongee: Still no comment here, and our clarifying my understanding.

noisecapella:
I've moved the delete cascade portion of the test to a separate test for clarity. I've also added some code removing the option to skip OPTION assertions because it's not needed anymore

carsongee:
Looks good.  The commit message needs to be changed, but after that and tests pass, I'll merge :watch:

pwilkins:
@noisecapella If you add the commit message so this fix will appear in the release notes, I can merge it.

PR #235, opened 2015-06-25
Switched to using per item type glyphs for learning resources

carsongee:
Closes #228
[finished #97740486]
image
Defaults to using the little XML things if the type isn't known.

PR #236, opened 2015-06-25
Styled left hand navigation pane with glyphs

carsongee:
Dependent on #235, so do not review until that merges.  This also hides Vocabularies that do not have terms in the listing.

image


pdpinch:
Re: `This also hides Vocabularies that do not have terms in the listing.`

FYI, @carsongee and I discussed briefly and in the interests of expediency we agreed that this was the best choice for handling cases where a vocabulary existed but no relevant terms had been applied to learning resources. 

A better solution may be to display a 0 count for each term in the vocabulary, but the search isn't configured to do that currently. @ferdi et al. if you agree this is preferable, please open a new issue & tag @ShawnMilo

Ferdi:
Sounds good.

PR #237, opened 2015-06-25
Reduce logging level of elasticsearch and urllib3

carsongee:


ShawnMilo:
Looks good; will merge once Travis is done with it.

bdero:
@ShawnMilo He's going to try changing the log level to info to see if that's alright

ShawnMilo:
So hold off on merging even if it passes?

bdero:
Nah, it should be good to merge after it passes. I forgot that the build takes forever and there was therefore no chance of it finishing before it was updated again. ;)

PR #220, opened 2015-06-23
Add JavaScript infrastructure to run unit tests

noisecapella:

PR #238, opened 2015-06-25
Update readme for JavaScript testing

noisecapella:
Something I forgot to include in #220 and don't want to hold up that build for


noisecapella:
Going to do this in #220 instead since I need to make another change

PR #240, opened 2015-06-25
Use get_perms for cleaner code

### LINKED ISSUE #239: ManageTaxonomyPermission in rest.permissions
The check currently is:

```
           manage_taxonomies_permission = '{}.{}'.format(
                Repository._meta.app_label,
                RepoPermission.manage_taxonomy[0]
            )
            return request.user.has_perm(manage_taxonomies_permission, repo)
```

instead it should be:

```
from guardian.shortcuts import get_perms 
RepoPermission.manage_taxonomy[0] in get_perms(request.user, repo)
```

the reason is that in the first case it does not check permissions in the right way in case the permissions are applied to the group.

noisecapella:
Fixes #239

### DIFF
--- a/rest/permissions.py
+++ b/rest/permissions.py
@@ -9,8 +9,8 @@ from rest_framework.permissions import (
     SAFE_METHODS,
 )
 from django.http import Http404
+from guardian.shortcuts import get_perms
 
-from roles.permissions import RepoPermission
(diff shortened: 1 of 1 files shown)


giocalitri:
Looks good: merging as soon as tests pass

Issue #163, opened 2015-06-15
UI: Edit LearningResource item


Ferdi:
UI for editing a learning resource

Depends on #162 . See also #161 
- [X] Slide out panel
- [X] XML tab 
- [x] add terms #298
- [x] edit description #299
- [x] Preview Tab
- [x] display analytics

![editlearningresource](https://cloud.githubusercontent.com/assets/125561/8151465/ec967c08-12db-11e5-81bc-de2481f6da7c.png)

PR #241, opened 2015-06-25
Added slide out panel for learning resource details [#96967708]

carsongee:
Dependent on #235 (mostly just for the indentation).
https://www.youtube.com/watch?v=29h7WsyrwKc&feature=youtu.be

### DIFF
--- a/ui/static/ui/js/main.js
+++ b/ui/static/ui/js/main.js
@@ -11,13 +11,30 @@ define(['jquery', 'bootstrap', 'icheck'], function(jQuery) {
     //open the lateral panel
     $('.cd-btn').on('click', function(event) {
       event.preventDefault();
+      var url = $(this).attr('href');
+      var title = $(this).html();
+      var textarea = $('.cd-panel .textarea-xml');
+      $('.cd-panel header h1').html(title);
+
+      // Clear out text area while waiting for AJAX to return
+      textarea.val('');
+      $.ajax({url: url, dataType: 'text'}).done(function(data) {
+        $('.cd-panel .textarea-xml').val(data);
+      }).error(function() {
+        $('.cd-panel .textarea-xml').val('Unable to retrieve XML.');
+      });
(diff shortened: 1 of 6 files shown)


pdpinch:
This is part of #163 right?

carsongee:
Yeah

noisecapella:
The copy to clipboard button doesn't work, in the javascript it only selects the text

noisecapella:
Also, is there any error reporting that should happen if the client can't connect to the server? Is error handling in the client handled in a future issue?

carsongee:
The copy to clipboard can't work with javascript, and I didn't want to include Flash to do it.  I thought it was intuitive enough to have it select, but I can change the button text.  How about "Select XML" and then maybe a balloon or something to say "Hit Cmd+C or Ctrl+C to Copy"?  Or do we want to add Flash? @Ferdi @pdpinch ?

Ferdi:
no flash, but I think I have seen a js library that does that

carsongee:
I added error reporting on failure

pdpinch:
I'm fine with it as is for now. Please open an issue to deal with it later. 

---

I am mobile.

Ferdi:
I agree with @pdpinch and maybe I was wrong about js library. It could have been flash.

ui/static/ui/js/main.js
```diff
@@ -1,4 +1,4 @@
-define(['jquery', 'bootstrap', 'icheck'], function(jQuery) {
+define(['jquery', 'bootstrap', 'icheck'], function(rjQuery) {
```
noisecapella: Is rjQuery a typo?

carsongee: Good catch, thanks

carsongee:
I snuck one last thing in here (can pull out if it is in the way):
image

to show descriptions if they exist, and default text if they don't on the listing page

noisecapella:
Looks good, will merge after tests pass

PR #246, opened 2015-06-26
Rc/0.2.0

carsongee:


RELEASE.rst
```diff
@@ -1,6 +1,20 @@
...
+- Styled full text search box per the designer.
+- Updated REST API docs.
+- Imported static assets during course import.
+- Added abstract base model to contain audit fields.
+- Added faceted search for taxonomy items.
+- Added Additional APIs for roles.
+- Switched to using get_perms for cleaner code.
+- Added JavaScript infrastructure to run unit tests.
+- Added slide out panel for learning resource details.
+- Styled left hand navigation pane with glyphs.
```
pwilkins: Consider replacing the bulleted list with this:

# Version 0.2.0
- The search bar performs full-text search over the learning resources in 
  the repository, the search results replace the contents of the listing page. 
- Full-text search includes taxonomy facets.
- Learning resources details are displayed in a panel that slides out from 
  the right side of the page.
- Glyphs for learning resources types are displayed in the left side panel for 
  facets.
- LORE's RESTful web service documentation is available. 
  (http://docs.lore.apiary.io)
- Authorizations are in place for taxonomy endpoints in LORE's web service.
- Static assets are imported during course import.
- Roles app has additional features. 

Other Changes

---
- Switched to using get_perms for cleaner code.
- Added JavaScript infrastructure to run unit tests.

carsongee: My concern with this, is that makes the release notes unable to be automated.  The list you see is a copy of the merged feature commit messages.  I think that probably means we want better commit messages going forward, but it will make releases much easier and faster if we don't have to manually rewrite the messages. @pdpinch ?

carsongee:
I have replaced the release notes with your version

pwilkins:
For the record, we've modified the release notes for this release only.  The next release will use the automated process that produces the release notes from commit messages.  Our intent is to improve the quality of the commit messages to make manual editing unnecessary.

PR #251, opened 2015-06-29
Exposed members app features through the RESTful API [#96516348]

giocalitri:
Front end for the REST APIs

### DIFF
--- a/rest/permissions.py
+++ b/rest/permissions.py
@@ -120,13 +120,13 @@ class ManageRepoMembersPermission(BasePermission):
     """Checks manage_repo_users permission"""
 
     def has_permission(self, request, view):
-        if request.method in SAFE_METHODS:
-            return True
         try:
             repo = get_repo(view.kwargs['repo_slug'], request.user.id)
         except NotFound:
-            raise Http404()
+            raise Http404()  # pragma: no cover
         except PermissionDenied:
             return False
+        if request.method in SAFE_METHODS:
+            return True
         return RepoPermission.manage_repo_users[0] in get_perms(
             request.user, repo)
--- a/rest/permissions.py
+++ b/rest/permissions.py
@@ -4,12 +4,12 @@ Permission classes for REST views
 
 from __future__ import unicode_literals
 
+from django.http import Http404
+from guardian.shortcuts import get_perms
 from rest_framework.permissions import (
     BasePermission,
     SAFE_METHODS,
 )
-from django.http import Http404
-from guardian.shortcuts import get_perms
 
 from learningresources.models import Repository
 from learningresources.api import (
@@ -17,11 +17,11 @@ from learningresources.api import (
     PermissionDenied,
     NotFound,
 )
+from roles.permissions import RepoPermission
 from taxonomy.api import (
     get_vocabulary,
     get_term,
 )
-from roles.permissions import RepoPermission
 
 
 # pylint: disable=protected-access
@@ -114,3 +114,19 @@ class ManageTaxonomyPermission(BasePermission):
                 RepoPermission.manage_taxonomy[0]
                 in get_perms(request.user, repo)
             )
+
+
+class ManageRepoMembersPermission(BasePermission):
+    """Checks manage_repo_users permission"""
+
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        try:
+            repo = get_repo(view.kwargs['repo_slug'], request.user.id)
+        except NotFound:
+            raise Http404()
+        except PermissionDenied:
+            return False
+        return RepoPermission.manage_repo_users[0] in get_perms(
+            request.user, repo)
--- a/rest/serializers.py
+++ b/rest/serializers.py
@@ -4,17 +4,23 @@ REST serializers for taxonomy models
 
 from __future__ import unicode_literals
 
+from django.contrib.auth.models import User
+from rest_framework.generics import get_object_or_404
 from rest_framework.serializers import (
+    Serializer,
     ModelSerializer,
     HiddenField,
     CurrentUserDefault,
     CreateOnlyDefault,
+    CharField,
+    ChoiceField,
+    ValidationError,
 )
-from rest_framework.generics import get_object_or_404
 
-from taxonomy.models import Vocabulary, Term
 from learningresources.models import Repository
-from .util import LambdaDefault, RequiredBooleanField
+from rest.util import LambdaDefault, RequiredBooleanField
+from roles.permissions import BaseGroupTypes
+from taxonomy.models import Vocabulary, Term
 
 
 class RepositorySerializer(ModelSerializer):
@@ -97,3 +103,49 @@ class TermSerializer(ModelSerializer):
             'id',
             'slug',
         )
+
+
+class GroupSerializer(Serializer):
+    """
+    Serializer for base_group_type
+    """
+    # pylint: disable=abstract-method
+    group_type = ChoiceField(
+        choices=BaseGroupTypes.all_base_groups()
+    )
+
+    def validate_group_type(self, value):
+        """Validate group_type"""
+        # pylint: disable=no-self-use
+        if value not in BaseGroupTypes.all_base_groups():
+            raise ValidationError(
+                'group "{group}" is not valid'.format(group=value)
+            )
+        return value
+
+
+class UserSerializer(Serializer):
+    """
+    Serializer for user
+    """
+    # pylint: disable=abstract-method
+    username = CharField()
+
+    def validate_username(self, value):
+        """Validate username"""
+        # pylint: disable=no-self-use
+        try:
+            User.objects.get(username=value)
+        except User.DoesNotExist:
+            raise ValidationError(
+                'user "{user}" is not available'.format(user=value)
+            )
+        return value
+
+
+class UserGroupSerializer(UserSerializer, GroupSerializer):
+    """
+    Serializer for username base_group_type association
+    """
+    # pylint: disable=abstract-method
+    pass
--- a/rest/serializers.py
+++ b/rest/serializers.py
@@ -109,7 +109,6 @@ class GroupSerializer(Serializer):
     """
     Serializer for base_group_type
     """
-    # pylint: disable=abstract-method
     group_type = ChoiceField(
         choices=BaseGroupTypes.all_base_groups()
     )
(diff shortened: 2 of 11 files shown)


rest/tests/test_rest.py
```diff
@@ -576,3 +585,566 @@ def test_root(self):
...
+            username="admin_user", password=self.PASSWORD
+        )
+        self.admin_user2 = User.objects.create_user(
+            username="admin_user2", password=self.PASSWORD
+        )
+        # List of all user.
+        self.all_user_repo = [
+            self.author_user,
+            self.curator_user,
+            self.admin_user
```
noisecapella: Is admin_user2 missing intentionally?

giocalitri: yes

rest/tests/test_rest.py
```diff
@@ -576,3 +585,566 @@ def test_root(self):
...
+            admins
+        )
+
+    def test_members_get(self):
+        """
+        Members GET
+        """
+        all_members = sorted(
+            [
+                {u'username': u'admin_user', u'group_type': u'administrators'},
```
noisecapella: We don't need the `u` here since unicode_literals is imported

giocalitri: leftover from copy/paste from shell: will remove

rest/permissions.py
```diff
@@ -114,3 +114,19 @@ def has_permission(self, request, view):
...
                 in get_perms(request.user, repo)
             )
+
+
+class ManageRepoMembersPermission(BasePermission):
+    """Checks manage_repo_users permission"""
+
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
```
ShawnMilo: [Same comment](https://github.com/mitodl/lore/pull/183#discussion_r32950841) as given previously in another PR.

roles/tests/test_user_models.py
```diff
@@ -51,3 +51,11 @@ def __init__(self, foo_var):
 
         user_group2 = FooObject('bar')
         self.assertFalse(user_group1.__eq__(user_group2))
+
+    def test_user_group_hash(self):
+        """Test for UserGroup __hash__"""
+        user_group = UserGroup('foo1', 'bar')
+        self.assertEqual(
+            user_group.__hash__(),
+            hash((user_group.username, user_group.group_type))
```
ShawnMilo: I think it would be nicer if the test created another `UserGroup` instance (even if it wasn't saved to the database) and called `hash()` on both and compared the results. And maybe make a one-character change to `username` and `group_type` and prove `hash()` returns different results.

giocalitri: Don't you think this means I'm going to write unit tests for the `hash` function? 
The only thing I was planning to test was the way I was implementing the `__hash__` not the behavior of the `hash` function I call inside

rest/urls.py
```diff
@@ -39,4 +44,35 @@
         r'vocabularies/(?P<vocab_slug>[-\w]+)/terms/(?P<term_slug>[-\w]+)/$',
         TermDetail.as_view(),
         name='term-detail'),
+    # Section for repository group members
+    url(
+        r'^repositories/(?P<repo_slug>[-\w]+)/'
```
ShawnMilo: All five of these URLs have `r'^repositories/(?P<repo_slug>[-\w]+)/members` in common. Maybe it should be a constant above, making that more obvious and possibly shortening the regexes to fit on one line.

roles/tests/test_user_models.py
ShawnMilo: I don't think this is testing Python's `hash()` function. It tests that when you `hash()` a `UserGroup`, you're definitely hashing on the two fields you think you are.

Even if you decide not to instantiate a second `UserGroup`, which isn't strictly necessary, I definitely think there should be tests for changing the two fields (one at a time) and proving that makes the hashes different.

giocalitri: You still haven't convinced me :-) :
since the `__hash__` function is defined as `hash` of a tuple and since tuples are hashable in python, testing that changing a value in the tuple gives a different hash, to me it means testing the behavior of the `hash` function.

But in any case, it's no big deal: I can add the tests.

rest/permissions.py
```diff
@@ -114,3 +114,19 @@ def has_permission(self, request, view):
...
+        try:
+            repo = get_repo(view.kwargs['repo_slug'], request.user.id)
+        except NotFound:
+            raise Http404()  # pragma: no cover
```
ShawnMilo, line 126: There are three `pragma: no cover`s in this PR. It looks like all three could have coverage.

rest/serializers.py
```diff
@@ -97,3 +103,49 @@ class Meta:
             'id',
             'slug',
         )
+
+
+class GroupSerializer(Serializer):
+    """
+    Serializer for base_group_type
+    """
+    # pylint: disable=abstract-method
```
ShawnMilo: This seems like something we could disable globally. It's not the kind of thing we're going to do by accident. I think it's one of those pylint warnings that's overzealous.

roles/tests/test_user_models.py
ShawnMilo: Yeah, I see your point. The extra tests aren't necessary. I do like seeing apples compared to apples, which is why I would like to see `__hash__` on two separate `UserGroup` instances compared instead of `hash(('string' ,' string'))`, because otherwise it kind of _is_ testing `hash()`.

I think what I was shooting for with the "change string" tests is that if the `__hash__` method is ever changed, that test will break and that would be an indicator that we needed to search the codebase looking for our own uses of `hash()` to make sure we updated them, but that should be obvious during the development and code review.

So, how about just my first comment (instantiating another `UserGroup`), and ignoring the other parts?

rest/serializers.py
```diff
@@ -97,3 +103,49 @@ class Meta:
...
+            )
+        return value
+
+
+class UserGroupSerializer(UserSerializer, GroupSerializer):
+    """
+    Serializer for username base_group_type association
+    """
+    # pylint: disable=abstract-method
+    pass
```
ShawnMilo: Itsy-bitsy nit: You can leave out the `pass` here.

rest/tests/base.py
```diff
@@ -388,3 +388,80 @@ def delete_term(self, repo_slug, vocab_slug, term_slug,
...
+        self.assertEqual(expected_status, resp.status_code)
+        if resp.status_code == HTTP_201_CREATED:
+            result_dict = as_json(resp)
+            if not skip_assert:
+                for key, value in mem_dict.items():
+                    self.assertEqual(value, result_dict[key])
+            if reverse_str is not None:
+                self.assertIn(reverse(
+                    reverse_str, kwargs={
+                        'username': username or mem_dict['username'],
```
ShawnMilo, line 452: Interesting. I was sure this was invalid; I thought it needed to be `username if username is not None else mem_dict['username']`. So, I learned something today. This syntax is nice, but confusing because it _looks_ like it should return a boolean.

rest/tests/test_rest.py
```diff
@@ -1,9 +1,12 @@
 """
 Unit tests for REST api
 """
-
+# pylint: disable=too-many-lines
```
ShawnMilo: Is there a good reason not to split the file up? Do we, as a group, care about `too-many-lines`? It seems like a good thing to adhere to.

roles/tests/test_user_models.py
giocalitri: ok, will do

rest/tests/test_rest.py
```diff
@@ -576,3 +585,566 @@ def test_root(self):
...
+            # get one group for a specific users
+            # author
+            resp_dict = self.get_members(
+                urlfor='users',
+                repo_slug=self.repo.slug,
+                username=self.author_user.username,
+                group_type=BaseGroupTypes.AUTHORS
+            )
+            self.assertEqual(resp_dict, author_member[0])
+            # OPTION request for the following 2 test returns 200 (is it ok?)
```
ShawnMilo: A 200 here seems reasonable. It's just validating the the URL is valid. The 404 returned later is artificial, masking user error to prevent information leakage.

rest/tests/test_rest.py
```diff
@@ -576,3 +585,566 @@ def test_root(self):
...
+        # First part: assign group to user
+        # Assign unexpected group will fail
+        self.create_member(
+            urlfor='users',
+            repo_slug=self.repo.slug,
+            mem_dict={'group_type': 'foo'},
+            username=self.author_user.username,
+            expected_status=HTTP_400_BAD_REQUEST
+        )
+        # Assign real group to inexistent user will fail
```
ShawnMilo: `s/inexistent/nonexistent/` (here and again below)

rest/permissions.py
giocalitri, line 126: the three `pragma: no cover`s are in section of the code that will be never reached because the same checks run in other sections of the code, but I still did not want to remove the code because (like in this case) if we use this permission without others, the check matters.
To properly test this I should write, for example, a test api view with only this check and assert there is 404, but honestly, I'm not going to do it this week. If you want we can open a new Issue for this.

rest/serializers.py
giocalitri: No strong opinions about this, but I guess we want a larger consensus in this case.

rest/serializers.py
giocalitri: Actually, I can only if I remove the pylint disable, otherwise it will not be applied

rest/tests/base.py
giocalitri, line 452: I am pretty used to this syntax and I never think about a boolean. 
I have the opposite problem:

```
x = None
# this is wrong, but I tend to forget that I don't have to do it
if not x:
    ...
else:
    ...
```

Actually that type of syntax is suggested for clarity by Raymond Hettinger in several talks.

rest/tests/test_rest.py
giocalitri: OK, will remove comments

ShawnMilo:
Will merge after tests.

pwilkins:
Consider changing the commit message to: Exposed members app features through the RESTful API.

giocalitri:
do you mean the PR message?

pwilkins:
I meant consider changing "Front end for the REST APIs" to  "Exposed members app features through the RESTful API."  I don't know what you call it. :-(

PR #256, opened 2015-06-29
Switched from using main.js for everything to multiple modules

carsongee:
@noisecapella


noisecapella:
Looks good, waiting for tests

Issue #162, opened 2015-06-15
REST APIs:  LearningResource


Ferdi:
(see #138 )
- LearningResource REST API 
- Documentation

pdpinch:
Should include:
- terms
- related static asset set

carsongee:
is this done @noisecapella or do we need docs still?

noisecapella:
Yes, all done. Docs should be included.

PR #259, opened 2015-06-29
Implemented learning_resources API

noisecapella:
addresses #162
[finishes #96967016]

### DIFF
--- a/rest/permissions.py
+++ b/rest/permissions.py
@@ -11,7 +11,11 @@ from rest_framework.permissions import (
     SAFE_METHODS,
 )
 
-from learningresources.models import Repository
+from learningresources.models import (
+    Repository,
+    LearningResource,
+    StaticAsset,
+)
 from learningresources.api import (
     get_repo,
     PermissionDenied,
@@ -130,3 +134,69 @@ class ManageRepoMembersPermission(BasePermission):
             return True
         return RepoPermission.manage_repo_users[0] in get_perms(
             request.user, repo)
+
+
+class AddEditMetadataPermission(BasePermission):
+    """Checks add_edit_metadata permission"""
+
+    def has_permission(self, request, view):
+        try:
+            repo = get_repo(view.kwargs['repo_slug'], request.user.id)
+        except NotFound:
+            raise Http404()
+        except PermissionDenied:
+            return False
+
+        if request.method in SAFE_METHODS:
+            return True
+
+        return (
+            RepoPermission.add_edit_metadata[0]
+            in get_perms(request.user, repo)
+        )
+
+
+class ViewLearningResourcePermission(ViewRepoPermission):
+    """
+    Checks that user has view_repo permission for the repository
+    and that that repo owns the LearningResource being requested.
+    """
+
+    def has_permission(self, request, view):
+        if not super(ViewLearningResourcePermission, self).has_permission(
+                request, view):
+            return False
+
+        repo_slug = view.kwargs['repo_slug']
+        lr_id = int(view.kwargs['lr_id'])
+
+        try:
+            LearningResource.objects.filter(
+                course__repository__slug=repo_slug).get(id=lr_id)
+        except LearningResource.DoesNotExist:
+            raise Http404()
+
+        return True
+
+
+class ViewStaticAssetPermission(ViewLearningResourcePermission):
+    """
+    Checks that user has view_repo permission for the repository
+    and that that repo owns the LearningResource being requested.
+    """
+
+    def has_permission(self, request, view):
+        if not super(ViewStaticAssetPermission, self).has_permission(
+                request, view):
+            return False
+
+        lr_id = int(view.kwargs['lr_id'])
+        sa_id = int(view.kwargs['sa_id'])
+
+        try:
+            StaticAsset.objects.filter(
+                learningresource__id=lr_id).get(id=sa_id)
+        except StaticAsset.DoesNotExist:
+            raise Http404()
+
+        return True
--- a/rest/serializers.py
+++ b/rest/serializers.py
@@ -15,12 +15,18 @@ from rest_framework.serializers import (
     CharField,
     ChoiceField,
     ValidationError,
+    StringRelatedField,
+    SlugRelatedField,
 )
 
-from learningresources.models import Repository
 from rest.util import LambdaDefault, RequiredBooleanField
 from roles.permissions import BaseGroupTypes
 from taxonomy.models import Vocabulary, Term
+from learningresources.models import (
+    Repository,
+    LearningResource,
+    StaticAsset,
+)
 
 
 class RepositorySerializer(ModelSerializer):
@@ -145,3 +151,62 @@ class UserGroupSerializer(UserSerializer, GroupSerializer):
     """
     Serializer for username base_group_type association
     """
+
+
+class LearningResourceSerializer(ModelSerializer):
+    """Serializer for LearningResource"""
+
+    learning_resource_type = StringRelatedField()
+    terms = SlugRelatedField(
+        many=True, slug_field='slug', queryset=Term.objects.all())
+
+    class Meta:
+        # pylint: disable=missing-docstring
+        model = LearningResource
+        fields = (
+            'id',
+            'learning_resource_type',
+            'static_assets',
+            'title',
+            'description',
+            'content_xml',
+            'materialized_path',
+            'url_path',
+            'parent',
+            'copyright',
+            'xa_nr_views',
+            'xa_nr_attempts',
+            'xa_avg_grade',
+            'xa_histogram_grade',
+            'terms',
+        )
+        read_only_fields = tuple(set(fields) - {'description'})
+
+    def validate_terms(self, terms):
+        """
+        Validate that this LearningResource's learning_resource_type
+        is supported by the vocabulary of each term being added
+        """
+        resource_type = self.instance.learning_resource_type
+        for term in terms:
+            if not term.vocabulary.learning_resource_types.filter(
+                    name=resource_type.name).exists():
+                raise ValidationError(
+                    "Term {} is not supported for this learning resource")
+        return terms
+
+
+class StaticAssetSerializer(ModelSerializer):
+    """Serializer for StaticAsset"""
+
+    class Meta:
+        # pylint: disable=missing-docstring
+        model = StaticAsset
+        fields = (
+            'id',
+            'asset',
+        )
+        read_only_fields = (
+            'id',
+            'asset',
+        )
--- a/rest/tests/base.py
+++ b/rest/tests/base.py
@@ -11,8 +11,11 @@ from rest_framework.status import (
     HTTP_204_NO_CONTENT,
 )
 from rest_framework.reverse import reverse
+from django.db.models import Count
 
+from importer.tasks import import_file
 from learningresources.tests.base import LoreTestCase
+from learningresources.api import get_resources
 
 API_BASE = '/api/v1/'
 REPO_BASE = '/api/v1/repositories/'
@@ -45,6 +48,22 @@ class RESTTestCase(LoreTestCase):
     DEFAULT_TERM_DICT = {
         'label': 'term label',
         'weight': 1000,
+    }
+    DEFAULT_LR_DICT = {
+        "learning_resource_type": -1,
+        "static_assets": [],
+        "title": "Getting Help",
+        "description": "x",
+        "content_xml": "<vertical />",
+        "materialized_path":
+            "/course/chapter[4]/sequential[1]/vertical[3]",
+        "url_path": "",
+        "parent": 93,
+        "copyright": "",
+        "xa_nr_views": 0,
+        "xa_nr_attempts": 0,
+        "xa_avg_grade": 0.0,
+        "xa_histogram_grade": 0.0
     }
 
     def assert_options_head(self, url, expected_status):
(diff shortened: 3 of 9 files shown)


ShawnMilo:
This needs to be updated from master; there are a bunch of merge conflicts. Just mentioning it now before it gets worse.

rest/permissions.py
```diff
@@ -114,3 +118,88 @@ def has_permission(self, request, view):
...
+            return False
+
+        lr_id = int(view.kwargs['lr_id'])
+
+        try:
+            learning_resource = LearningResource.objects.get(id=lr_id)
+        except LearningResource.DoesNotExist:
+            raise Http404()
+
+        if learning_resource.course.repository.slug != repo_slug:
```
ShawnMilo: The check for repo slug can be added to the `.get` query above, and take advantage of the same `DoesNotExist` check.

rest/permissions.py
```diff
@@ -114,3 +118,88 @@ def has_permission(self, request, view):
...
+            raise Http404()
+
+        try:
```
ShawnMilo: Since this duplicates `ViewLearningResourcePermission` exactly up to this point, is it possible to subclass it and just call `super()` within `has_permission`?

rest/serializers.py
```diff
@@ -97,3 +102,65 @@ class Meta:
...
+            'materialized_path',
+            'url_path',
+            'parent',
+            'copyright',
+            'xa_nr_views',
+            'xa_nr_attempts',
+            'xa_avg_grade',
+            'xa_histogram_grade',
+        )
+        read_only_fields = (
```
ShawnMilo: Can `read_only_fields` just be set to `= fields`, since they're the same? It would be easier to keep them in sync.

rest/tests/base.py
```diff
@@ -388,3 +404,109 @@ def delete_term(self, repo_slug, vocab_slug, term_slug,
...
+        self.assert_options_head(url, expected_status=expected_status)
+        resp = self.client.get(url)
+        self.assertEqual(expected_status, resp.status_code)
+        if resp.status_code == HTTP_200_OK:
+            return as_json(resp)
+
+    def patch_learning_resource(
+            self, repo_slug, learning_resource_id, lr_dict,
+            expected_status=HTTP_200_OK, skip_assert=False):
+        """Update a term"""
```
ShawnMilo: `term` is probably a copy/paste typo here.

rest/tests/base.py
```diff
@@ -388,3 +404,109 @@ def delete_term(self, repo_slug, vocab_slug, term_slug,
...
+            result_dict = as_json(resp)
+            if not skip_assert:
+                for key, value in lr_dict.items():
+                    self.assertEqual(value, result_dict[key])
+            return result_dict
+
+    def put_learning_resource(
+            self, repo_slug, learning_resource_id, lr_dict,
+            expected_status=HTTP_200_OK, skip_assert=False):
+        """Update a term"""
```
ShawnMilo: Same here with `term`.

rest/tests/base.py
```diff
@@ -388,3 +404,109 @@ def delete_term(self, repo_slug, vocab_slug, term_slug,
...
+        self.assert_options_head(url, expected_status=expected_status)
+        resp = self.client.get(url)
+        self.assertEqual(expected_status, resp.status_code)
+        if resp.status_code == HTTP_200_OK:
+            return as_json(resp)
+
+    def get_static_asset(self, repo_slug, learning_resource_id,
+                         static_asset_id,
+                         expected_status=HTTP_200_OK):
+        """Get a learning resource"""
```
ShawnMilo: Should be `static asset`.

rest/tests/test_rest.py
```diff
@@ -524,9 +532,34 @@ def test_immutable_fields(self):
...
+        # these keys should be different since they are immutable or set by
+        # the serializer
+        self.assertNotEqual(vocab_dict['id'], result['id'])
+        self.assertNotEqual(vocab_dict['slug'], result['slug'])
+        self.assertNotIn('repository', result)
+        vocabulary = Vocabulary.objects.get(slug=result['slug'])
+        self.assertEqual(self.repo.id, vocabulary.repository.id)
+        vocab_slug = result['slug']
 
+        result = self.put_vocabulary(
```
ShawnMilo, line 651: This set of asserts is in this test three times. I could be a good use for a nested function.

rest/tests/test_rest.py
```diff
@@ -535,8 +568,32 @@ def test_immutable_fields(self):
             "vocabulary": -1,
         }
         result = self.create_term(
-            repo_dict['slug'], vocab_dict['slug'], term_dict,
-            skip_assert=True
+            self.repo.slug, vocab_slug, term_dict, skip_assert=True
+        )
+        # these keys should be different since they are immutable or set by
+        # the serializer
+        self.assertNotEqual(term_dict['id'], result['id'])
```
ShawnMilo: All the testing of the immutable fields seems like it's testing the restful app package, not our code. It's verbose and repetitive and isn't required for full test coverage.

rest/serializers.py
noisecapella: They're not exactly the same, 'description' can be mutated

rest/tests/test_rest.py
```diff
@@ -544,7 +601,131 @@ def test_immutable_fields(self):
...
+        self.assertEqual(vocab_slug, term.vocabulary.slug)
+
+    def test_immutable_fields_learning_resource(self):
+        """Test immutable fields for term"""
+        tarball_file = self.get_course_single_tarball()
+        import_file(
+            tarball_file, self.repo.id, self.user.id)
+        resource = get_resources(self.repo.id).first()
+        resource.static_assets.add(StaticAsset.objects.first())
+        lr_id = self.get_learning_resources(self.repo.slug)['results'][0]['id']
```
ShawnMilo: Is this line meant to get the primary key of `resource` on the line above?

rest/tests/test_rest.py
```diff
@@ -544,7 +601,131 @@ def test_immutable_fields(self):
...
+            "xa_nr_attempts": 2,
+            "xa_avg_grade": 3.0,
+            "xa_histogram_grade": 4.0
+        }
+
+        # these keys should be different since they are immutable or set by
+        # the serializer
+        result = self.patch_learning_resource(self.repo.slug, lr_id, lr_dict,
+                                              skip_assert=True)
+        self.assertNotEqual(lr_dict['id'], result['id'])
```
ShawnMilo: All these `assertNotEqual` lines could be done in a loop, since the dict keys are all the same. Although as mentioned above, all immutable field tests should be unnecessary.

rest/tests/test_rest.py
```diff
@@ -544,7 +601,131 @@ def test_immutable_fields(self):
...
+        repo_slug1 = self.repo.slug
+        repo_id1 = self.repo.id
+        import_file(
+            tarball_file, repo_id1, self.user.id)
+        resource1 = get_resources(repo_id1).first()
+        as1 = StaticAsset.objects.all()[0]
+        resource1.static_assets.add(as1)
+        lr1_id = resource1.id
+
+        tarball_file = self.get_course_single_tarball()
```
ShawnMilo: Duplicate line from above.

rest/serializers.py
ShawnMilo: Okay, but the comment is still valid and you can still get the same benefits. Make the one tuple, then copy it and add the one value for the other.

rest/tests/test_rest.py
```diff
@@ -544,7 +581,143 @@ def test_immutable_fields(self):
...
+        new_repo_dict = self.create_repository()
+        repo_slug2 = new_repo_dict['slug']
+        repo_id2 = new_repo_dict['id']
+        import_file(
+            tarball_file, repo_id2, self.user.id)
+        resource2 = get_resources(repo_id2).first()
+        # Warning: this points to the second static asset which was
```
ShawnMilo: This should be fixed in this PR.

rest/tests/test_rest.py
noisecapella: It's important in making sure we don't miss a field and let the user edit fields they shouldn't be allowed to

rest/tests/test_rest_authorization.py
```diff
@@ -642,3 +648,234 @@ def test_members_delete(self):
...
+        """Test retrieve learning resource"""
+        self.assertEqual(
+            1, self.get_learning_resources(self.repo.slug)['count'])
+
+        tarball_file = self.get_course_single_tarball()
+        import_file(
+            tarball_file, self.repo.id, self.user.id)
+        resource = get_resources(self.repo.id).first()
+        resource.static_assets.add(StaticAsset.objects.first())
+        lr_id = self.get_learning_resources(self.repo.slug)['results'][0]['id']
```
ShawnMilo: It looks like this is the ID of `resource` in the line above. If not, it's unclear what this is doing.

rest/tests/test_rest_authorization.py
```diff
@@ -642,3 +648,234 @@ def test_members_delete(self):
...
+            self.repo.slug, lr_id, expected_status=HTTP_403_FORBIDDEN)
+
+        # as anonymous
+        self.logout()
+        self.get_learning_resources(
+            self.repo.slug, expected_status=HTTP_403_FORBIDDEN)
+        self.get_learning_resource(
+            self.repo.slug, lr_id, expected_status=HTTP_403_FORBIDDEN)
+
+    def test_resource_create(self):
```
ShawnMilo: This test should probably be named `test_resource_post`, since it's not testing creation.

rest/tests/test_rest.py
noisecapella: `tarball_file` is deleted in `import_file` so we need to call it again

rest/tests/test_rest_authorization.py
```diff
@@ -642,3 +648,234 @@ def test_members_delete(self):
...
+                repo_slug=self.repo.slug,
+                lr_id=lr_id,
+            ),
+            self.DEFAULT_LR_DICT
+        )
+        self.assertEqual(resp.status_code, HTTP_405_METHOD_NOT_ALLOWED)
+
+    def test_resource_put_patch(self):
+        """Test update learning resource"""
+        tarball_file = self.get_course_single_tarball()
```
ShawnMilo: These lines getting the tarball, importing, and getting a resource are repeated many times. They should probably be a function. Perhaps one that returns a resource or sets `self.resource`.

rest/tests/test_rest_authorization.py
```diff
@@ -642,3 +648,234 @@ def test_members_delete(self):
...
+
+        new_description_dict = {"description": "new description"}
+
+        # as curator who has add_edit_metadata permission
+        self.logout()
+        self.login(self.curator_user.username)
+        self.patch_learning_resource(
+            self.repo.slug, lr_id, new_description_dict
+        )
+        self.put_learning_resource(
```
ShawnMilo, line 721: The result of these `patch` and `put` calls is never checked; apparently only the response code being 200. If they were checked, the `put` may or may not work, since it's duplicating the description from the `patch`.

rest/tests/test_rest.py
noisecapella: I'll change this in anticipation of #270 getting merged first

rest/tests/test_rest_authorization.py
```diff
@@ -642,3 +648,234 @@ def test_members_delete(self):
...
+            self.repo.slug, lr_id, new_description_dict
+        )
+
+        # author does have add_edit_metadata permission
+        self.logout()
+        self.login(self.author_user.username)
+        self.patch_learning_resource(
+            self.repo.slug, lr_id, new_description_dict
+        )
+        self.put_learning_resource(
```
ShawnMilo, line 733: No check on what it does (or doesn't do) here either.

noisecapella, line 733: The functions assert that for every key present in the input dictionary, the output value matches up with the input value. I'll make sure the input data is updated before each call to properly test a change in state

rest/permissions.py
```diff
@@ -130,3 +134,69 @@ def has_permission(self, request, view):
             return True
         return RepoPermission.manage_repo_users[0] in get_perms(
             request.user, repo)
+
+
+class AddEditMetadataPermission(BasePermission):
+    """Checks add_edit_metadata permission"""
+
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
```
ShawnMilo: Same comment about this as [here](https://github.com/mitodl/lore/pull/183#discussion_r32950841)

rest/permissions.py
```diff
@@ -130,3 +134,69 @@ def has_permission(self, request, view):
...
+        return (
+            RepoPermission.add_edit_metadata[0]
+            in get_perms(request.user, repo)
+        )
+
+
+class ViewLearningResourcePermission(ViewRepoPermission):
+    """
+    Checks that user has view_repo permission for the repository
+    and that that repo owns the learning resource being requested
```
ShawnMilo: nit: Missing period at the end of the sentence.

rest/permissions.py
```diff
@@ -130,3 +134,69 @@ def has_permission(self, request, view):
...
+        except LearningResource.DoesNotExist:
+            raise Http404()
+
+        return True
+
+
+class ViewStaticAssetPermission(ViewLearningResourcePermission):
+    """
+    Checks that user has view_repo permission for the repository
+    and that that repo owns the learning resource being requested
```
ShawnMilo: nits:

"learning resource" here, but "LearningResource" in lots of other comments.

No period at the end.

rest/tests/test_rest.py
```diff
@@ -553,16 +558,23 @@ def test_immutable_fields(self):
...
-        self.assertNotEqual(repo_dict['slug'], result['slug'])
-        self.assertNotEqual(repo_dict['date_created'], result['date_created'])
-        self.assertNotIn('created_by', result)
-        repository = Repository.objects.get(slug=result['slug'])
-        self.assertEqual(repository.created_by.id, self.user.id)
-        repo_dict = result
+
+        def assert_not_changed(new_dict):
+            """Check that fields have not changed"""
+            # these keys should be different since they are immutable or set by
```
ShawnMilo: Nit: capitalization and punctuation missing. I'll stop commenting on writing stuff after this.

rest/tests/test_rest.py
```diff
@@ -592,17 +622,115 @@ def test_immutable_fields(self):
...
+            "parent": 22,
+            "copyright": "copyright",
+            "xa_nr_views": 1,
+            "xa_nr_attempts": 2,
+            "xa_avg_grade": 3.0,
+            "xa_histogram_grade": 4.0,
+            "terms": []
+        }
+
+        def assert_not_changed(new_dict):
```
ShawnMilo, line 721: These four `assert_not_changed` functions would be easier to read and maintain if written like this. That's what I meant in a previous comment, but I probably wasn't clear enough.

``` python
def assert_not_changed(new_dict):
    """Make sure immutable fields haven't been changed."""

    fields = (
        "content_xml", "id", "learning_resource_type", "materialized_path",
        "parent", "static_assets", "title", "url_path", "xa_avg_grade",
        "xa_histogram_grade", "xa_nr_attempts", "xa_nr_views",
    )
    for field in fields:
        self.assertNotEqual(lr_dict[field], new_dict[field])
    self.assertNotEqual(lr_dict['copyright'], new_dict['id'])
```

rest/urls.py
```diff
@@ -76,4 +80,19 @@
         RepoMemberUserGroupDetail.as_view(),
         name='repo-members-user-group-detail'
     ),
+    url(r'^repositories/(?P<repo_slug>[-\w]+)/learning_resources/$',
```
ShawnMilo: Suggestion:

There are four URLs which have `r'^repositories/(?P<repo_slug>[-\w]+)/learning_resources/'` in common. Maybe that could be a variable, for clarity, like [this](https://github.com/mitodl/lore/blob/master/rest/urls.py#L23).

rest/tests/test_rest.py
noisecapella, line 721: That last line was a typo but probably drives home the need to make that change

ShawnMilo, line 721: The last line was intentional; all of the dict keys matched except that one. Unless you're saying the original dict comparing `copyright` to `id` was a typo. If so, cool, because that simplifies it.

noisecapella, line 721: Yeah, that was my typo

noisecapella:
This is waiting on #270 before tests will pass. I'll repush after that's in master

noisecapella:
Changing to waiting on author until I update the docs

noisecapella:
@ShawnMilo In addition to changes in docs I also changed a setting in `lore/settings.py` to output `FileField` without the URL prefix, and a test at the end of `test_rest.py` for that. Other than that I just addressed your comments

ShawnMilo:
Okay. I'll review again now.

ShawnMilo:
Lots of comments are all-lower-case with no punctuation. That may be a nit; I'll let others weigh in. Other than that, I think it can merge after tests pass.

PR #260, opened 2015-06-30
Removed duplicate ManyToMany field

noisecapella:
Since both `LearningResource` and `StaticAsset` have a `ManyToMany` field there are two database tables covering the same relation.


ShawnMilo:
Looks good to me. Since we're not populating it yet, we don't need to worry about a data migration, so this looks complete.

carsongee:
:+1: will merge when tests finish

PR #261, opened 2015-06-30
Renamed 'Copy to Clipboard' to 'Select Text'

noisecapella:
Since copy to clipboard functionality isn't on the roadmap we should rename this button to better set user expectations

### DIFF
--- a/ui/templates/includes/resource_panel.html
+++ b/ui/templates/includes/resource_panel.html
@@ -18,7 +18,7 @@
             <textarea class="form-control textarea-xml">
             </textarea>
             <p class="text-right">
-              <a id="copy-textarea-xml" href="#" class="btn btn-white">Copy to Clipboard</a>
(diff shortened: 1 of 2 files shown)


ui/templates/includes/resource_panel.html
```diff
@@ -18,7 +18,7 @@
             <textarea class="form-control textarea-xml">
             </textarea>
             <p class="text-right">
-              <a id="copy-textarea-xml" href="#" class="btn btn-white">Copy to Clipboard</a>
+              <a id="copy-textarea-xml" href="#" class="btn btn-white">Select Text</a>
```
carsongee: I think "Select XML" would be clearer

noisecapella: I'll change it

carsongee:
@pdpinch and @Ferdi?

pdpinch:
Sure.

PR #262, opened 2015-06-30
Pylint: global disable for abstract-method

giocalitri:


ShawnMilo:
"Reviewed." I'm going to throw caution to the wind and merge it before tests pass.

PR #263, opened 2015-06-30
Added ability to filter vocabularies by learning resource type

noisecapella:
This adds the ability to filter vocabularies by learning resource types, for example "/api/v1/repositories/repo_slug/vocabularies/?learning_resource_type=chapter"

### DIFF
--- a/rest/views.py
+++ b/rest/views.py
@@ -42,6 +42,7 @@ from rest.permissions import (
 )
 from rest.util import CheckValidMemberParamMixin
 from learningresources.models import Repository
+from taxonomy.models import Vocabulary
 from learningresources.api import (
     get_repos,
 )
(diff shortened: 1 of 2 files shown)


rest/views.py
```diff
@@ -89,8 +90,15 @@ class VocabularyList(ListCreateAPIView):
 
     def get_queryset(self):
         """Filter to vocabularies for a repository"""
```
pwilkins: Out of scope question: What does this description mean? "Filter to vocabularies..."?

noisecapella: It means to return a list of vocabularies filtered by ownership by the repository

carsongee: This should get updated now that it is handling querystring filters as well.

rest/views.py
```diff
@@ -89,8 +90,15 @@ class VocabularyList(ListCreateAPIView):
...
-            slug=self.kwargs['repo_slug']).vocabulary_set.order_by('id')
+        queryset = Vocabulary.objects.filter(
+            repository__slug=self.kwargs['repo_slug']
+        )
+        learning_resource_type = self.request.query_params.get(
+            'learning_resource_type', None)
+        if learning_resource_type is not None:
+            queryset = queryset.filter(
+                learning_resource_types__name=learning_resource_type)
+        return queryset.order_by('id')
```
carsongee, line 118: Why are we ordering this at all?

rest/views.py
```diff
@@ -89,8 +90,15 @@ class VocabularyList(ListCreateAPIView):
...
-        return Repository.objects.get(
-            slug=self.kwargs['repo_slug']).vocabulary_set.order_by('id')
+        queryset = Vocabulary.objects.filter(
+            repository__slug=self.kwargs['repo_slug']
+        )
+        learning_resource_type = self.request.query_params.get(
+            'learning_resource_type', None)
+        if learning_resource_type is not None:
+            queryset = queryset.filter(
+                learning_resource_types__name=learning_resource_type)
```
carsongee, line 117: Wouldn't this be better by slug?  We don't have any types with weird names because they have to be valid XML, but semantically it would make more sense to do this by slug instead of name, or by type id.  Also, this kind of makes the querystring a bit ambiguous since it doesn't really say in the name what it is expecting.  Maybe `type_slug` or `resource_slug` or something.

rest/views.py
noisecapella, line 118: If we don't order it results may be in a random order. It may not matter though, I was erring on the side of caution here

Issue #223, opened 2015-06-23
As a user I would like to be redirected to the login page if I haven't logged in


carsongee:
Currently when you go to a protected listing page without logging in, it raises an error but doesn't redirect you to the login page.

amir-qayyum-khan:
@carsongee by protected listing page you mean open a repo? Because I found issue when I open a repo direct without login it throws exception instead of redirect

pdpinch:
That's exactly what we want to change. 

---

I am mobile.

amir-qayyum-khan:
fix @pdpinch 

Thanks

PR #267, opened 2015-06-30
fix redirect issue on list of repository.html page

amir-qayyum-khan:
Hi

In this PR I fixed issue when user is not login and try to access repository.html then currently app is showing exception instead of redirecting to login page.

I fix it by redirecting to login page if user is not login

https://github.com/mitodl/lore/issues/223
cc @pdpinch @ShawnMilo @carsongee

### DIFF
--- a/ui/tests/test_learningresources_views.py
+++ b/ui/tests/test_learningresources_views.py
@@ -188,6 +188,17 @@ class TestViews(LoreTestCase):
         self.assertTrue("This field is required." in body)
         self.assertTrue(Repository.objects.count() == 1)
 
+    def test_access_without_login(self):
+        """
+        Tests the repository page without login
+        """
+        self.logout()
+        response = self.client.get(self.repository_url, follow=True)
+        self.assertEqual(response.status_code, 200)
+        # we were redirected to login
+        self.assertEqual(len(response.redirect_chain), 2)
+        self.assertTrue(302 in response.redirect_chain[0])
+
     def test_export_good(self):
         """Get raw XML of something I should be allowed to see."""
         url = reverse("export", args=(self.repo.slug, self.resource.id))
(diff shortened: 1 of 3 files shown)


carsongee:
This fixes only a very narrow use case.  I was hoping we could do this more widespread, i.e. have it work for all views, since the XML detail view for example, has the same behavior.  There are probably just a couple more spots than this

carsongee:
This will also need a test to validate the new behavior

carsongee:
You should really just add the `@login_required` decorator to these views as it will handle all of that.  We don't want to redirect them to login if they already are logged in but don't have permissions as that would cause a redirect loop.

amir-qayyum-khan:
@carsongee if I use `@login_required on __call__` it gives me exception https://gist.github.com/amir-qayyum-khan/57c407522ffaf6b905fb

If I try to add `@login_required on top of class RepositoryView(FacetedSearchView):` i get exception https://gist.github.com/amir-qayyum-khan/8377ed82602b6cdda0fc

@carsongee what If I check if user is in logged in and instead of redirecting him to login page i can redirect him to listing page in case where user has no permission. To avoid any redirect loop.

cc @pdpinch

carsongee:
Have you tried: http://stackoverflow.com/questions/14694754/haystack-login-required ?

amir-qayyum-khan:
Done @carsongee

amir-qayyum-khan:
test this please

amir-qayyum-khan:
please test this

carsongee:
Can you add this to all of these views?  For the other ones, you should be able to just use the regular login_required decorator

amir-qayyum-khan:
@carsongee as we discuss login_required is present all over the ui/view and i just remove one use less decorator from `RepositoryView.dispatch`, and test

PR #270, opened 2015-06-30
Feature/skm/parse static assets

### LINKED ISSUE #230: static assets: parse video blocks for links to transcript files

ShawnMilo:
closes #230.


importer/api/__init__.py
```diff
@@ -110,6 +106,7 @@ def import_course(bundle, repo_id, user_id):
         bundle (xbundle.XBundle): Course as xbundle XML
         repo_id (int): Primary key of repository course belongs to
         user_id (int): Primary key of Django user doing the import
+        static_dir: location of static files
```
noisecapella: Add a type to this line

importer/api/__init__.py
```diff
@@ -144,6 +142,16 @@ def import_children(course, element, parent):
...
     )
+    if element.tag == "video":
+        subname = get_video_sub(element)
+        if subname != "":
+            assets = StaticAsset.objects.filter(
+                course__id=resource.course_id,
+                path=subname,
+            )
+            for asset in assets:
+                resource.static_assets.add(asset)
```
noisecapella, line 153: Does Django allow a bulk insert in this case? Not important for here, just curious

ShawnMilo, line 153: Nope, tried that already. Error. At least with `.add()`, and I don't see another method available in [the docs](https://docs.djangoproject.com/en/1.8/topics/db/examples/many_to_many/).

ShawnMilo, line 153: I'm looking at it on the command line, and it still doesn't look like it. I can `bulk_create`, but that actually creates. There's an `in_bulk` as well, but that retrieves. None of the other methods appear to be "`bulk_add`."

learningresources/api.py
```diff
@@ -222,30 +224,69 @@ def get_resource(resource_id, user_id):
...
+            to_load.name = "subname"
+            StaticAsset.objects.create(
+                course=course, asset=to_load, path=subname)
+
+
+def create_static_asset(course, handle):
+    """
+    Create a static asset.
+    Args:
+        course id learningresources.models.Course pk: int
```
noisecapella: The type should be put in parentheses and any description or further information should go after the `:`

learningresources/api.py
```diff
@@ -222,30 +224,69 @@ def get_resource(resource_id, user_id):
...
+    else:
+        return u'{0}_subs_{1}.srt.sjson'.format(lang, subs_id)
+
+
+def get_video_sub(xml):
+    """
+    Get subtitle IDs from <video> XML.
+
+    Args:
+        xml lxml.etree: xml for a LearningResource
```
noisecapella: Parentheses for args

learningresources/models.py
```diff
@@ -61,6 +61,7 @@ class StaticAsset(BaseModel):
     Holds static assets for a course (css, html, javascript, images, etc)
     """
     course = models.ForeignKey(Course)
+    path = models.CharField(max_length=256)  # path in static folder
```
noisecapella: We should probably increase the max_length to 64k since Linux allows up to 4k for the most part and 64k would use all of the 2 bytes

learningresources/tests/test_utils.py
```diff
@@ -0,0 +1,45 @@
+"""
+Tests for utilities within the API>
```
noisecapella: Typo at the end of line

learningresources/tests/test_utils.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+
+class TestUtils(TestCase):
+    """
+    Test manipulation utilities.
+    """
+
+    def test_sub_language(self):
+        """Test foo."""
```
noisecapella: Description looks like it needs updating

test_requirements.txt
```diff
@@ -9,6 +9,7 @@ pytest-capturelog
 pytest-watch
 pytest-xdist
 mock
+ipdb
```
noisecapella, line 12: Do we use this anywhere?

learningresources/api.py
pwilkins: As @noisecapella proposed, the line should be:
`course (int): learningresources.models.Course primary key`

noisecapella:
Some minor comments but looks good overall. After your changes I'll run through it on my computer, make sure tests pass and merge

test_requirements.txt
ShawnMilo, line 12: No, this can't be used in production. It's optionally used during development in place of pdb, with the advantage of tab-completion.

pdpinch:
@pwilkins and/or @carsongee should look at this as well before its merged

pwilkins:
I thought that we were associating the static assets with the learningresources that they appear in. My reading of the code is that static assets are parsed out of the xbundle and associated with the course before the course's learningresources even exist.  

Is PR #260 related to this?

noisecapella:
@pwilkins `StaticAsset`s are added to `LearningResource` at https://github.com/mitodl/lore/pull/270/files#diff-ed31ac373341fa1f3b41d869954a2022R153

noisecapella:
I looked into the database after inserting the course. I'm good with merging :+1:

test_requirements.txt
pwilkins, line 12: Ouch.  That probably snuck in when Shawn and I were pair programming.  I'll take responsibility for that though I never got it to work in docker on the mac.

noisecapella:
Tests are failing with a weird error message. It looks like something to do with a change to `pytest.ini`

ShawnMilo:
That pylintrc is a local change I committed by mistake. I re-pushed.

test_requirements.txt
ShawnMilo, line 12: I left it in because I thought a solution was found (that service port command-line flag). It doesn't hurt anything, and if anyone is going to use it, it's nice not to have to rebuild your Docker containers to do so.

Having said that, we can remove it if nobody thinks they're going to use it.

noisecapella, line 12: I'm happy to keep it in

importer/api/__init__.py
```diff
@@ -144,6 +142,16 @@ def import_children(course, element, parent):
         content_xml=etree.tostring(element),
         mpath=mpath,
     )
+    if element.tag == "video":
```
carsongee, line 145: Since we expect to get quite a few of these parsers on different elements, maybe move this out to a function.  It might be cool to just have a dictionary of functions to do parsing based on the tag

importer/api/__init__.py
```diff
@@ -144,6 +142,16 @@ def import_children(course, element, parent):
...
+        subname = get_video_sub(element)
+        if subname != "":
+            assets = StaticAsset.objects.filter(
+                course__id=resource.course_id,
+                path=subname,
+            )
+            for asset in assets:
```
carsongee, line 152: why would there be multiple assets with one filename and course?  That doesn't seem possible

learningresources/models.py
```diff
@@ -61,6 +61,7 @@ class StaticAsset(BaseModel):
     Holds static assets for a course (css, html, javascript, images, etc)
     """
     course = models.ForeignKey(Course)
+    path = models.TextField()  # path in static folder
```
carsongee: I don't think you should need this, this is just the asset.name minus the basepath so you can just replace to get the path.

importer/tests/test_import.py
```diff
@@ -199,28 +203,46 @@ def test_import_static_recurse(self):
...
-        self.assertEqual(course.count(), 1)
-        assets = StaticAsset.objects.filter(course=course)
-        for asset in assets:
-            self.addCleanup(default_storage.delete, asset.asset)
-        self.assertEqual(assets.count(), 2)
-        for asset in assets:
-            base_path = static_asset_basepath(asset, '')
-            self.assertIn(
-                asset.asset.name.replace(base_path, ''),
-                ['test.txt', 'subdir/subtext.txt']
```
carsongee, line 225: You are removing specific tests here by not checking for specific and expected file names unless I'm missing something

carsongee, line 225: Also, you aren't cleaning up assets anymore unless you moved that somewhere else.  Leaving behind artifacts will mess with tests.  That is one of the reasons for the asserting of the expected file names.

importer/api/__init__.py
ShawnMilo, line 145: Yep. Will do that when there are more. The original issue is for all types; this is just the "proof of concept" branch for this.

learningresources/api.py
```diff
@@ -222,30 +224,69 @@ def get_resource(resource_id, user_id):
...
+    # Handle videos first.
+    videos = xml.findall(".//video")
+    for video in videos:
+        subname = get_video_sub(video)
+        if subname == "":
+            continue
+        with open(os.path.join(static_dir, subname), "rb") as raw:
+            to_load = File(raw)
+            to_load.name = "subname"
+            StaticAsset.objects.create(
```
carsongee: This code looks like you aren't storing assets unless they are a type you understand.  We really need to import all assets regardless and just do links when we find them so we can provide an interface for them to provide corrections/find missing files

importer/api/__init__.py
ShawnMilo, line 152: You're right. I think I had the `.filter()` there before I added the `path` field. Will fix.

learningresources/tests/test_api.py
```diff
@@ -157,36 +152,6 @@ def test_invalid_create_repo(self):
         self.assertIn("description", ex.exception.args[0])
 
 
-class TestStaticAsset(LoreTestCase):
```
carsongee, line 160: why did you remove this?

learningresources/models.py
ShawnMilo: Okay. I'll check it out. Incidentally, asset.name is an actual site; Gmail highlighted it and I clicked it without thinking.

test_requirements.txt
carsongee, line 12: ipdb good, maybe add pudb while we are it for the more graphically oriented

learningresources/models.py
carsongee: feel free to add a helper function to the model for that.

ShawnMilo: It added it because I didn't know you could query the `.name` in a filter, since it's not the fieldname. From a StackOverflow posting, it appears that the ORM uses the name as the filter.

learningresources/tests/test_api.py
ShawnMilo, line 160: It was causing an error in `tearDown` that I spent too much time trying to figure out (no errors in the test itself), and it wasn't needed for coverage due to the new tests I added.

learningresources/api.py
ShawnMilo: My understanding is that we'd eventually add all the types as part of this effort. We can add everything up front, but it seemed like there were likely to be unused detritus in courses.

However, the interface you mentioned, allowing them to "shop" for arbitrary assets, makes sense. I'll import everything.

learningresources/tests/test_api.py
```diff
@@ -157,36 +152,6 @@ def test_invalid_create_repo(self):
...
-            )
-
-            # Close file and assert that exception is raised.
-            test_file.close()
-            test_file.name = 'new_name'
-            with self.assertRaises(ValueError):
-                asset = create_static_asset(self.course.id, test_file)
-        self.assertEqual(file_contents, asset.asset.read())
-
-
```
carsongee, line 189: There aren't any unit tests for `create_static_assets` that I can find, just the big integration test.  We should have both

learningresources/api.py
carsongee: Yep, we will never be able to fully parse static assets, so we import them all, that is why I wrote this the way it is

learningresources/tests/test_api.py
carsongee, line 160: It is the only unit level test for that method, and it isn't causing errors on master...It is generally never a good idea to remove a test, and multiple coverage is important, generally you should have tests for each individual function, and then a few integration tests, not just integration tests just for coverage.

ShawnMilo, line 160: Okay. I'll restore it.

importer/tests/test_import.py
ShawnMilo, line 225: The cleanup is happening in the `LoreTestCase.tearDown()`. It seems like a good place to cover all static assets instead of calling `addCleanup()` in individual tests. Any objections?

learningresources/api.py
```diff
@@ -222,30 +225,71 @@ def get_resource(resource_id, user_id):
...
+        sub string: subtitle string
+    """
+    subs = xml.xpath("@sub")
+    # It's not possible to have more than one.
+    if len(subs) == 0:
+        return ""
+    return _subs_filename(subs[0])
+
+
+def import_static_assets(course, path):
```
carsongee, line 277: fine to leave as is, but I'm curious why you moved this out of import

ShawnMilo, line 277: I think it was because it's directly creating model instances within the `learningresources` app, so it seemed to fit better there.

ui/templates/includes/resource_panel.html
```diff
@@ -11,6 +11,11 @@
             <span class="hidden-xs">XML</span>
           </a>
         </li>
+        <li>
```
carsongee: I'm thinking you can leave this in, but leave it commented out.  Also, the tab looks funny due to the words stacking on top of each other, so we will probably want to change it to just `Assets` or `Static`, or `Files` maybe.  Either that or tweak the CSS so `Static Assets` shows on one line.

carsongee:
Other than the small html note, and a squash :+1: to merge

ui/templates/includes/resource_panel.html
ShawnMilo: It's branch pollution; it's in its own branch and I'll delete it from here and squash.

ShawnMilo:
Removed the rogue HTML and rebased.

carsongee:
Maybe commit message should be something like "Added linking of transcripts to learning resources", @pwilkins ?

pwilkins:
Or, "Linked video transcripts to learning resources"?  I'm fine with either.

ShawnMilo:
Updated commit message, fixed rebase to reflect Peter's commit.

PR #271, opened 2015-07-01
Added UI to add and remove repository members. [#96844288]

giocalitri:
UI to add and remove repository members.
Does not include tests.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,4 +1,5 @@
 {
   "preset": "google",
-  "validateQuoteMarks": null
+  "validateQuoteMarks": null,
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties"
 }
--- a/ui/static/ui/js/listing.js
+++ b/ui/static/ui/js/listing.js
@@ -1,9 +1,80 @@
 define(
   ['jquery', 'setup_manage_taxonomies', 'bootstrap', 'icheck', 'csrf'],
-  function(jQuery, setupManageTaxonomies) {
+  function($, setupManageTaxonomies) {
     'use strict';
 
-    jQuery(document).ready(function($) {
+    var EMAIL_EXTENSION = '@mit.edu';
+    function formatGroupName(string) {
+      string = string.charAt(0).toUpperCase() + string.slice(1);
+      return string.substring(0, string.length - 1);
+    }
+
+    /* This is going to grow up to check whether
+     * the name is a valid Kerberos account
+     */
+    function isEmail(username) {
+      var email = username + EMAIL_EXTENSION;
+      var regex = /^[-a-z0-9~!$%^&*_=+}{\'?]+(\.[-a-z0-9~!$%^&*_=+}{\'?]+)*@([a-z0-9_][-a-z0-9_]*(\.[-a-z0-9_]+)*\.(aero|arpa|biz|com|coop|edu|gov|info|int|mil|museum|name|net|org|pro|travel|mobi|[a-z][a-z])|([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}))(:[0-9]{1,5})?$/i;
+      return regex.test(email);
+    }
+
+    function formatUserGroups(userList, dest) {
+      var $dest = $(dest);
+      for (var i = 0; i < userList.length; i++) {
+        $dest.append(
+          '<div class="row">\n' +
+            '<div class="col-sm-7">\n' +
+              '<div class="cd-panel-members-list-username">' +
+                userList[i].username + EMAIL_EXTENSION + '</div>\n' +
+            '</div>\n' +
+            '<div class="col-sm-3">\n' +
+              '<div class="cd-panel-members-list-group_type">' +
+                formatGroupName(userList[i].group_type) + '</div>\n' +
+            '</div>\n' +
+            '<div class="col-sm-2">\n' +
+              '<button ' +
+                 'class="btn btn-default cd-panel-members-remove" ' +
+                 'data-username="' + userList[i].username + '" ' +
+                 'data-group_type="' + userList[i].group_type + '">\n' +
+                 '<i class="fa fa-minus"></i>\n' +
+              '</button>\n' +
+            '</div>\n' +
+          '</div>\n');
+      }
+    }
+
+    function resetUserGroupForm() {
+      $('input[name=\'members-username\']').val('');
+      $('select[name=\'members-group_type\']').prop('selectedIndex', 0);
+    }
+
+    function showMembersAlert(message, mtype) {
+      //default value for mtype
+      mtype = typeof mtype !== 'undefined' ? mtype : 'success';
+      //reset all the classes
+      $('#members-alert').html(
+        '<div class="alert alert-' + mtype +
+            ' fade in out" data-alert="alert">\n' +
+          '<a href="#" class="close" data-dismiss="alert" ' +
+            'aria-label="close">&times;</a>\n' + message + '\n' +
+        '</div>');
+    }
+
+    function showUpdateAllMembers() {
+      //retrieve all the members
+      var url = $('.cd-panel-members').data('members-url');
+      $('#cd-panel-members-list').empty();
+      $.ajax({url: url})
+        .done(function(data) {
+          formatUserGroups(data.results, '#cd-panel-members-list');
+        })
+        .error(function() {
+          var message = 'Unable to retrieve list of members.';
+          showMembersAlert(message, 'danger');
+        });
+    }
+
+    $(document).ready(function() {
       $('input.icheck-11').iCheck({
         checkboxClass: 'icheckbox_square-blue',
         radioClass: 'iradio_square-blue'
@@ -55,6 +126,97 @@ define(
           $('.cd-panel-2').removeClass('is-visible');
           event.preventDefault();
         }
+      });
+
+      //open the lateral panel for members
+      $('.btn-members').on('click', function(event) {
+        event.preventDefault();
+        //remove any alert
+        $('#members-alert').empty();
+        //reset the form values
+        resetUserGroupForm();
+        //make panel visible
+        $('.cd-panel-members').addClass('is-visible');
+        //retrieve all the members
+        showUpdateAllMembers();
+        //
+      });
+      //close the lateral panel for members
+      $('.cd-panel-members').on('click', function(event) {
+        if ($(event.target).is('.cd-panel-members') ||
+            $(event.target).is('.cd-panel-close')) {
+          $('.cd-panel-members').removeClass('is-visible');
+          event.preventDefault();
+        }
+      });
+      //add button for the members
+      $(document).on('click', '.cd-panel-members-add', function() {
+        var url = $('.cd-panel-members').data('members-url');
+        var username = $('input[name=\'members-username\']').val();
+        var groupType = $('select[name=\'members-group_type\']').val();
+        // /api/v1/repositories/my-rep/members/groups/<group_type>/users/
+        url += 'groups/' + groupType + '/users/';
+        if (!isEmail(username)) {
+          var message = '<strong>' + username + EMAIL_EXTENSION +
+            '</strong> does not seem to be a valid email';
+          showMembersAlert(message, 'danger');
+          return;
+        }
+        $.ajax({
+          url: url,
+          type: 'POST',
+          data: {username: username}
+        })
+        .done(function() {
+          //reset the values
+          resetUserGroupForm();
+          //show alert
+          var message = '<strong>' + username + EMAIL_EXTENSION +
+            '</strong> added to group <strong>' +
+            formatGroupName(groupType) + '</strong>';
+          showMembersAlert(message);
+          //retrieve the members lists
+          showUpdateAllMembers();
+        })
+        .error(function(data) {
+          //show alert
+          var message = 'Error adding user ' + username + EMAIL_EXTENSION +
+            ' to group ' + formatGroupName(groupType);
+          message = message + '<br>' + data.responseJSON.username[0];
+          showMembersAlert(message, 'danger');
+        });
+      });
+      //remove button for the members
+      $(document).on('click', '.cd-panel-members-remove', function() {
+        var url = $('.cd-panel-members').data('members-url');
+        var username = $(this).data('username');
+        var groupType = $(this).data('group_type');
+        // /api/v1/repositories/my-rep/members/groups/<group_type>/users/<username>/
+        url += 'groups/' + groupType + '/users/' + username + '/';
+        $.ajax({
+          url: url,
+          type: 'DELETE'
+        })
+        .done(function() {
+          //show alert
+          var message = '<strong>' + username + EMAIL_EXTENSION +
+            '</strong> deleted from group <strong>' +
+            formatGroupName(groupType) + '</strong>';
+          showMembersAlert(message);
+          //retrieve the members lists
+          showUpdateAllMembers();
+        })
+        .error(function(data) {
+          //show alert
+          var message = 'Error deleting user <strong>' +
+            username + EMAIL_EXTENSION + '</strong> from group <strong>' +
+            formatGroupName(groupType) + '</strong>';
+          try {
+            message += '<br>' + data.responseJSON.detail;
+          }
+          catch (err) {}
+          showMembersAlert(message, 'danger');
+        });
       });
 
       var repoSlug = $("#repo_slug").val();
--- a/ui/templates/includes/members_panel.html
+++ b/ui/templates/includes/members_panel.html
@@ -1,0 +1,41 @@
+<!-- Members Panel -->
+<div class="cd-panel-members from-right" 
+  data-members-url="{% url 'repo-members' repo.slug %}">
+  <header class="cd-panel-header">
+    <h1>Members for Repository: {{ repo.name }}</h1>
+    <a href="#0" class="cd-panel-close">Close</a>
+  </header>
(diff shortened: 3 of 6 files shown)


pdpinch:
I'm OK with the lack of tests (at the moment), but I'd like @pwilkins to do a QA review.

pwilkins:
@giocalitri Consider changing the commit message to "Added UI to add and remove repository members."  This change will make it more consistent for the release notes.

giocalitri:
@pwilkins Done

ui/static/ui/js/listing.js
```diff
@@ -54,5 +123,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+        var message = '<strong>' + username +
+          '@mit.edu</strong> successfully deleted from group <strong>' +
+          formatGroupName(groupType) + '</strong>';
+        showMembersAlert(message);
+        //retrieve the members lists
+        showUpdateAllMembers();
+      })
+      .error(function(data) {
+        //show alert
+        var message = 'Error trying to delete user <strong>' +
```
pwilkins: Consider changing this to `'Error deleting user <strong>'`.

ui/static/ui/js/listing.js
```diff
@@ -54,5 +123,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+      // /api/v1/repositories/my-rep/members/groups/<group_type>/users/<username>/
+      url += 'groups/' + groupType + '/users/' + username + '/';
+      $.ajax({
+        url: url,
+        type: 'DELETE'
+      })
+      .done(function() {
+        //show alert
+        var message = '<strong>' + username +
+          '@mit.edu</strong> successfully deleted from group <strong>' +
```
pwilkins: Consider changing this to `'@mit.edu</strong> deleted from group <strong>'`.

ui/static/ui/js/listing.js
```diff
@@ -54,5 +123,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+        var message = '<strong>' + username +
+          '@mit.edu</strong> successfully added to group <strong>' +
+          formatGroupName(groupType) + '</strong>';
+        showMembersAlert(message);
+        //retrieve the members lists
+        showUpdateAllMembers();
+      })
+      .error(function(data) {
+        //show alert
+        var message = 'Error trying to add user ' + username +
```
pwilkins: Consider changing this to `'Error adding user '`.

ui/static/ui/js/listing.js
```diff
@@ -54,5 +123,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+        url: url,
+        type: 'POST',
+        data: {username: username}
+      })
+      .done(function() {
+        //reset the values
+        resetUserGroupForm();
+        //show alert
+        var message = '<strong>' + username +
+          '@mit.edu</strong> successfully added to group <strong>' +
```
pwilkins: Consider changing this to `'@mit.edu</strong> added to group <strong>'`.

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,75 @@
...
+  function showUpdateAllMembers() {
+    //retrieve all the members
+    var url = $('.cd-panel-members').data('members-url');
+    $('#cd-panel-members-list').empty();
+    $.ajax({url: url})
+      .done(function(data) {
+        formatUserGroups(data.results, '#cd-panel-members-list');
+      })
+      .error(function() {
+        var message = 'Impossible to retrieve list of members.';
```
pwilkins: Consider changing to `'Unable to retrieve list of members.'`

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,75 @@
...
+  function formatGroupName(string) {
+    string = string.charAt(0).toUpperCase() + string.slice(1);
+    return string.substring(0, string.length - 1);
+  }
+
+  function isMITEmail(kerberosId) {
+    var email = kerberosId + '@mit.edu';
+    var regex = /^([a-zA-Z0-9_.+-])+\@(([a-zA-Z0-9-])+\.)+([a-zA-Z0-9]{2,4})+$/;
+    return regex.test(email);
+  }
```
pwilkins: This currently doesn't check if the email is a valid MIT email.  If this is going to grow up to check whether the name is a valid Kerberos account, please add that as a comment.  Otherwise the name of the function is somewhat misleading.

.jscsrc
```diff
@@ -1,3 +1,4 @@
 {
-  "preset": "google"
+  "preset": "google",
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties"
```
carsongee, line 4: Why not camel case on properties?  We've been doing that elsewhere with javascript

ui/static/ui/js/listing.js
```diff
@@ -54,5 +126,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+    //close the lateral panel for members
+    $('.cd-panel-members').on('click', function(event) {
+      if ($(event.target).is('.cd-panel-members') ||
+          $(event.target).is('.cd-panel-close')) {
+        $('.cd-panel-members').removeClass('is-visible');
+        event.preventDefault();
+      }
+    });
+    //add button for the members
+    $(document).on('click', '.cd-panel-members-add', function() {
```
carsongee: why are you doing this the document?

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
...
+    string = string.charAt(0).toUpperCase() + string.slice(1);
+    return string.substring(0, string.length - 1);
+  }
+
+  /* This is going to grow up to check whether
+   * the name is a valid Kerberos account
+   */
+  function isMITEmail(kerberosId) {
+    var email = kerberosId + '@mit.edu';
+    var regex = /^([a-zA-Z0-9_.+-])+\@(([a-zA-Z0-9-])+\.)+([a-zA-Z0-9]{2,4})+$/;
```
carsongee: this isn't the right regex to validate an email address (http://emailregex.com/)

.jscsrc
noisecapella, line 4: One argument for this is that properties may come straight from JSON which doesn't obey these rules

giocalitri, line 4: because the object coming from the REST API is not camel case and I unless I change it, I cannot fix this

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
...
+
+  function formatGroupName(string) {
+    string = string.charAt(0).toUpperCase() + string.slice(1);
+    return string.substring(0, string.length - 1);
+  }
+
+  /* This is going to grow up to check whether
+   * the name is a valid Kerberos account
+   */
+  function isMITEmail(kerberosId) {
```
carsongee: I'm failing to understand what this does for us.

ui/static/ui/js/listing.js
giocalitri: because the HTML is injected into the page dynamically and unless you attach the event handler to the document, it will not work

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
...
+    var email = kerberosId + '@mit.edu';
+    var regex = /^([a-zA-Z0-9_.+-])+\@(([a-zA-Z0-9-])+\.)+([a-zA-Z0-9]{2,4})+$/;
+    return regex.test(email);
+  }
+
+  function formatUserGroups(userList, dest) {
+    var $dest = $(dest);
+    for (var i = 0; i < userList.length; i++) {
+      $dest.append(
```
carsongee: This would be better with lodash

ui/static/ui/js/listing.js
```diff
@@ -54,5 +126,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+      .error(function(data) {
+        //show alert
+        var message = 'Error adding user ' + username +
+          '@mit.edu to group ' + formatGroupName(groupType);
+        message = message + '<br>' + data.responseJSON.username[0];
+        showMembersAlert(message, 'danger');
+      });
+    });
+    //remove button for the members
+    $(document).on('click', '.cd-panel-members-remove', function() {
```
carsongee: again, why on document?

ui/static/ui/js/listing.js
giocalitri: this is checking in the form if the kerberosId + `@mit.org` is a valid email

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
 define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
```
carsongee: We should change the jQuery here to $

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
...
+    $('input[name=\'members-username\']').val('');
+    $('select[name=\'members-group_type\']').prop('selectedIndex', 0);
+  }
+
+  function showMembersAlert(message, mtype) {
+    //default value for mtype
+    mtype = typeof mtype !== 'undefined' ? mtype : 'success';
+    //reset all the classes
+    $('#members-alert').html(
+      '<div class="alert alert-' + mtype +
```
carsongee: Again, might be better with _., but eventually react should do this, so it may be a moot point if we are going to change it later anyway. Up to you, but you should file an issue to do this all with react

ui/templates/includes/members_panel.html
```diff
@@ -0,0 +1,41 @@
+<!-- Members Panel -->
```
carsongee, line 1: replace tabs with spaces, two spaces per level

ui/templates/repo_base.html
```diff
@@ -6,9 +6,14 @@
 	  	<a href="{% url 'upload' repo.slug %}"><i class="fa fa-arrow-circle-o-right"></i> Import Course</a>
 	  </li>
 	{% endif %}
+	{% if 'manage_repo_users' in perms_on_cur_repo %}
```
carsongee: tabs -> spaces

ui/templates/repository.html
```diff
@@ -56,6 +56,10 @@
   </div>
 </div>
 {% include "includes/resource_panel.html" %}
+{% if 'manage_repo_users' in perms_on_cur_repo %}
+	  {% include "includes/members_panel.html" %}
```
carsongee: tabs -> spaces

ui/static/ui/js/listing.js
giocalitri: my code cannot be currently merged with master and I already noticed the differences

ui/static/ui/js/listing.js
giocalitri: same reason explained before

carsongee:
With the models and validation needed here, seems like another place we need an M in our client side MVC.

.jscsrc
carsongee, line 4: ah, makes sense, thanks

ui/static/ui/js/listing.js
carsongee: um, I think you were replying to something different, it is still wrongly jQuery in master, and should be $ since we are using it as such everywhere

ui/static/ui/js/listing.js
carsongee: Doesn't this really just ends need to be a check for the @ symbol? .   I think it is maybe poorly named as well.  If I pass beaver@mit.edu this will come out false, but just beaver will be true.  I think it would be better defined as `isNotEmailAddress(username)`, or `isUsername(username)` or flip the logic in the check and call it, `isEmailAddress`.

carsongee:
The minus button is off center for me:
image
but the plus is fine:
image

carsongee:
It looks like you are checking to see if they have logged in already.  I thought we had wanted to go ahead and allow them to add potentially invalid users and just create those accounts so that you can give permissions to users who haven't logged in yet.

ui/static/ui/js/listing.js
```diff
@@ -1,6 +1,78 @@
...
+  function formatGroupName(string) {
+    string = string.charAt(0).toUpperCase() + string.slice(1);
+    return string.substring(0, string.length - 1);
+  }
+
+  /* This is going to grow up to check whether
+   * the name is a valid Kerberos account
+   */
+  function isMITEmail(kerberosId) {
+    var email = kerberosId + '@mit.edu';
```
carsongee: This ends up being really confusing in the UI:
image
when my e-mail address on the system is actually `x@1.com`  I think you should just show the username only.  That will be more flexible and doesn't require this to be hard coded in if anyone else wanted to use this.

ui/static/ui/js/listing.js
```diff
@@ -54,5 +126,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+      // /api/v1/repositories/my-rep/members/groups/<group_type>/users/<username>/
+      url += 'groups/' + groupType + '/users/' + username + '/';
+      $.ajax({
+        url: url,
+        type: 'DELETE'
+      })
+      .done(function() {
+        //show alert
+        var message = '<strong>' + username +
+          '@mit.edu</strong> deleted from group <strong>' +
```
carsongee: nit: The deletion might be nicer as `alert-warning` since you did do something destructive instead of green

ui/static/ui/js/listing.js
carsongee: sorry, this comment is totally on the wrong line of code, should have been: https://github.com/mitodl/lore/pull/271/files#diff-5880d9dfd3a8657f9866e255341629a9R27

ui/static/ui/js/listing.js
carsongee: ah, missed that it was added after documentReady, hadn't made it to the template piece yet.

ui/static/ui/js/listing.js
```diff
@@ -54,5 +126,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+        url: url,
+        type: 'POST',
+        data: {username: username}
+      })
+      .done(function() {
+        //reset the values
+        resetUserGroupForm();
+        //show alert
+        var message = '<strong>' + username +
+          '@mit.edu</strong> added to group <strong>' +
```
carsongee: another place `@mit.edu` is not needed and at the very least it is hardcoded several times instead of being in one place

ui/static/ui/js/listing.js
```diff
@@ -54,5 +126,96 @@ define(['jquery', 'bootstrap', 'icheck', 'csrf'], function(jQuery) {
...
+          formatGroupName(groupType) + '</strong>';
+        showMembersAlert(message);
+        //retrieve the members lists
+        showUpdateAllMembers();
+      })
+      .error(function(data) {
+        //show alert
+        var message = 'Error adding user ' + username +
+          '@mit.edu to group ' + formatGroupName(groupType);
```
carsongee: and another

ui/templates/includes/members_panel.html
```diff
@@ -0,0 +1,41 @@
...
+    <div class="cd-panel-content">
+    	<div id="members-alert"></div>
+		<div class="container-fluid">
+			<div class="row">
+				<div class="col-sm-7">
+					<div class="form-group">
+						<div class="input-group">
+					      	<input type="text" name="members-username"
+					      	  placeholder="kerberos id" class="form-control">
+					      	<div class="input-group-addon">@mit.edu</div>
```
carsongee: I would remove this, and maybe do a placeholder instead of `beaver@mit.edu` or similar, also hardcoded again

ui/static/ui/js/listing.js
giocalitri: changed to

```
define(
  ['jquery', 'setup_manage_taxonomies', 'bootstrap', 'icheck', 'csrf'],
  function($, setupManageTaxonomies) {
    'use strict';
```

is this what you meant?

ui/static/ui/js/listing.js
giocalitri: I was under the impression that we were only supporting `@mit.edu`: if this is not the case, also the backend is wrong.
@pdpinch ?

ui/templates/includes/members_panel.html
giocalitri: the idea was not simplify the form to include only the part before `@mit.edu` to not strip this before the request to the REST API

giocalitri:
@carsongee  "It looks like you are checking to see if they have logged in already. I thought we had wanted to go ahead and allow them to add potentially invalid users and just create those accounts so that you can give permissions to users who haven't logged in yet."

I think I was pretty clear about that constraint when I wrote the REST API: the user needs to login at least once.
I'm fine adding this new feature, but I guess I want to know which API I can call to check if the user is a valid kerberos user.

pdpinch:
I told Giovanni it was ok to expect the user to exist, since this the way most of our other systems work. I agree that it would be nice to do something more user-friendly -- like send an invitation -- but this is OK for this PR.

carsongee:
Ok, I was just hoping to avoid having to do some manual work from the ops side by just creating any user they add and letting CAS sort it out.

ui/static/ui/js/listing.js
carsongee: yep

carsongee:
will merge after tests pass :+1:

pdpinch:
closed by #203

Issue #169, opened 2015-06-15
Add a link to request permission to create repositories


pdpinch:
When a user doesn't have permission to create repositories, replace the button with a link to request it by e-mail (using the django-support-email in settings)

pdpinch:
The setting e-mail should be mitx-support@mit.edu

pwilkins:
@pdpinch Once the link to request permission is clicked and the email is sent, what should happen next?  Flash a message that an email has been sent?  Redirect to another page?

pdpinch:
I thought we were just going to use a mailto: link to open in the user's mail client. 

If it's automatically sending a message, I guess there should be some signal to the user that the message has been sent. Anything that's easy would be fine: could be http://getbootstrap.com/javascript/#alerts

pwilkins:
A mailto: link would be far simpler and provide us with more information.  I'll do that.

carsongee:
:+1: to mailto, but you can just use django messages to do this type of thing.  That is what does the nice little green login confirmation text after login now.  The template just automatically picks those up and shows them

pdpinch:
@carsongee @bdero Jolyon saw the link on lore-ci on Friday, but he noticed that the e-mail address wasn't configured properly -- It's still going to support@example.com

pdpinch:
@pwilkins can you check to see if this is working now?

PR #278, opened 2015-07-01
Add link to request create repository permission

pwilkins:
Add `mailto:` link to welcome page. The link opens the user's email client
with the `to:` field filled in by the contents of `LORE_SUPPORT_EMAIL`.

Fixes: #169
[finishes #97013414]

### DIFF
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -225,6 +225,7 @@ EMAIL_PORT = get_var('LORE_EMAIL_PORT', 25)
 EMAIL_HOST_USER = get_var('LORE_EMAIL_USER', '')
 EMAIL_HOST_PASSWORD = get_var('LORE_EMAIL_PASSWORD', '')
 EMAIL_USE_TLS = get_var('LORE_EMAIL_TLS', False)
+EMAIL_SUPPORT = get_var('LORE_SUPPORT_EMAIL', 'support@example.com')
 DEFAULT_FROM_EMAIL = get_var('LORE_FROM_EMAIL', 'webmaster@localhost')
 
 # e-mail configurable admins
--- a/ui/templates/welcome.html
+++ b/ui/templates/welcome.html
@@ -3,7 +3,12 @@
     <div class="section-header">
         <div class="row">
             <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a>{% endif %}</div>
+                                                     class="btn btn-primary btn-large">Create repository</a>
+                {% else %}
+                    <a href="mailto:{{ support_email }}?subject=[LORE-app]Request%20permission%20to%20create%20repositories" role="button"
+                       class="btn btn-primary btn-large">Request permission to create repositories</a>
+                {% endif %}
+            </div>
         </div>
     </div>
 
--- a/ui/tests/test_learningresources_views.py
+++ b/ui/tests/test_learningresources_views.py
@@ -55,6 +55,8 @@ class TestViews(LoreTestCase):
         body = self.assert_status_code("/", HTTP_OK, return_body=True)
         self.assertTrue("<title>MIT - LORE </title>" in body)
         self.assertTrue('>Create repository</a>' in body)
+        self.assertFalse('Request permission to create '
+                         'repositories</a>' in body)
 
     def test_get_home_norepo(self):
         """Home Page with no authorization to create repositories"""
@@ -67,6 +69,8 @@ class TestViews(LoreTestCase):
         self.assertTrue("<title>MIT - LORE </title>" in body)
         self.assertFalse('<a href="/lore/create_repo/">'
                          'Create repository</a>' in body)
+        self.assertTrue('Request permission to create '
+                        'repositories</a>' in body)
 
     def test_create_repo_post(self):
         """Create repo."""
--- a/ui/views.py
+++ b/ui/views.py
@@ -173,7 +173,10 @@ def welcome(request):
     return render(
         request,
         "welcome.html",
-        {"repos": get_repos(request.user.id)}
+        {
+            "repos": get_repos(request.user.id),
+            "support_email": settings.EMAIL_SUPPORT,
+        }
     )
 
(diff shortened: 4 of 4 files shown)


ui/templates/welcome.html
```diff
@@ -3,7 +3,11 @@
     <div class="section-header">
         <div class="row">
             <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a>{% endif %}</div>
+                                                     class="btn btn-primary btn-large">Create repository</a>
```
carsongee, line 6: the spacing is really weird here, can you line wrap at less, something like:

``` html
<div class="col-md-2 col-md-offset-5">
  {% if perms.learningresources.add_repository %}
  <a href="{% url 'create_repo' %}" role="button" class="btn btn-primary btn-large">
    Create repository
  </a>
...
```

lore/settings.py
```diff
@@ -224,6 +224,7 @@ def get_var(name, default):
 EMAIL_HOST_USER = get_var('LORE_EMAIL_USER', '')
 EMAIL_HOST_PASSWORD = get_var('LORE_EMAIL_PASSWORD', '')
 EMAIL_USE_TLS = get_var('LORE_EMAIL_TLS', False)
+EMAIL_SUPPORT = get_var('LORE_SUPPORT_EMAIL', 'mitx-support@mit.edu')
```
carsongee: Not sure if we want to default this to a real address, maybe `support@example.com` as the default

ui/templates/welcome.html
pwilkins, line 6: It looked weird to me, too, but I just added to what was there.  I'll change it.

pwilkins:
@carsongee @giocalitri @Ferdi please review.  Consider also whether this is the user experience we want.

ui/templates/welcome.html
```diff
@@ -2,8 +2,17 @@
...
         <div class="row">
-            <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a>{% endif %}</div>
+            <div class="col-md-2 col-md-offset-5">
+                {% if perms.learningresources.add_repository %}
+                    <a href="{% url 'create_repo' %}"
+                       role="button"
+                       class="btn btn-primary btn-large">Create repository</a>
+                {% else %}
+                    <a href="mailto:mitx-support@mit.edu?subject=[LORE-app]Request%20permission%20to%20create%20repositories"
```
giocalitri: I do not understand why you hardcoded the email here and you did not take it from the settings.EMAIL_SUPPORT

ui/templates/welcome.html
```diff
@@ -2,8 +2,17 @@
 {% block content %}
     <div class="section-header">
         <div class="row">
-            <div class="col-md-2 col-md-offset-5">{% if perms.learningresources.add_repository %}<a href="{% url 'create_repo' %}" role="button"
-                                                     class="btn btn-primary btn-large">Create repository</a>{% endif %}</div>
+            <div class="col-md-2 col-md-offset-5">
+                {% if perms.learningresources.add_repository %}
+                    <a href="{% url 'create_repo' %}"
+                       role="button"
+                       class="btn btn-primary btn-large">Create repository</a>
```
giocalitri: Did you check the result on the browsers? I wrote the template in that weird way because I did not want to put spaces between the different tags.

ui/tests/test_learningresources_views.py
```diff
@@ -48,6 +47,8 @@ def test_get_home_norepo(self):
         self.assertTrue("<title>MIT - LORE </title>" in body)
         self.assertFalse('<a href="/lore/create_repo/">'
                          'Create repository</a>' in body)
+        self.assertTrue('Request permission to create '
+                        'repositories</a>' in body)
```
giocalitri, line 73: I guess you also need to add a test to check that the string is not there when the user has already the right to create a repo

pdpinch:
Based on Giovanni's comments, I think there are three things to do here:
- [x] instead of hard-coding mitx-support@mit.edu in the template, use LORE_SUPPORT_EMAIL as in the description
- [x] test how it looks in Chrome. In other HTML views, Chrome interpreted white space in the template differently 
- [x] Write a test for when the link is not present (or open a new issue for the test)

ui/tests/test_learningresources_views.py
pwilkins, line 73: I added a test to test_get_home().  Is that sufficient, or is that still not the right combination of conditions?

ui/templates/welcome.html
pwilkins, line 6: On @giocalitri comment, I changed it back for cross-browser support.

ui/tests/test_learningresources_views.py
giocalitri, line 73: No, that's enough

ui/views.py
```diff
@@ -168,7 +169,11 @@ def welcome(request):
     return render(
         request,
         "welcome.html",
-        {"repos": get_repos(request.user.id)}
+        {
+            "repos": get_repos(request.user.id),
+            "support_email": get_var('EMAIL_SUPPORT',
+                                     'support@example.com')
```
giocalitri: I think you need to take the value from the setting and not from the environment
So if you import the settings

```
...
from django.conf import settings
...
```

then your code is just:

```
...
"support_email": settings.EMAIL_SUPPORT,
...
```

giocalitri:
you also need to rebase master

ui/templates/welcome.html
giocalitri, line 6: if you want to keep the nice indentation, you an use the `{% spaceless %}` tag.
[look here](https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#spaceless)

giocalitri, line 6: made a couple of tests with the `{% spaceless %}` tag and it does not solve the problem of an extra space in front of the string in the button.
So I'm going to merge it as it is

PR #280, opened 2015-07-01
Fixed error message behavior for manage taxonomies tab

noisecapella:


carsongee:
This still isn't clearing the message for me

pwilkins:
It isn't working for me either.  Firefox or Chrome.  The error message also doesn't go away when I close and reopen the panel.

PR #284, opened 2015-07-01
Release 0.3.0

carsongee:


RELEASE.rst
```diff
@@ -1,6 +1,25 @@
 Release Notes
 -------------
 
+Version 0.3.0
+=============
+
+- Added UI to add and remove repository members.
+- Added form for adding new vocabularies.
+- Added manage taxonomies panel and button
+- Added Requires.io badge
```
pdpinch: no need for requires.io badge in the release notes

RELEASE.rst
```diff
@@ -1,6 +1,24 @@
...
+Version 0.3.0
+=============
+
+- Added UI to add and remove repository members.
+- Added form for adding new vocabularies.
+- Added manage taxonomies panel and button
+- Tests for members REST.
+- REST for repo members
+- Implemented taxonomy model delete cascading.
+- Renamed &#39;Copy to Clipboard&#39; to &#39;Select Text&#39;
```
pdpinch: somehow I introduced these double-quote entities. They need to be replaced with " or similar.

RELEASE.rst
```diff
@@ -1,6 +1,24 @@
 Release Notes
 -------------
 
+Version 0.3.0
+=============
+
+- Added UI to add and remove repository members.
+- Added form for adding new vocabularies.
+- Added manage taxonomies panel and button
+- Tests for members REST.
```
pdpinch: I think we should drop this item from the release notes. It's part of the next item, "REST API for repo members"

RELEASE.rst
```diff
@@ -1,6 +1,24 @@
...
+- REST for repo members
+- Implemented taxonomy model delete cascading.
+- Renamed &#39;Copy to Clipboard&#39; to &#39;Select Text&#39;
+- Removed duplicate m2m field.
```
pdpinch: Does this need to be in the release notes? I don't think so.

RELEASE.rst
pdpinch: Also, as @pwilkins pointed out, the new button is "Select XML" not "Select Text"

PR #289, opened 2015-07-02
Fixed buttons alignment problem in members panel.

giocalitri:
Bug caused by spaces between tags: Firefox ignores them, Chrome (and Safari) do not.

Fixes #282

PR #292, opened 2015-07-02
Allow only usernames and not emails in members input form

### LINKED ISSUE #290: Allow only usernames in the "add members" form
Currently there is a weird user experience if the user types a full email: limit the field to usernames.

giocalitri:
to be reviewed and merged ONLY AFTER #289 

Fixes #290


pwilkins:
This works as intended, though I won't attempt to verify the email regex.  :-)  When the tests pass I'll merge.

PR #295, opened 2015-07-02
Added Members REST API docs

### LINKED ISSUE #129: Members REST API for managing users in repositories [#96516348]

giocalitri:
closes #129

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -428,25 +428,10 @@ Create a repository by providing its JSON representation.
 
 Represents users / members of the repository
 
-## Member [/repositories/{repo_slug}/members/{member_id}/]
+## Members Collection [/repositories/{repo_slug}/members/]
 
 + Parameters
     + repo_slug: `physics-1` (string, required) - slug for the repository
-    + member_id: `107` (number, required) - ID of the user in the system
-
-### View a Member [GET]
-
-+ Response 200 (application/json)
-
-        {
-            "tbd": "tbd"
-        }
-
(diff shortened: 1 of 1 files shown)


apiary.apib
```diff
@@ -428,35 +428,171 @@ Create a repository by providing its JSON representation.
...
-### View a Member [GET]
+
+### List All Members [GET]
 
 + Response 200 (application/json)
 
+    + Body
+
+            {
+                "count": 5,
```
noisecapella: Change to `4` just for consistency's sake

apiary.apib
```diff
@@ -428,35 +428,171 @@ Create a repository by providing its JSON representation.
...
+
+        {
+            "username": "james"
+        }
+
+## Member for a Group [/repositories/{repo_slug}/members/groups/{group_type}/users/{member_username}/]
+
++ Parameters
+    + repo_slug: `physics-1` (string, required) - slug for the repository
+    + member_username: `james` (string, required) - username of the user in the system
```
noisecapella: Can we swap this line with the next one so that they appear in the same order as the parameters in the URL?

giocalitri:
Done!

Issue #273, opened 2015-07-01
Implement code coverage for JSX files


noisecapella:
I think this is a trivial fix, changing the extension used or something to that effect

PR #296, opened 2015-07-02
Added coverage for JSX files

noisecapella:
This PR does a few javascript related things:
- Added a `REQUIRE_PATHS` global so our tests and Django use the same set of paths for configuration
- Made test configuration more flexible
- Added coverage for JSX files
- Removed empty file manage_vocabularies.jsx
- Renamed test-listing back to test-main (this is independent of listing.js which underwent a name change at the same time, so it makes sense to use test-main here)
- Added all relevant javascript files to code coverage. Previously only files which were explicitly included were counted in our code coverage count.
- Excluded listing.js from testing since we can't test it independently at the moment


noisecapella:
Oh, and fixes #273

carsongee:
Very nice :+1:

Issue #182, opened 2015-06-17
Implement root for API


noisecapella:
This would ideally be part of #106 but I'd like that to get merged ASAP and the behavior of what happens when you go to `/api/v1/` is not critical to anything. In #106 the root just returns 404 but the docs say it should return a JSON list of endpoints

amir-qayyum-khan:
Done @pdpinch

PR #300, opened 2015-07-03
Implemented root api

amir-qayyum-khan:
Hi,

In this PR i added root view for lore rest service app

https://github.com/mitodl/lore/issues/182
cc @pdpinch @ShawnMilo @carsongee 

screen shot 2015-07-03 at 4 21 25 pm


rest/views.py
```diff
@@ -48,6 +49,39 @@
 from django.http.response import Http404
 
 
+def get_urls(raw_urls, list_urls, urlbase=''):
+    """
+    Recursively builds a list of all the urls in the current project and
+    the name of their associated view
+    """
+    list_urls = list_urls or []
+    from operator import itemgetter
```
giocalitri: is there a particular reason to not put this at the beginning of the file?

rest/views.py
```diff
@@ -48,6 +49,39 @@
 from django.http.response import Http404
 
 
+def get_urls(raw_urls, list_urls, urlbase=''):
+    """
+    Recursively builds a list of all the urls in the current project and
+    the name of their associated view
+    """
+    list_urls = list_urls or []
```
giocalitri, line 69: I guess here you mean that `list_urls` can be `None`.
If this is the case you should change the definition to `def get_urls(raw_urls, list_urls=None, urlbase='')`  otherwise this line does not seem that useful.

rest/views.py
```diff
@@ -48,6 +49,39 @@
...
+        full_url = (urlbase + entry.regex.pattern).replace('^', '')
+        if '$' in full_url:
+            full_url = full_url[0: full_url.index("$")]
+        if entry.callback:  # if it points to a view
+            list_urls.append({"pattern": full_url})
+        else:  # if it points to another urlconf, recur!
+            get_urls(entry.url_patterns, list_urls, full_url)
+
+    # sort alphabetically
+    list_urls = sorted(list_urls, key=itemgetter('pattern'))
```
giocalitri: nit: why not sorting in place? Allocating a new variable here is not that useful

rest/views.py
```diff
@@ -48,6 +49,39 @@
...
+
+
+@api_view(('GET',))
+def index(request):
+    """
+    Root service , returns list of end points
+    """
+    import rest.urls as rest_urls
+    list_urls = get_urls(
+        rest_urls.urlpatterns, list_urls=[], urlbase=request.path
```
giocalitri: I guess `list_urls=[]` is not necessary if you have `list_urls = list_urls or []` in `get_urls`

rest/views.py
amir-qayyum-khan: yeah, putting it on top of file gave me error on travis, but was working fine on local env.

giocalitri: what error?

giocalitri:
Don't you need to update the API docs as well?

rest/views.py
giocalitri: just run tests on my local instance of docker and everything looks fine. I moved the line in this way

```

from operator import itemgetter

from django.contrib.auth.models import User
```

rest/views.py
```diff
@@ -48,6 +49,39 @@
...
 
+def get_urls(raw_urls, list_urls, urlbase=''):
+    """
+    Recursively builds a list of all the urls in the current project and
+    the name of their associated view
+    """
+    list_urls = list_urls or []
+    from operator import itemgetter
+    for entry in raw_urls:
+        full_url = (urlbase + entry.regex.pattern).replace('^', '')
```
noisecapella, line 71: String manipulation of regexes is full of potential problems. Why not just create a view which outputs a predefined list of URLs?

noisecapella:
We really should not output regexes in our REST output, especially not twice

rest/views.py
giocalitri, line 71: I actually like the idea of pulling all the URLs automatically from the URL definition. Is there a better way to do this?

noisecapella, line 71: If we used [DefaultRouter](http://www.django-rest-framework.org/api-guide/routers/#defaultrouter) it will automatically create these URLs for us, however it doesn't know about nested resources so we would have a weird combination of autogenerated URL regexes and manually specified ones

amir-qayyum-khan:
Done @noisecapella 

cc @pwilkins 
screen shot 2015-07-07 at 4 35 21 pm

amir-qayyum-khan:
@giocalitri  please guide me where are docs? which doc i need to update.

Thanks

CC @pwilkins

giocalitri:
I was talking about the API docs in the file `apiary.apib`

giocalitri:
In any case me and @noisecapella  had an extensive discussion yesterday about this issue and we are not sure if this is the right way to solve the problem.

**T.L.D.R:** these URLs that you generate give a description of what is available but they are only human readable. To have something really useful we should generate the list of "child" url for each rest view and return all the URLs in that view.

**Full Explanation:**
Let's assume that the app contains 2 repos "foo", "bar", when you hit the URL
`/api/v1/repositories/` you need to get back the full URL of the repositories together with the rest of the data; so something like

```
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "slug": "foo",
            "name": "foo",
            "description": "foo repo",
            "date_created": "2015-07-01T18:23:37.176845Z",
            "url": "/api/v1/repositories/foo/"
        },
        {
            "id": 2,
            "slug": "bar",
            "name": "bar",
            "description": "bar repo",
            "date_created": "2015-07-02T16:26:06.093215Z",
            "url": "/api/v1/repositories/bar/"
        }
    ]
}
```

This part is pretty straight forward, but things become much trickier when you hit other types of URLs, like, for example, `/api/v1/repositories/<repo_slug>/`.
In this case you need to return the infos about the repo, but also all the child URLs that in this case can be `/api/v1/repositories/<repo_slug>/learningresources/`, `/api/v1/repositories/<repo_slug>/members/` etc.

I'm not sure what we want here, but the current implementation is probably too simplistic.
We probably need @Ferdi and @pdpinch opinion about this to understand if implementing this in the right way is the best use of our time.

noisecapella:
Yes, I'm skeptical we need this functionality at all. I originally filed #182 because it was mentioned in the existing apiary docs but I think we need a clear use case from @Ferdi or @pdpinch to know what we're looking for.

If we wanted to move to something more like HATEOAS then we would need to make other changes like @giocalitri suggested.

pwilkins:
@giocalitri @amir-qayyum-khan  Updating the docs in `apiary.apib` is not part of this PR.

amir-qayyum-khan:
@pwilkins I am not sure this functionality i achieved in this PR is correct or not, can you verify it.

Thanks,

noisecapella:
@pwilkins Aren't we always supposed to update `apiary.apib` when we make API changes?

pwilkins:
@noisecapella Yes, you are correct.  I misspoke.

Ferdi:
I agree with @giocalitri and @noisecapella . I suggest we close this PR for now. It's not helpful at this stage.

giocalitri:
OK. Do we want to update #182 with a better explanation of what is needed?

noisecapella:
I'd vote for closing #182 as well. We can file a new issue if we come up with a use case for it

PR #304, opened 2015-07-06
Remove unused patterns to limit memory use

noisecapella:
My laptop has 8GB memory but others with less memory on their Docker containers ran out of memory due to the large amount of imported javascript. On further investigation importing `node_modules` isn't actually necessary so it's removed

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -40,14 +40,6 @@ module.exports = function(config) {
       {
         pattern:'ui/jstests/**/*.jsx',
         included: false
-      },
-      {
-        pattern: 'node_modules/**/*.js',
-        included: false
-      },
-      {
-        pattern: 'node_modules/**/*.jsx',
-        included: false
       }
     ],
 


carsongee:
Commit message tense

pwilkins:
How can I functionally test that this resolves the memory error?

noisecapella:
@pwilkins The error occurs on machines with 4GB or less of memory in their Docker containers. I think you should just be able to check out master and run the javascript tests to reproduce the failure

carsongee:
I have verified it fixes me for my 4GiB boot2docker image

giocalitri:
I confirm it fixes the problem also on my docker machine.

pwilkins:
@noisecapella Was there an issue for this PR?  I'm asking simply so that I can back fill Pivotal.

noisecapella:
@pwilkins No, it was something @carsongee and @giocalitri noticed earlier today

PR #306, opened 2015-07-06
Fixed `KeyError` on repository page if search index is reset.

### LINKED ISSUE #283: KeyError at /repositories/<repo_slug>/ when ES index is not available any more
Edge case (I guess).

Steps to reproduce:
- have a running repository with learning resources inside
- run tox
- reload /repositories/<repo_slug>/

ShawnMilo:
This wouldn't happen in production, but it can happen (for a brief window)
running the application in Docker containers.

Closes #283.

Note about tests:
The system behaves differently during testing than when running. When testing,
a test for this passes without the fix. When Django is running normally,
it will return the `KeyError` after a forced index wipe, but only until
Django is restarted. Then it works normally.

Digging into exactly what changes in Haystack (or Elasticsearch) when Django is restarted (some kind of caching?) is too
time-consuming for the repayment in test coverage. As a result, this branch causes a minor
hit in test coverage.

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -245,16 +245,17 @@ class RepositoryView(FacetedSearchView):
                 )
             qs_prefix = "?{0}&".format("&".join(qs_prefix))
 
-        context.update({
-            "repo": self.repo,
-            "perms_on_cur_repo": get_perms(self.request.user, self.repo),
-            "vocabularies": {
-                k: v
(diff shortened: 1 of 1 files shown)


ShawnMilo:
Deprecated in favor of #344, which includes this fix and a new test from Carson.

PR #310, opened 2015-07-07
Default checked radio button for Manage Taxonomies -> Add Vocabulary

### LINKED ISSUE #291: Default value for Managed/Free Tagging radio button
This is pretty minor but instead of filling in a value when we add the vocabulary if the user didn't select one, we should set a default value in the UI so the user knows which value is being sent to the server

giocalitri:
Fixes #291

### DIFF
--- a/ui/static/ui/js/manage_taxonomies.jsx
+++ b/ui/static/ui/js/manage_taxonomies.jsx
@@ -207,17 +207,19 @@ define('setup_manage_taxonomies', ['reactaddons', 'lodash', 'jquery'],
           </p>
           <p>
             <div className="radio">
-              <label htmlFor="managed_vocabulary_type">
+              <label>
                 <input id="managed_vocabulary_type" type="radio"
                   name="vocabulary_type" value="m"
+                  checked={this.state.vocabulary_type === 'm'}
                   onChange={this.updateVocabularyType} />
                     Managed
               </label>
             </div>
             <div className="radio">
-              <label htmlFor="managed_vocabulary_type">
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -210,14 +210,16 @@ define('setup_manage_taxonomies', ['reactaddons', 'lodash', 'jquery'],
...
                 <input id="managed_vocabulary_type" type="radio"
                   name="vocabulary_type" value="m"
+                  checked={this.state.vocabulary_type === 'm'}
                   onChange={this.updateVocabularyType} />
                     Managed
               </label>
             </div>
             <div className="radio">
-              <label htmlFor="managed_vocabulary_type">
+              <label htmlFor="free_vocabulary_type">
```
noisecapella: We shouldn't need the for attribute if the item being pointed to is contained inside the label element

giocalitri: Ok, but the way it was defined was wrong, so I just linked to the right ID.

giocalitri: Do you want I just remove them?

noisecapella: I suppose it doesn't matter, both with and without are valid, but for cleanliness sake it would be good to remove it

giocalitri: ok, I will remove and push again

PR #316, opened 2015-07-08
Fixed error message on unsupported terms in learning resource

noisecapella:
Looks like I missed an argument during `format()`

### DIFF
--- a/rest/serializers.py
+++ b/rest/serializers.py
@@ -192,7 +192,8 @@ class LearningResourceSerializer(ModelSerializer):
             if not term.vocabulary.learning_resource_types.filter(
                     name=resource_type.name).exists():
                 raise ValidationError(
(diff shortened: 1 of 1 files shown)


pwilkins:
@noisecapella Is there a way for me to verify this change?  I can merge it without verification, but I wonder if there is even a way to see this validation message in action.

noisecapella:
You would need to add a vocabulary and a term then PATCH /api/v1/repositories/repo/learning_resources/1 with {'terms': ['new_term_slug']} to get the error message.

ShawnMilo:
This was accidentally closed in PR #397 (I meant to close 318, typed 316 instead).

@noisecapella, @pwilkins it looks like this was already merged, so is it good to leave this closed?

noisecapella:
Yes, should be fine

PR #323, opened 2015-07-09
Updated Django to 1.8.3

carsongee:


carsongee:
https://docs.djangoproject.com/en/1.8/releases/1.8.3/

giocalitri:
I did a quick regression and everything looks fine to me
(maybe a second opinion is better)

carsongee:
Go ahead and merge this then, we can roll back if we discover new issues, but I'd like to prevent the security issue it addresses

giocalitri:
ok

PR #324, opened 2015-07-09
Allowed blank descriptions for LearningResource model

noisecapella:
Since we populate blank `LearningResource` descriptions during import it makes sense that we allow users to leave descriptions blank too


carsongee:
:+1: once tests pass

Issue #322, opened 2015-07-09
Extend reactjs test coverage


pwilkins:
Extend reactjs test coverage

amir-qayyum-khan:
Done

giocalitri:
what is this about?

noisecapella:
Adding unit tests to cover most of manage_taxonomies.jsx

giocalitri:
is it finished?

noisecapella:
No, it depends on the linked PR

pdpinch:
@noisecapella #328 is merged now. Can this be closed?

noisecapella:
Completely forgot to answer that :)

PR #309, opened 2015-07-07
Added learning resource UI to edit description and terms

noisecapella:

PR #328, opened 2015-07-10
Extended tests for manage_taxonomies.jsx file

amir-qayyum-khan:
Hi

In this PR I wrote test cases for all reactjs classes i.e VocabularyComponent, TermComponent, AddTermsComponent, AddVocabulary, TaxonomyComponent

https://github.com/mitodl/lore/issues/322

@pdpinch @pwilkins @carsongee @noisecapella

### DIFF
--- a/ui/jstests/test_manage_taxonomies.jsx
+++ b/ui/jstests/test_manage_taxonomies.jsx
@@ -5,13 +5,10 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
 
   var VocabularyComponent = ManageTaxonomies.VocabularyComponent;
   var waitForAjax = TestUtils.waitForAjax;
-
-  var addTermResponse = {
-    "id": 9,
-    "slug": "aa",
-    "label": "aa",
-    "weight": 1
-  };
+  var TermComponent = ManageTaxonomies.TermComponent;
+  var AddTermsComponent = ManageTaxonomies.AddTermsComponent;
+  var AddVocabulary = ManageTaxonomies.AddVocabulary;
+  var TaxonomyComponent = ManageTaxonomies.TaxonomyComponent;
   var vocabulary = {
     "id": 1,
     "slug": "difficulty",
@@ -20,6 +17,9 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
     "vocabulary_type": "f",
     "required": false,
     "weight": 2147483647,
+    "learning_resource_types": [
+      "course"
+    ],
     "terms": [
       {
         "id": 1,
@@ -35,18 +35,153 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       }
     ]
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  /**
+   * Common assertions for Add terms, use in term and taxonomy
+   * components tests
+   * @param {QUnit.assert} assert - For assertions
+   * @param {ReactComponent} component - Root DOM object in react class
+   * @return {ReactComponent} devItems - DOM object of class 'panel-body'
+   * */
+  function  assertAddTermCommon(assert, component) {
+    // check term title
+    var listOfLinks = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        component,
+        'a'
+    );
+    assert.equal(listOfLinks.length, 2);
+    var linkHeader = React.findDOMNode(listOfLinks[0]);
+    assert.equal(linkHeader.innerHTML, vocabulary.name);
+    //test items
+    var devItems = React.addons.TestUtils.
+      findRenderedDOMComponentWithClass(
+        component,
+        'panel-body'
+    );
+    var itemList = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        devItems,
+        'li'
+    );
+    assert.equal(itemList.length, 2);
+    return devItems;
   }
   QUnit.module('Test taxonomies panel', {
     beforeEach: function() {
+      var learningResourceTypes = {
+        "count": 8,
+        "next": null,
+        "previous": null,
+        "results": [
+          {
+            "name": "course"
+          },
+          {
+            "name": "chapter"
+          },
+          {
+            "name": "sequential"
+          },
+          {
+            "name": "vertical"
+          },
+          {
+            "name": "html"
+          },
+          {
+            "name": "video"
+          },
+          {
+            "name": "discussion"
+          },
+          {
+            "name": "problem"
+          }
+        ]
+      };
+      var listOfVocabularies = {
+        "count": 1,
+        "next": null,
+        "previous": null,
+        "results": [
+          {
+            "id": 1,
+            "slug": "difficulty",
+            "name": "difficulty",
+            "description": "easy",
+            "vocabulary_type": "f",
+            "required": false,
+            "weight": 2147483647,
+            "terms": vocabulary.terms
+          }
+        ]
+      };
+      var listOfTerms = {
+        "count": 2,
+        "next": null,
+        "previous": null,
+        "results": vocabulary.terms
+      };
+      var term = {
+        "id": 9,
+        "slug": "test",
+        "label": "test",
+        "weight": 1
+      };
+      var duplicateVocabularyResponse = {
+        "non_field_errors":[
+          "The fields repository, name must make a unique set."
+        ]
+      };
       TestUtils.setup();
       TestUtils.initMockjax({
-        url: "/api/v1/repositories/*/vocabularies/*/terms/",
-        responseText: addTermResponse,
+        url: "/api/v1/repositories/repo/vocabularies/easy/terms/",
+        type: "POST",
+        status: 400
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/",
+        responseText: term,
         type: "POST"
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/",
+        responseText: listOfTerms,
+        type: "GET"
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/",
+        contentType: "application/json; charset=utf-8",
+        responseText: vocabulary,
+        dataType: 'json',
+        type: "POST"
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo2/vocabularies/",
+        type: "POST",
+        status: 400
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo3/vocabularies/",
+        contentType: "application/json; charset=utf-8",
+        responseText: duplicateVocabularyResponse,
+        dataType: 'json',
+        type: "POST",
+        status: 400
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/",
+        contentType: "application/json; charset=utf-8",
+        responseText: listOfVocabularies,
+        dataType: 'json',
+        type: "GET"
+      });
+      TestUtils.initMockjax({
+        url: "/api/v1/learning_resource_types/",
+        contentType: "application/json; charset=utf-8",
+        responseText: learningResourceTypes,
+        dataType: 'json',
+        type: "GET"
       });
     },
     afterEach: function() {
@@ -54,30 +189,94 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
     }
   });
 
+  QUnit.test('Assert that TermComponent renders properly',
+    function(assert) {
+      assert.ok(TermComponent, "class object not found");
+      var term = {
+        "id": 9,
+        "slug": "test",
+        "label": "test",
+        "weight": 1
+      };
+      var termComponentRendered = React.addons.TestUtils.
+        renderIntoDocument(
+          <TermComponent
+            term={term}
+          />
+        );
+
+      var labelComponent = React.addons.TestUtils.
+        findRenderedDOMComponentWithTag(
+          termComponentRendered,
+          'label'
+      );
+      //testing label value render
+      var label = labelComponent.getDOMNode();
+      assert.equal(label.innerHTML, term.label);
+    }
+  );
+
   QUnit.test('Assert that VocabularyComponent renders properly',
     function(assert) {
       assert.ok(VocabularyComponent, "class object not found");
-
       var done = assert.async();
 
       var addTermCalled = 0;
       var addTerm = function() {
         addTermCalled += 1;
       };
-
+      var reportError = function() {};
       var afterMount = function(component) {
         var node = React.findDOMNode(component);
         var textbox = $(node).find("input")[0];
-        React.addons.TestUtils.Simulate.keyUp(textbox, {key: "x"});
-        assert.equal(addTermCalled, 0);
 
-        React.addons.TestUtils.Simulate.keyUp(textbox, {key: "Enter"});
-        waitForAjax(1, function() {
-          assert.equal(addTermCalled, 1);
-          done();
+        // check term title
+        var listOfLinks = React.addons.TestUtils.
+          scryRenderedDOMComponentsWithTag(
+            component,
+            'a'
+        );
+        assert.equal(listOfLinks.length, 2);
+        var linkHeader = React.findDOMNode(listOfLinks[0]);
+        assert.equal(linkHeader.innerHTML, vocabulary.name);
+
+        //test items
+        var devItems = React.addons.TestUtils.
+          findRenderedDOMComponentWithClass(
+            component,
+            'panel-body'
+        );
+        var itemList = React.addons.TestUtils.
+          scryRenderedDOMComponentsWithTag(
+            devItems,
+            'li'
+        );
+        assert.equal(itemList.length, 2);
+        //test enter text in input text
+        var inputNode = React.addons.TestUtils.
+          findRenderedDOMComponentWithTag(
+            component,
+            'input'
+        );
+        React.addons.TestUtils.Simulate.change(
+          inputNode,
+          {target: {value: 'test12'}}
+        );
+        component.forceUpdate(function() {
+          assert.equal(
+            'test12',
+            component.state.newTermLabel
+          );
+          React.addons.TestUtils.Simulate.keyUp(textbox, {key: "x"});
+          assert.equal(addTermCalled, 0);
+
+          React.addons.TestUtils.Simulate.keyUp(textbox, {key: "Enter"});
+          waitForAjax(1, function() {
+            assert.equal(addTermCalled, 1);
+            done();
+          });
         });
       };
-
       React.addons.TestUtils.
         renderIntoDocument(
           <VocabularyComponent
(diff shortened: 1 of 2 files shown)


noisecapella:
There will be some changes coming in regarding the javascript tests so you will need to adjust this PR after my PR is reviewed and merged (#309). Hopefully we'll get a good process nailed down soon for javascript unit tests.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -32,17 +45,79 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       }
     ]
   };
+  var vocabularies = [
+    {
+      "vocabulary": vocabulary,
+      "terms": vocabulary.terms
+    }
+  ];
+  var saveVocabularyResponse;
```
noisecapella: It looks like you could move this and `updateParent` inside the test where they are being used. In general we want to define variables as close to their use as possible

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -32,17 +45,79 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
   function reportError(msg) {
```
noisecapella: Also move this inside the test where it is used

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -32,17 +45,79 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+  }
   QUnit.module('Test taxonomies panel', {
     beforeEach: function() {
       $.mockjax({
-        url: "/api/v1/repositories/*/vocabularies/*/terms/",
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/",
         contentType: "application/json; charset=utf-8",
-        responseText: addTermResponse,
+        responseText: term,
+        responseTime: 0,
```
noisecapella: I've added `TestUtils.initMockjax` which you should use instead in the future, it defines some common settings and also counts mocked AJAX requests. In general look at `test_learning_resources.jsx` and use that for an example.

We want to avoid just making jQuery synchronous because then we won't be able to detect accidental race conditions which may occur. I'll go into more detail below on what to do.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -53,10 +128,29 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       $.ajaxSetup({async: true});
```
noisecapella: This is removed, we shouldn't need it anymore. I added `TestUtils.waitForAjax` to allow for asynchronous testing of AJAX calls.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -66,6 +160,47 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
             repoSlug="repo"
           />
         );
+      //test initial state
+      assert.equal(
+        vocabularyComponentRendered.getInitialState().terms.length,
```
noisecapella: You should be able to just assert `state.terms.length` here. `state` should have `getInitialState()`s values before any `setState` is called.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -32,17 +45,79 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    var inputNodes = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        formNode,
+        'input'
+      );
+    assert.equal(inputNodes.length, 4);
+    inputNodes[0].value = "TestA";
+    inputNodes[1].value = "TestA";
+    React.addons.TestUtils.Simulate.submit(formNode);
+    assert.equal(
```
noisecapella: You will need to use `TestUtils.waitForAjax` and `assert.async()` to wait for the POST to complete before finishing the test assertions

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -32,17 +45,79 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
-        responseText: addTermResponse,
+        responseText: term,
+        responseTime: 0,
+        dataType: 'json'
+      });
+      $.mockjax({
+        url: "/api/v1/repositories/repo/vocabularies/",
+        contentType: "application/json; charset=utf-8",
+        responseText: vocabularyWithoutTerms,
         responseTime: 0,
```
noisecapella: Remember to remove all `responseTime: 0` since we will be testing asynchronous AJAX calls

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -66,6 +160,47 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        inputNode,
+        {target: {value: 'test12'}}
+      );
+      assert.equal(
+        'test12',
+        vocabularyComponentRendered.state.newTermLabel
+      );
+      //test enter key
       var randomEvent = $.Event("keyup", {keyCode: 64});
       var eventEnter = $.Event("keyup", {keyCode: 13});
```
noisecapella: It is recommended to use `React.addons.TestUtils.Simulate.keyUp` instead here since React will pass instances of `SyntheticEvent` to event handlers and not regular javascript events like you have here.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -75,9 +210,93 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      var addVocabularyRendered = React.addons.TestUtils.
+        renderIntoDocument(
+          <AddVocabulary
+            vocabularies={vocabularies}
+            repoSlug="repo"
+            updateParent={updateParent}
+          />
+        );
+      assert.equal(
+        addVocabularyRendered.getInitialState().name,
```
noisecapella: Test against `state`, not `getInitialState()`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -75,9 +210,93 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      assert.ok(TaxonomyComponent, "class object not found");
+      var taxonomyComponentRendered = React.addons.TestUtils.
+        renderIntoDocument(
+          <TaxonomyComponent
+            vocabularies={vocabularies}
+            repoSlug="repo"
+          />
+        );
+      assert.equal(
+        taxonomyComponentRendered.getInitialState().vocabularies.length,
```
noisecapella: Use `state` in place of `getInitialState()` here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -75,9 +210,93 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      assert.ok(AddTermsComponent, "class object not found");
+      var addTermsComponentRendered = React.addons.TestUtils.
+        renderIntoDocument(
+          <AddTermsComponent
+            vocabularies={vocabularies}
+            repoSlug="repo"
+          />
+        );
+      assert.equal(
+        addTermsComponentRendered.getInitialState().errorText,
```
noisecapella: replace `getInitialState()` here

noisecapella:
I also did some refactoring in PR #340 FYI, you may want to adjust your code accordingly. The most significant change is that `state.terms` was removed since it was a duplicate piece of state, a callback is used instead to update `TaxonomyComponent`

noisecapella:
One more thing, I noticed that the vocabulary type radio box is not covered by tests, and that failures of various AJAX calls are not covered either. You can look at `test_learning_resources.jsx` for examples of this.

Thanks for all your help!

amir-qayyum-khan:
@noisecapella  please review, I think it is ready to merge

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,25 +69,152 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      }
+    ]
+  };
+  var listOfTerms = {
+    "count": 2,
+    "next": null,
+    "previous": null,
+    "results": vocabulary.terms
+  };
+  var saveVocabularyResponse;
```
noisecapella: This variable is used in two tests. Please create a local variable within each test instead. We want to limit global state as much as possible, especially given the complexity of our test framework. The way it is now, one test may affect another tests result which will make test failures hard to debug.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,25 +69,152 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      scryRenderedDOMComponentsWithTag(
+        formNode,
+        'input'
+      );
+    assert.equal(inputNodes.length, 4);
+    inputNodes[0].value = "TestA";
+    inputNodes[1].value = "TestA";
+    inputNodes[2].checked = true;
+    inputNodes[3].checked = false;
+    React.addons.TestUtils.Simulate.submit(formNode);
```
noisecapella: You should assert a change in state here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,25 +69,152 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    inputNodes[0].value = "TestA";
+    inputNodes[1].value = "TestA";
+    inputNodes[2].checked = false;
+    inputNodes[3].checked = true;
+    React.addons.TestUtils.Simulate.submit(formNode);
```
noisecapella: And here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -64,10 +225,15 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       var addTerm = function() {
         addTermCalled += 1;
       };
-
+      var reportError = function(msg) {
+        if (msg) {
+          console.error(msg);
```
noisecapella: If you're not doing anything important inside this function it should be empty

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +243,96 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        addTermCalled += 1;
+      };
+      var reportError = function(msg) {
+        errorMessage = msg;
+      };
+      var afterMount = function(component) {
+        // wait for calls to populate form
+        var node = React.findDOMNode(component);
+        var textbox = $(node).find("input")[0];
+        component.state.newTermLabel = "term2";
```
noisecapella: Use `component.setState` instead. `state` is not supposed to be mutated like this

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +243,96 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        assert.ok(component.state.newTermLabel);
```
noisecapella: There's no point in having this, the value was set in the previous line

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +243,96 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        React.addons.TestUtils.Simulate.keyUp(textbox, {key: "Enter"});
+        waitForAjax(1, function () {
+          assert.equal(addTermCalled, 0);
+          assert.equal(
+            errorMessage,
+            "Error occurred while adding new term."
```
noisecapella, line 324: Please add a comment indicating what is causing the error, for example, the mocked URL returned a 400 or something of that effect

ui/jstests/test_manage_taxonomies.jsx
noisecapella: Actually it looks like this doesn't affect testing the error, so you should just remove it.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +346,191 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          );
+        assert.equal(inputNodes.length, 4);
+        inputNodes[0].value = "TestA";
+        inputNodes[1].value = "TestA";
+        inputNodes[2].checked = true;
+        inputNodes[3].checked = false;
+        React.addons.TestUtils.Simulate.submit(formNode);
+        waitForAjax(1, function() {
+          assert.ok(
+            component.state.errorMessage
```
noisecapella: Please change this to assert the specific error message

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +346,191 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+            'input'
+          );
+        assert.equal(inputNodes.length, 4);
+        inputNodes[0].value = "TestA";
+        inputNodes[1].value = "TestA";
+        inputNodes[2].checked = true;
+        inputNodes[3].checked = false;
+        React.addons.TestUtils.Simulate.submit(formNode);
+        waitForAjax(1, function() {
+          assert.ok(
```
noisecapella: Also assert the change of state in the component

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -178,7 +178,7 @@ define('setup_manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
       }).fail(function(data) {
         var jsonData = data.responseJSON;
         var i = 0;
-        if (jsonData.non_field_errors &&
+        if (jsonData && jsonData.non_field_errors &&
           jsonData.non_field_errors.length > 0) {
```
noisecapella: You can remove this check, jsonData.non_field_errors.length is checked within the for loop

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +346,191 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          <AddVocabulary
+            vocabularies={vocabularies}
+            repoSlug="repo2"
+            updateParent={updateParent}
+            ref={afterMount}
+          />
+        );
+    }
+  );
+  QUnit.test('Assert that add term works in TaxonomyComponent',
```
noisecapella, line 751: Where do you add terms? It looks like you only add a vocabulary in this test

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,25 +69,152 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    ]
+  };
+  var listOfTerms = {
+    "count": 2,
+    "next": null,
+    "previous": null,
+    "results": vocabulary.terms
+  };
+  var saveVocabularyResponse;
+  function  addTermsComponentRenderTest(assert, rootNode) {
```
noisecapella: It looks like you are only asserting existing behavior in this function. Please rename to `assertTerms` or something similar instead, or change the function to add a term and assert the newly inserted term

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan, line 324: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/static/ui/js/manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan, line 751: Done

amir-qayyum-khan:
@noisecapella please review 

CC @pdpinch

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +32,205 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       }
     ]
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  function  assertAddTermCommon(assert, rootNode) {
+    //Common tests for Add terms, use in term and taxonomy components
```
noisecapella: You should use [JSDoc](http://usejsdoc.org/about-getting-started.html) for these kinds of comments

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +32,205 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       }
     ]
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  function  assertAddTermCommon(assert, rootNode) {
```
noisecapella: Please rename to `component` since `rootNode` has the connotation of being part of the DOM

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +32,205 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    var itemList = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        devItems,
+        'li'
+    );
+    assert.equal(itemList.length, 2);
+    return devItems;
+  }
+
+  function assertAddVocabularyCommon(assert, rootNode) {
```
noisecapella: Please rename to `component` since `rootNode` has the connotation of being part of the DOM

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +342,313 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    function(assert) {
+      assert.ok(AddVocabulary, "class object not found");
+      var vocabularies = [
+        {
+          "vocabulary": vocabulary,
+          "terms": vocabulary.terms
+        }
+      ];
+      var done = assert.async();
+      var saveVocabularyResponse;
```
noisecapella: It looks like we're not using `saveVocabularyResponse` in this function so we can remove this variable.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +240,95 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        errorMessage = msg;
+      };
+      var afterMount = function(component) {
+        // wait for calls to populate form
+        var node = React.findDOMNode(component);
+        var textbox = $(node).find("input")[0];
+        React.addons.TestUtils.Simulate.keyUp(textbox, {key: "Enter"});
+        waitForAjax(1, function () {
+          assert.equal(addTermCalled, 0);
+          assert.equal(//Mock url return status 400
```
noisecapella: Can you change this comment to 'Error is caused by a 400 status code'?

noisecapella:
I looked at the code coverage and a few cases are missing:
- Can you add a second vocabulary to the mock data? This will let us make sure that a term was added and that it was added to the correct vocabulary. (specifically manage_taxonomies.jsx line 303 which is not covered)
- `updateLearningResourceType` on `VocabularyComponent` is uncovered. I think this is because you are editing input elements directly (ie `inputNodes[2].checked = true;`) rather than using `React.addons.TestUtils.Simulate`. I think you may need to use both, not sure, but make sure that function is covered.
- `updateVocabularyType` is also uncovered

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

amir-qayyum-khan:
@noisecapella done please review

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      scryRenderedDOMComponentsWithTag(
+        devItems,
+        'li'
+    );
+    assert.equal(itemList.length, 2);
+    return devItems;
+  }
+
+  function assertAddVocabularyCommon(assert, component) {
+    /** Common tests for Add Vocabulary, use in term and taxonomy components */
```
noisecapella: JSDoc comments go before the function definition, like Javadoc comments. You also need to add information about the parameters and the return value. For example (from the JSDoc site):

```
/**
 * Represents a book.
 * @constructor
 * @param {string} title - The title of the book.
 * @param {string} author - The author of the book.
 */
function Book(title, author) {
}
```

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
       }
     ]
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  function  assertAddTermCommon(assert, component) {
+    /** Common tests for Add terms, use in term and taxonomy components */
```
noisecapella: These are not tests, they are assertions. There is nothing being changed in this function, just values being asserted. Can you change the JSDoc accordingly? Also place this above the function definition and add information about parameters (see below comment).

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      );
+    assert.ok(formNode);
+    //test form submission
+    var inputNodes = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        formNode,
+        'input'
+      );
+    assert.equal(inputNodes.length, 12);
+    inputNodes[0].value = "TestA";
```
noisecapella: This is tricky, but you need to get the React component to adjust its state and verify that it has done so. I would suggest

```
React.addons.TestUtils.Simulate.change(inputNodes[0], {target: {value: "TestA"}});
```

This will tell the React component to change the state on the name textbox, however React doesn't update the state immediately. To assert the change in state you need a callback, something like this would do:

```
var done = assert.async();
component.forceUpdate(function() {
  assert.equal(component.state.name, "TestA");
  done();
}
```

The callback will execute after `state` has been updated and you can assert changes there.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    inputNodes[1].value = "TestA";
```
noisecapella: Please add a comment saying what `inputNodes[1]` is, or store it in a variable with a descriptive name.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +247,91 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+            vocabulary={vocabulary}
+            terms={vocabulary.terms}
+            reportError={reportError}
+            addTerm={addTerm}
+            repoSlug="repo"
+            ref={afterMount}
+          />
+        );
+      // check term title
+      var listOfLinks = React.addons.TestUtils.
```
noisecapella: Please move the rest of the code in this function inside `afterMount` and use `component` instead of `vocabularyComponentRendered`. This will ensure that they take place after the component has been mounted

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -77,7 +247,91 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      var inputNode = React.addons.TestUtils.
+        findRenderedDOMComponentWithTag(
+          vocabularyComponentRendered,
+          'input'
+      );
+      React.addons.TestUtils.Simulate.change(
+        inputNode,
+        {target: {value: 'test12'}}
+      );
+      assert.equal(
```
noisecapella: You will probably need to move this assertion inside `forceUpdate()` as described above

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+      var addTermsComponentRendered = React.addons.TestUtils.
+        renderIntoDocument(
+          <AddTermsComponent
+            vocabularies={vocabularies}
+            repoSlug="repo"
+            addTerm={addTerm}
+            ref={afterMount}
+          />
+        );
+      assert.equal(
```
noisecapella: Please move this code into the `afterMount` function so we ensure it executes after the component is mounted

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+            );
+          inputNodes[11].checked = true;
+          React.addons.TestUtils.Simulate.change(
+              inputNodes[11],
+              {target: {value: 'f'}}
+            );
+          React.addons.TestUtils.Simulate.submit(formNode);
+          waitForAjax(1, function() {
+            assert.equal(
+              saveVocabularyResponse.learning_resource_types[0],
```
noisecapella: You are just asserting against the mocked AJAX here so the assertion is meaningless. The next assertion is enough to say that the AJAX call executed and `saveVocabularyResponse` was set properly

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          component.state.vocabularyType,
+          'm'
+        );
+        assert.equal(
+          component.state.learningResourceTypes.length,
+          0
+        );
+        var formNode = assertAddVocabularyCommon(assert, component);
+        waitForAjax(1, function() {
+          assert.equal(
```
noisecapella: Please add a comment indicating that `assertAddVocabularyCommon` submits the form and that you are checking that the state is cleared properly.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    //test form submission
+    var inputNodes = React.addons.TestUtils.
+      scryRenderedDOMComponentsWithTag(
+        formNode,
+        'input'
+      );
+    assert.equal(inputNodes.length, 12);
+    inputNodes[0].value = "TestA";
+    inputNodes[1].value = "TestA";
+    inputNodes[10].checked = true;
```
noisecapella: Please add a comment for what `inputNodes[10]` is or store it in a variable with a descriptive name.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,209 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    React.addons.TestUtils.Simulate.change(
+      inputNodes[10],
+      {target: {value: 'm'}}
```
noisecapella: Since the default value of `vocabularyType` is 'm' this won't do anything useful. Please set it to 'f' and then assert the 'f' value in the state.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          );
+          var inputNodes = React.addons.TestUtils.
+          scryRenderedDOMComponentsWithTag(
+            formNode,
+            'input'
+          );
+          inputNodes[0].value = "TestA";
+          inputNodes[1].value = "TestA";
+          inputNodes[2].checked = true;
+          React.addons.TestUtils.Simulate.change(
```
noisecapella: These changes should be asserted in `component.state`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          );
+        assert.ok(formNode);
+        //test form submission
+        var inputNodes = React.addons.TestUtils.
+          scryRenderedDOMComponentsWithTag(
+            formNode,
+            'input'
+          );
+        assert.equal(inputNodes.length, 4);
+        inputNodes[0].value = "TestA";
```
noisecapella: We don't need to set any of these since we're just getting mocked AJAX back anyway

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +345,348 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+              'form-horizontal'
+            );
+          assert.ok(formNode);
+          //test form submission
+          var inputNodes = React.addons.TestUtils.
+            scryRenderedDOMComponentsWithTag(
+              formNode,
+              'input'
+            );
+          inputNodes[0].value = "TestA";
```
noisecapella: Note that the result of `findRenderedDOMComponentWithClass` is actually a React component. If you wanted to deal with the DOM you need to call `inputNodes[0].getDOMNode()`

noisecapella: Actually it looks like `getDOMNode()` is deprecated according to [this](https://facebook.github.io/react/docs/top-level-api.html#react.finddomnode). So you would use `React.findDOMNode(component)` instead.

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Done

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: just to make sure fields are editable,

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: react component is ok

amir-qayyum-khan:
Done @noisecapella

noisecapella:
After looking at code coverage I see these two cases which are uncovered:
- unchecking a learningresourcetype that was previously checked
- asserting that the duplicate vocabulary error occurs when we get the appropriate validation error AJAX from the server

After tests run you can see code coverage results in coverage/lcov-report/index.html. Unfortunately coveralls is down for us at the moment so it won't show up here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  /**
+   * Common assertions for Add terms, use in term and taxonomy
+   * components tests
+   * @param {Object} assert - For assertions
+   * @param {Object} component - Root DOM object in react class
```
noisecapella: `Object` here should be `ReactComponent` instead

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+   * @return {Object} devItems - DOM object of class 'panel-body'
```
noisecapella: `Object` here should be changed to `ReactComponent` also, due to the return type for `findRenderedDOMComponentWithClass` listed in the React docs.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        'li'
+    );
+    assert.equal(itemList.length, 2);
+    return devItems;
+  }
+  /**
+   * Common assertions for Add Vocabulary, use in term and taxonomy
+   * components tests
+   * @param {Object} assert - For assertions
+   * @param {Object} component - Root DOM object in react class
```
noisecapella: Change `Object` here and the next line as described in previous comment

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
     ]
   };
-  function reportError(msg) {
-    if (msg) {
-      console.error(msg);
-    }
+  /**
+   * Common assertions for Add terms, use in term and taxonomy
+   * components tests
+   * @param {Object} assert - For assertions
```
noisecapella: `Object` here should be set to `QUnit.assert` described in [their docs](https://api.qunitjs.com/QUnit.assert/)

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+        devItems,
+        'li'
+    );
+    assert.equal(itemList.length, 2);
+    return devItems;
+  }
+  /**
+   * Common assertions for Add Vocabulary, use in term and taxonomy
+   * components tests
+   * @param {Object} assert - For assertions
```
noisecapella: `Object` here should be set to `QUnit.assert` described in [their docs](https://api.qunitjs.com/QUnit.assert/)

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    var radioManaged =  inputNodes[10];
+    var radioTagStyle = inputNodes[11];
+
+    React.addons.TestUtils.Simulate.change(
+      inputVocabularyName, {target: {value: "TestA"}}
+    );
+    React.addons.TestUtils.Simulate.change(
+      inputVocabularyDesc, {target: {value: "TestA"}}
+    );
+    radioManaged.checked = true;
```
noisecapella: You are setting the checked attribute to true here but in the next line you are telling React to set it to false. It seems like you can just remove this line since we're only asserting React's state here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    React.addons.TestUtils.Simulate.change(
+      radioManaged,
+      {target: {value: 'f'}}
+    );
+    radioTagStyle.checked = false;
```
noisecapella: This line can also be removed, see above comment

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+    var done = assert.async();
+    component.forceUpdate(function() {
+      assert.equal(component.state.name, "TestA");
```
noisecapella: This can cause a race condition in the tests because `forceUpdate` will run this piece of code asynchronously but `Simulate.submit` below will trigger an AJAX request which also runs asynchronously. This should be reorganized to put `Simulate.submit` inside the `forceUpdate` callback to ensure they happen in that order then and have `done` execute after the ajax finishes.

It would probably be easiest to inline this function since it's only used in one place. Otherwise you would need to create and return a promise to ensure proper ordering of these callbacks.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
         var textbox = $(node).find("input")[0];
+
+        // check term title
+        var listOfLinks = React.addons.TestUtils.
+          scryRenderedDOMComponentsWithTag(
+            component,
+            'a'
+        );
+        assert.equal(listOfLinks.length, 2);
+        var linkHeader = listOfLinks[0].getDOMNode();
```
noisecapella: Change this to `React.findDOMNode(listOfLinks[0])` since `getDOMNode()` is deprecated

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -35,39 +35,278 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+          findRenderedDOMComponentWithTag(
+            component,
+            'input'
+        );
+        React.addons.TestUtils.Simulate.change(
+          inputNode,
+          {target: {value: 'test12'}}
+        );
+        component.forceUpdate(function() {
+          assert.equal(
```
noisecapella, line 266: You should move the rest of the code inside this function so that things are executed in proper order and `done()` is called only after everything else completes.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +378,411 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+            inputVocabularyName, {target: {value: "TestB"}}
+          );
+          React.addons.TestUtils.Simulate.change(
+            inputVocabularyDesc, {target: {value: "TestB"}}
+          );
+          React.addons.TestUtils.Simulate.change(
+            checkboxCourse,
+            {target: {value: 'course', checked: true}}
+          );
+          radioTagStyle.checked = true;
```
noisecapella: This line can be removed since we're only asserting a change in state

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +378,411 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+              {target: {value: 'f', checked: true}}
+            );
+          component.forceUpdate(function() {
+            assert.equal(component.state.name, "TestB");
+            assert.equal(component.state.description, "TestB");
+            assert.equal(component.state.vocabularyType, "f");
+            assert.equal(
+              component.state.learningResourceTypes[0],
+              'course'
+            );
```
noisecapella: Move the rest of the code in the test inside this function so that it executes in the correct sequence

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -91,4 +378,411 @@ define(['QUnit', 'jquery', 'setup_manage_taxonomies', 'reactaddons',
...
+            component.state.vocabularies.length,
+            1
+          );
+          assert.equal(
+            component.state.vocabularies[0].terms.length,
+            2
+          );
+          assertAddTermCommon(assert, component);
+          // Adding second Vocabulary to make sure term added to correct vocabulary
+          component.addVocabulary(vocabularyWithoutTerms);
```
noisecapella, line 790: Since `addVocabulary` calls `setState` and `setState` is technically asynchronous you will need to put the rest of the code in this function inside `component.forceUpdate(function() {... })`

noisecapella:
Looks good overall, just address the code coverage comment I made and move the code around to avoid race conditions as indicated in my comments, and this should be good to go

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan, line 790: addVocabulary is not asynchronous , it just update state. See https://github.com/mitodl/lore/blob/master/ui/static/ui/js/manage_taxonomies.jsx#L293

amir-qayyum-khan:
Done @noisecapella

ui/jstests/test_manage_taxonomies.jsx
noisecapella, line 790: I thought so too, however according to the [React docs](https://facebook.github.io/react/docs/component-api.html) `setState` is asynchronous:

"setState() does not immediately mutate this.state but creates a pending state transition. Accessing this.state after calling this method can potentially return the existing value.

There is no guarantee of synchronous operation of calls to setState and calls may be batched for performance gains."

noisecapella:
Unchecking the learning resource type checkbox is still uncovered. You should be able to test this case by just unchecking it with `Simulate.change`, use `forceUpdate` to wait for state change and verify that `this.state.learningResourceTypes.length` is 0

amir-qayyum-khan:
@noisecapella which test you are referring to, please tell me. I have added this check here https://github.com/mitodl/lore/pull/328/files#diff-0f3ee7bcc78eb4497e0092ea2083d2e5R801 for TaxonomyComponent

amir-qayyum-khan:
@noisecapella also i am testing it is AddVocabulary 
https://github.com/mitodl/lore/pull/328/files#diff-0f3ee7bcc78eb4497e0092ea2083d2e5R520

noisecapella:
I see the confusion, `state.learningResourceTypes` in `TaxonomyComponent` is a list of all available learning resource types but `state.learningResourceTypes` in `AddVocabulary` is a list of the currently checked learning resource types. You need to uncheck a learning resource type checkbox then verify that it is unchecked in `state.learningResourceTypes` in `AddVocabulary`. 

You currently only check checkboxes which are previously unchecked so the code path that unchecks checkboxes is uncovered.

amir-qayyum-khan:
got it

amir-qayyum-khan:
Done @noisecapella

noisecapella:
This is great, thanks for your help!

PR #330, opened 2015-07-10
Added terms to the readonly list

noisecapella:
`learning_resources` terms are editable but because it's a related field its presence in `read_only_fields` is ignored. This shouldn't change any functionality but the change would prevent some confusion


giocalitri:
Looks fine to me: merging

PR #334, opened 2015-07-10
Switched to key property which is more self documenting

noisecapella:
React wraps events in [SyntheticEvent](https://facebook.github.io/react/docs/events.html) so this should work for all browsers


noisecapella:
Will do this again after my PR merges which makes many changes in the same file

Issue #254, opened 2015-06-29
Static: add tab in detail view for viewing list of static assets (with links)


pdpinch:
As a user, I need to be able to see which static files are associated with a learning resource and view/download them. 

Add an additional tab in the learning resource view, alongside of `XML` and `Preview` (Preview is not yet implemented). The tab will list the filenames of associated assets and link to them on S3. Links can (naively) open in a new browser window.

PR #341, opened 2015-07-12
Changes to the Django app for the Static Assets 

giocalitri:
Created view to properly serve static assets if the storage is local
Modified REST StaticAsset serializer to include the URL
Tests for media view

first part of #254

### DIFF
--- a/rest/tests/test_rest.py
+++ b/rest/tests/test_rest.py
@@ -765,11 +765,11 @@ class TestRest(RESTTestCase):
         self.get_learning_resource(repo_slug2, lr2_id)
 
     def test_filefield_serialization(self):
-        """Make sure that URL output is turned off in settings"""
+        """Make sure that URL output is turned on in settings"""
         resource = self.import_course_tarball(self.repo)
         static_assets = self.get_static_assets(
             self.repo.slug, resource.id)['results']
-        self.assertFalse(static_assets[0]['asset'].startswith("http"))
+        self.assertTrue(static_assets[0]['asset'].startswith("http"))
 
     def test_root(self):
         """
--- a/ui/tests/test_learningresources_views.py
+++ b/ui/tests/test_learningresources_views.py
@@ -4,11 +4,17 @@ Test the importer views to make sure they work.
 
 from __future__ import unicode_literals
 
+import imp
+import importlib
 import logging
+import os
 
+from django.conf import settings
+from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
 
-from learningresources.models import Repository
+import ui.urls
+from learningresources.models import Repository, StaticAsset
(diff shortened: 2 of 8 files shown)


ui/urls.py
```diff
@@ -60,3 +61,11 @@
         edit_vocabulary,
         name="edit_vocabulary"),
 ]
+
+if (settings.DEFAULT_FILE_STORAGE ==
+        'django.core.files.storage.FileSystemStorage'):
+    urlpatterns.append(
+        url(r'^media/(?P<media_path>.*)$',
```
noisecapella: Can you change this to `.+` so it doesn't match the empty string?

ui/tests/test_learningresources_views.py
```diff
@@ -237,3 +256,63 @@ def test_repo_course_filter(self):
...
+        self.assertEqual(resp.status_code, NOT_FOUND)
+        # change the default file storage to S3
+        with self.settings(
+            DEFAULT_FILE_STORAGE=('storages.backends'
+                                  '.s3boto.S3BotoStorage')
+        ):
+            # force the reload of the urls
+            try:
+                reload(ui.urls)  # pylint: disable=undefined-variable
+            # this is in case of python 3.4
```
noisecapella, line 303: Can you write a brief comment above all this explaining why python 3.3 and 3.4 are raising exceptions here?

rest/tests/test_rest.py
```diff
@@ -765,11 +765,11 @@ def test_missing_learning_resource(self):
         self.get_learning_resource(repo_slug2, lr2_id)
 
     def test_filefield_serialization(self):
```
noisecapella, line 767: Can you add an assertion somewhere (maybe in `test_rest_authorization.py` in `test_static_assets_get()`) which asserts the new `name` field?

ui/views.py
```diff
@@ -292,3 +297,25 @@ def export(request, repo_slug, resource_id):
...
+    """
+    # first check if the user has access to the file
+    file_path = os.path.join(settings.MEDIA_ROOT, media_path)
+    static_asset = get_object_or_404(StaticAsset, asset=media_path)
+    if (RepoPermission.view_repo[0] not in
+            get_perms(request.user, static_asset.course.repository)):
+        raise PermissionDenied()
+    filename = os.path.basename(file_path)
+    response = StreamingHttpResponse(
+        FileWrapper(open(file_path), 1024),
```
noisecapella: You can leave out the 1024, it looks like there is a default blocksize of 8192 and we shouldn't override that unless there's a need to

noisecapella:
Looks like everything is addressed, will merge after tests pass

noisecapella:
Actually you need to change your commit message to past tense, then I'll merge :)

PR #344, opened 2015-07-13
Added fix and test for handling deleted Elasticsearch index.

### LINKED ISSUE #283: KeyError at /repositories/<repo_slug>/ when ES index is not available any more

ShawnMilo:
Closes #283.

I tried using mock, but I got a weird URL resolving issue. I simplified it to a simple replacement of the original function, but I get the same error as I did with mock. Does anyone know what might be causing this and how to fix it?

### DIFF
--- a/ui/tests/test_importer_views.py
+++ b/ui/tests/test_importer_views.py
@@ -7,11 +7,13 @@ from __future__ import unicode_literals
 import logging
 
 from django.core.files.storage import default_storage
+import mock
 
 from learningresources.models import LearningResource, Course
 from learningresources.tests.base import LoreTestCase
 from roles.api import assign_user_to_repo_group, remove_user_from_repo_group
 from roles.api import GroupTypes
+from ui.views import RepositoryView
 
 HTTP_OK = 200
 UNAUTHORIZED = 403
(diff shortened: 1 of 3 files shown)


ui/views.py
```diff
@@ -245,16 +245,17 @@ def extra_context(self):
...
-            "repo": self.repo,
-            "perms_on_cur_repo": get_perms(self.request.user, self.repo),
-            "vocabularies": {
-                k: v
-                for k, v in context["facets"]["fields"].items()
-                if k in vocabularies
-            },
-            "qs_prefix": qs_prefix,
-        })
+        if "fields" in context["facets"]:
```
ShawnMilo, line 248: Note: Nothing changed here except for this `if` statement, and indenting the `context.update` block.

ui/tests/test_importer_views.py
```diff
@@ -159,7 +161,44 @@ def test_invalid_form(self):
...
+        # Save to restore after this test.
+        old_func = FacetedSearchView.extra_context
+
+        def extra_context(self):
+            """
+            Simulate results of deleting the index while the system
+            is running. Specifically, the "fields" key will be missing
+            from the "facets" dict.
+            """
+            results = old_func(self)
```
ShawnMilo: Note that this function, as-is, works. If we leave line 195 commented out, everything is fine and all tests pass. We just don't have coverage of the new `if` statement in `ui/views.py`.

noisecapella:
I did some investigating, it looks like the reverse is failing because it's expecting the `"repo"` key in the context, which is no longer added since `"fields"` is now empty

ShawnMilo:
Okay, cool. So, the `KeyError` was handled, but that just let it break further down the line. So it seems that it's not a good idea to test the view. Fortunately, @cg added a better test.

carsongee:
:+1: once tests pass

PR #352, opened 2015-07-13
Implemented preview link for learning resource panel

noisecapella:


importer/api/__init__.py
```diff
@@ -141,7 +143,7 @@ def import_children(course, element, parent):
         course=course, parent=parent, resource_type=element.tag,
         title=element.attrib.get("display_name", "MISSING"),
         content_xml=etree.tostring(element),
-        mpath=mpath,
+        mpath=mpath, url_name=element.attrib.get("url_name", None),
```
giocalitri: Please, one element per line

learningresources/api.py
```diff
@@ -41,19 +41,21 @@ class NotFound(LearningResourceException):
     pass
 
 
+# pylint: too-many-arguments
```
giocalitri: i guess you are missing the `disable=` here, but probably if you put at this level it will disable it for all the functions

learningresources/models.py
```diff
@@ -129,6 +130,30 @@ class LearningResource(BaseModel):
     xa_nr_attempts = models.IntegerField(default=0)
     xa_avg_grade = models.FloatField(default=0)
     xa_histogram_grade = models.FloatField(default=0)
+    url_name = models.TextField(null=True)
+
+    @property
```
giocalitri: why are you using `@property`?

noisecapella: It seemed like a good idea at the time but I can't remember why. I'll change it to a getter for clarity

importer/api/__init__.py
```diff
@@ -92,7 +92,8 @@ def import_course_from_path(path, repo_id, user_id):
     Returns:
         course (learningresources.Course)
     """
-    bundle = XBundle()
+    bundle = XBundle(
+        keep_urls=True, keep_studio_urls=True, preserve_url_name=True)
```
carsongee: nit: prefer end paren to be on new line if newline character is first character after the opening paren

importer/api/__init__.py
```diff
@@ -130,9 +132,9 @@ def import_children(course, element, parent):
     of an XML tree.
 
     Args:
-        course (learningresources.Course): Course
+        course (Course): Course
```
carsongee: this should be `learningresources.models.Course` since `Course` isn't imported here

importer/api/__init__.py
```diff
@@ -130,9 +132,9 @@ def import_children(course, element, parent):
...
         element (lxml.etree): XML element within xbundle
-        parent (learningresources.LearningResource): Parent LearningResource
+        parent (LearningResource): Parent LearningResource
```
carsongee: Same here, for sphinx, should be `learningresources.models.LearningResource` to properly link in the api docs (I think anyway)

learningresources/api.py
```diff
@@ -75,20 +78,21 @@ def create_course(org, repo_id, course_number, run, user_id):
 
 
 # pylint: disable=too-many-arguments
-def create_resource(course, parent, resource_type, title, content_xml, mpath):
+def create_resource(
+        course, parent, resource_type, title, content_xml, mpath, url_name):
```
carsongee: same paren nit

lore/settings.py
```diff
@@ -219,6 +219,11 @@ def get_var(name, default):
     DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
 
 
+# Lore preview settings
+LORE_PREVIEW_BASE_URL = get_var(
+    'LORE_PREVIEW_BASE_URL', 'http://staging.mitx.mit.edu/courses/')
```
carsongee: change this to `https://example.com` or `https://www.sandbox.edx.org/. Also, the`courses` part will never change by configuration (it will always be courses no matter what server), so it isn't needed here and can be added in the URL construction.

rest/serializers.py
```diff
@@ -206,6 +207,7 @@ class Meta:
             'xa_avg_grade',
             'xa_histogram_grade',
             'terms',
+            'preview_url',
```
carsongee, line 210: don't we need to update the apiary docs too, or do we not specify fields?

learningresources/models.py
```diff
@@ -129,6 +130,29 @@ class LearningResource(BaseModel):
...
+                          '/{run}/jump_to_id/{preview_id}')
+            return url_format.format(
+                base_url=LORE_PREVIEW_BASE_URL,
+                org=self.course.org,
+                course_number=self.course.course_number,
+                run=self.course.run,
+                preview_id=self.url_name,
+            )
+        else:
+            url_format = ('{base_url}{org}/{course_number}'
```
carsongee: I think we can leave this as is for now, but this first part `{org}/{course_number}/{run}/` breaks with SplitMongoModuleStore, and I don't think there is a way (without an API call), to find the right URL for a split course.  Do you happen to know @bdero if using the opaque key repo may help here to create a `CourseKey` from the course info?

rest/tests/test_rest.py
```diff
@@ -928,3 +931,29 @@ def test_vocabulary_learning_resource_types(self):
...
                 self.repo.slug, vocab_slug)['learning_resource_types'])
+
+    def test_preview_url(self):
+        """
+        Assert preview url behavior for learning resources
+        """
+        learning_resource = LearningResource.objects.first()
+        expected_jump_to_id_url = (
+            "http://staging.mitx.mit.edu/courses/"
+            "test org/infinity/Febtober/jump_to_id/url_name1"
```
carsongee: shouldn't this be URL encoded (i.e. the space in org should be %20)?

rest/tests/test_rest.py
```diff
@@ -928,3 +931,29 @@ def test_vocabulary_learning_resource_types(self):
...
+
+        resource_dict = self.get_learning_resource(
+            self.repo.slug, learning_resource.id)
+        self.assertEqual(
+            expected_jump_to_id_url, resource_dict['preview_url'])
+
+        learning_resource.url_name = None
+        self.assertEqual(
+            "http://staging.mitx.mit.edu/courses/"
+            "test org/infinity/Febtober/courseware",
```
carsongee: same, URL encoding

learningresources/models.py
```diff
@@ -129,6 +130,29 @@ class LearningResource(BaseModel):
     xa_nr_attempts = models.IntegerField(default=0)
     xa_avg_grade = models.FloatField(default=0)
     xa_histogram_grade = models.FloatField(default=0)
+    url_name = models.TextField(null=True)
+
+    def get_preview_url(self):
+        """Create a preview URL"""
```
carsongee: Should have a period ala https://www.python.org/dev/peps/pep-0257/

learningresources/models.py
carsongee: I don't think we can tell if it is split or not, but I do think we should probably use https://github.com/edx/opaque-keys instead of manually formatting the string here.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -88,6 +88,9 @@ define('learning_resources', [
                className="btn btn-white"
                onClick={this.selectXml}>Select XML</a>
           </p>
+          <p className="text-right">
+            <a href={this.state.previewUrl} target="_blank">Preview</a>
```
carsongee: consider making this a button instead of just a textual link with class `.btn-primary`.  I also think it would be nice to have it float to the left of the `Select XML` button instead of below it.

learningresources/models.py
```diff
@@ -5,6 +5,14 @@
 from __future__ import unicode_literals
 
 import logging
+try:
```
carsongee: Replace this with [six.moves.urllib.parse](https://pythonhosted.org/six/#module-six.moves.urllib.parse), I think we should have it from Django, but it isn't explicit in our requirements file, so maybe add it there too.

learningresources/models.py
```diff
@@ -129,6 +139,34 @@ class LearningResource(BaseModel):
     xa_nr_attempts = models.IntegerField(default=0)
     xa_avg_grade = models.FloatField(default=0)
     xa_histogram_grade = models.FloatField(default=0)
+    url_name = models.TextField(null=True)
+
+    def get_preview_url(self):
+        """Create a preview URL."""
+        key = str(CourseLocator(
```
carsongee: Do you need the `str` here? I thought it serialized to a string automagically

noisecapella:
Apparently opaque-keys doesn't support Python 3 currently. I filed a ticket: https://openedx.atlassian.net/browse/PLAT-752

Will switch back to string concat

carsongee:
Cool, :rooster: once tests pass and @pwilkins approves

pwilkins:
I'm able to preview learning resources.  Clicking the Preview button in the LR detail panel opens a new window containing a live view of the LR in the course.  I checked all LR types.

PR #353, opened 2015-07-14
Added panel for static assets

### LINKED ISSUE #254: Static: add tab in detail view for viewing list of static assets (with links)

giocalitri:
Closes #254

### DIFF
--- a/ui/jstests/test-static-assets.jsx
+++ b/ui/jstests/test-static-assets.jsx
@@ -1,0 +1,118 @@
+define(['QUnit', 'jquery', 'static_assets', 'reactaddons',
+  'test_utils'], function(
+  QUnit, $, StaticAssets, React, TestUtils) {
+    'use strict';
+
+    var StaticAssetsPanel = StaticAssets.StaticAssetsPanel;
+    var waitForAjax = TestUtils.waitForAjax;
+
+    var staticAssetsResponse = {
+      "count": 2,
+      "next": null,
+      "previous": null,
+      "results": [
+        {
+          "id": 1,
+          "asset": "http://lore.test.mit/media/assets/edX/toy/" +
+            "TT_2012_Fall/subs_CCxmtcICYNc.srt_dZF2PcC.sjson",
+          "name": "subs_CCxmtcICYNc.srt_dZF2PcC.sjson"
+        },
+        {
+          "id": 2,
+          "asset": "http://lore.test.mit/media/assets/edX/toy/" +
+            "TT_2012_Fall/test_YpcBO8Y.txt",
+          "name": "test_YpcBO8Y.txt"
+        }
+      ]
+    };
+
+    QUnit.module('Test static assets panel', {
+      beforeEach: function() {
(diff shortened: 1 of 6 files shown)


ui/jstests/test-static-assets.jsx
```diff
@@ -0,0 +1,159 @@
+define(['QUnit', 'jquery', 'static_assets', 'reactaddons',
+  'test_utils', 'jquery_mockjax'], function(
```
noisecapella: I don't think you need `jquery_mockjax` here, we do all that through `TestUtils` now. I should remove it from `test-learning-resource.jsx` too

ui/jstests/test-static-assets.jsx
```diff
@@ -0,0 +1,159 @@
...
+        };
+        React.addons.TestUtils.renderIntoDocument(<StaticAssetsPanel
+          repoSlug="repo"
+          learningResourceId="1"
+          ref={afterMount} />);
+      }
+    );
+
+    QUnit.test(
+      'Assert that StaticAssetsPanel list contains links',
```
noisecapella: Can we merge this test with the previous one? It looks like there is just one extra assert

ui/jstests/test-static-assets.jsx
```diff
@@ -0,0 +1,159 @@
...
+        };
+        React.addons.TestUtils.renderIntoDocument(<StaticAssetsPanel
+          repoSlug="repo"
+          learningResourceId="1"
+          ref={afterMount} />);
+      }
+    );
+
+    QUnit.test(
+      'Assert that StaticAssetsPanel links are formatted properly',
```
noisecapella: This can be merged with the other two tests since it only has extra assertions

ui/static/ui/js/static_assets.jsx
```diff
@@ -0,0 +1,73 @@
...
+          "/learning_resources/" +
+          this.props.learningResourceId + "/static_assets/")
+          .done(function (data) {
+            if (!thiz.isMounted()) {
+              // In time AJAX call happens component may become unmounted
+              return;
+            }
+
+            thiz.setState({
+              data: data
```
noisecapella: Can you use data.results instead? It would make the state a little bit cleaner.

noisecapella:
I'm curious what we would want to put if the static assets list was blank, but that's not a blocker for this PR. Functionality looks good

giocalitri:
No idea, but I guess we will add other functionalities to this panel, so it will not look empty, eventually

pdpinch:
What @giocalitri said is correct. At some point we will add a function in this panel for the user to add missing static assets from the ones that were imported.

bdero:
v0.4.0 code freeze lifted, merging.

PR #360, opened 2015-07-15
Added static file parsing to HTML elements.

### LINKED ISSUE #231: static assets: parse problem, html blocks for javascript, CSS, image links
As a curator, when I import a course I want to have static asset files (in the /static folder) in my course associated with appropriate learning resources. 

We need to parse problem, and html learning resources for links to javascript, CSS, image and iframe files. For example:

`
<script src="">
<link href="">

<iframe src="">
`

where the src/href attribute starts with `/static/`

ShawnMilo:
Closes #231.

Code review suggestions:
1. Read `importer/api/__init__.py` first. It is where the new associations are created.

Everything else is new course material to support the tests, and test updates.


importer/api/__init__.py
```diff
@@ -4,11 +4,13 @@
 
 from __future__ import unicode_literals
 
+from bs4 import BeautifulSoup
 from shutil import rmtree
 import logging
 from tempfile import mkdtemp
 from os.path import join, exists, isdir
 from os import listdir
+from pprint import pformat
```
giocalitri: I do not think there is a proper guideline about the sequence of imports, but we should put the `from ...` all together (maybe after the `import ...`) in the same section (in this case standard libraries)

importer/api/__init__.py
```diff
@@ -142,6 +144,7 @@ def import_children(course, element, parent):
         content_xml=etree.tostring(element),
         mpath=mpath,
     )
+    target = "/static/"
```
giocalitri, line 146: why is this hardcoded?

importer/api/__init__.py
ShawnMilo: `pformat` was used during development and should be deleted. I don't know why it didn't trigger an "unused import" error. I'll take it out.

importer/api/__init__.py
```diff
@@ -151,6 +154,23 @@ def import_children(course, element, parent):
...
+                        try:
+                            asset = StaticAsset.objects.get(
+                                course__id=resource.course_id,
+                                asset=course_asset_basepath(course, path),
+                            )
+                            resource.static_assets.add(asset)
+                        except StaticAsset.DoesNotExist:
+                            continue
+                except AttributeError:
+                    continue  # not a string
```
giocalitri, line 179: Could you please explain briefly at the beginning of this section what is the logic you implement?

importer/api/__init__.py
ShawnMilo, line 146: It needs to be used twice; once to check for it, and again to strip it off of the path.

Before, it was like this:

``` python
if val.startswith("/static/"):
    # more code here
    asset=course_asset_basepath(course, val[8:])
```

I thought it was nicer to make it explicit as shown below (`val[len(target)]`). This makes it clear that the `path` stripping is based on the length of `/static/`.

importer/tests/test_import.py
```diff
@@ -258,13 +259,9 @@ def get_counts():
         )
         # There should be something.
         counts = get_counts()
-        self.assertTrue(counts.resources == 5)
+        self.assertTrue(counts.resources == self.toy_resource_count)
```
giocalitri: use `self.assertEqual`

importer/tests/test_import.py
```diff
@@ -258,13 +259,9 @@ def get_counts():
...
         self.assertTrue(counts.videos == 2)
-        self.assertTrue(counts.assets == 3)
-        # Only one video in the course has subtitles.
-        self.assertTrue(
-            StaticAsset.objects.filter(
-                learningresource__course__id__isnull=False).count() == 1)
+        self.assertTrue(counts.assets == self.toy_asset_count)
```
giocalitri: use `self.assertEqual`

learningresources/tests/testdata/courses/toy/static/webGLDemo.css
```diff
@@ -0,0 +1,5 @@
+#container {
+  background-color: black;
+  width: 400px;
+  height:400px;
+}
```
giocalitri: empty line missing at the end

importer/api/__init__.py
giocalitri, line 146: Your approach is right, but I am wandering if this should become a module constant

ShawnMilo, line 146: I don't have a problem with it either way. If you feel strongly about it I'll make that change.

rest/tests/test_rest_authorization.py
```diff
@@ -750,19 +754,30 @@ def test_resource_put_patch(self):
...
-        resource1 = self.import_course_tarball(self.repo)
+        def get_resource_with_asset(type_name):
+            """
+            Get a LearningResource with a StaticAsset.
+            """
+            for resource in get_resources(self.repo.id):
+                if resource.learning_resource_type.name != type_name:
+                    continue
+                if resource.static_assets.count() == 0:
+                    continue
```
giocalitri, line 764: I think you can merge the two if in one with an `or` between the conditions

importer/api/__init__.py
giocalitri, line 146: no, leave it there as it is: if it starts to be an issue we will change it later

ShawnMilo:
For your convenience, here is the diff for my last rebase based on your comments:

``` diff
diff --git importer/api/__init__.py importer/api/__init__.py
index fe2e7a5..cfdc38b 100644
--- importer/api/__init__.py
+++ importer/api/__init__.py
@@ -10,7 +10,6 @@ import logging
 from tempfile import mkdtemp
 from os.path import join, exists, isdir
 from os import listdir
-from pprint import pformat

 from archive import Archive, ArchiveException
 from django.core.files import File
@@ -155,6 +154,13 @@ def import_children(course, element, parent):
             for asset in assets:
                 resource.static_assets.add(asset)
     else:
+        # Recursively find all sub-elements, looking for anything which
+        # refers to /static/. Then make the association between the
+        # LearningResource and StaticAsset if the StaticAsset exists.
+        # This is like doing soup.findAll("a") and checking for whether
+        # "/static/" is in the href, which would work but also requires
+        # more code to check for link, img, iframe, script, and others,
+        # and within those, check for href or src existing.
         soup = BeautifulSoup(etree.tostring(element), 'lxml')
         for child in soup.findAll():
             for _, val in child.attrs.items():
diff --git importer/tests/test_import.py importer/tests/test_import.py
index 26047a2..ca716fb 100644
--- importer/tests/test_import.py
+++ importer/tests/test_import.py
@@ -249,9 +249,9 @@ class TestImportToy(LoreTestCase):

         # There should be nothing.
         counts = get_counts()
-        self.assertTrue(counts.resources == 0)
-        self.assertTrue(counts.videos == 0)
-        self.assertTrue(counts.assets == 0)
+        self.assertEqual(counts.resources, 0)
+        self.assertEqual(counts.videos, 0)
+        self.assertEqual(counts.assets, 0)
         # Import the course.
         import_course_from_file(
             self.get_course_single_tarball(),
@@ -259,9 +259,9 @@ class TestImportToy(LoreTestCase):
         )
         # There should be something.
         counts = get_counts()
-        self.assertTrue(counts.resources == self.toy_resource_count)
-        self.assertTrue(counts.videos == 2)
-        self.assertTrue(counts.assets == self.toy_asset_count)
+        self.assertEqual(counts.resources, self.toy_resource_count)
+        self.assertEqual(counts.videos, 2)
+        self.assertEqual(counts.assets, self.toy_asset_count)

         # There should be a single static asset.
         course = Course.objects.all().order_by("-id")[0]  # latest course
diff --git learningresources/tests/testdata/courses/toy/static/webGLDemo.css learningresources/tests/testdata/courses/toy/static/webGLDemo.css
index 0ba8df3..7bdda4e 100644
--- learningresources/tests/testdata/courses/toy/static/webGLDemo.css
+++ learningresources/tests/testdata/courses/toy/static/webGLDemo.css
@@ -2,4 +2,5 @@
   background-color: black;
   width: 400px;
   height:400px;
-}
\ No newline at end of file
+}
+
```

rest/tests/test_rest_authorization.py
ShawnMilo, line 764: Yeah, but then the line gets long. I prefer it the way it is. I think they're pretty much equal.

importer/api/__init__.py
ShawnMilo, line 179: Added. Just putting this comment here because this comment wasn't hidden by the change.

giocalitri:
there is just my last comment left

rest/tests/test_rest_authorization.py
giocalitri, line 764: you can have an `if` on multiple lines (using parenthesis)
I agree the functionality is the same, but two `if` statements with the same behavior should be merged

ShawnMilo:
``` patch
diff --git rest/tests/test_rest_authorization.py rest/tests/test_rest_authorization.py
index 1f217c3..9231912 100644
--- rest/tests/test_rest_authorization.py
+++ rest/tests/test_rest_authorization.py
@@ -759,9 +759,8 @@ class TestRestAuthorization(RESTTestCase):
             Get a LearningResource with a StaticAsset.
             """
             for resource in get_resources(self.repo.id):
-                if resource.learning_resource_type.name != type_name:
-                    continue
-                if resource.static_assets.count() == 0:
+                if (resource.learning_resource_type.name != type_name or
+                        resource.static_assets.count() == 0):
                     continue
                 return resource
         self.import_course_tarball(self.repo)
```

giocalitri:
seen: I'm rebuilding docker to test it

ShawnMilo:
Pushed with new test for the CSS and image in the HTML block.

giocalitri:
ok, retesting functionality

giocalitri:
everything looks good: will merge after tests pass

giocalitri:
build is red, please fix

ShawnMilo:
Fixed.

PR #361, opened 2015-07-15
Added HTML reporting of javascript tests

noisecapella:
Outputs `jstest_results.html` so we can see what failed without trial and error


giocalitri:
looks good to me: merging after everything is green

PR #362, opened 2015-07-15
Removed old duplicate version of qunit

noisecapella:
We're actually using `qunitjs` in the tests, `qunit` made it in there by accident


giocalitri:
looks good: merging

PR #363, opened 2015-07-15
Set QUnit timeout to fix test error reporting

noisecapella:
Karma times out after 10 seconds, this sets QUnit's per-test timeout to 5 seconds. When Karma times out it doesn't provide any useful information about where things went wrong. Since all tests are importing TestUtils we can set a smaller timeout for QUnit so that it will definitely fail the test before Karma's timeout can occur. This makes it so that all our tests provide pass or fail results in the test output.


giocalitri:
looks good: merging

PR #365, opened 2015-07-15
Added function to obtain collections

### LINKED ISSUE #347: Collections retrieved through API are limited to 20 items
The API limits all collections to 20 items with a link to the next 20 items, etc. Currently we do not retrieve the next 20 items for any API call in the UI. We can change this on a case by case basis, maybe move all AJAX calls into utility functions or install a model like Backbone to handle this automatically.

noisecapella:
`Utils.getCollections` will follow the `next` links that are provided with collections in our API so that we don't stop at the first 20 items

Fixes #347


noisecapella:
Fixes #347

giocalitri:
looks good to me: merging

giocalitri:
could you please put the `Fixes #347` message in the description of the PR?

noisecapella:
Fixes #347

PR #366, opened 2015-07-15
Removed unused piece of state

noisecapella:
`terms` was moved to `this.props` but I missed this instance of `terms` on the state. It is not used anywhere so removing shouldn't cause a problem.


giocalitri:
looks good: merging

PR #369, opened 2015-07-15
Removed vocabulary forms since we are doing this via REST API and React instead

### LINKED ISSUE #255: Editing taxonomies while keeping the same name causes an IntegrityError
In the form at `/repositories/repo_slug/vocabularies/vocab_slug/`, editing a vocabulary but not changing its name causes a unique constraint failure

noisecapella:
Fixes #255


carsongee:
:+1:

PR #370, opened 2015-07-16
Fixed mutating of this.state which is forbidden

noisecapella:
Since `this.setState` may act asynchronously we can't ever alter `this.state` directly.

### DIFF
--- a/ui/static/ui/js/manage_taxonomies.jsx
+++ b/ui/static/ui/js/manage_taxonomies.jsx
@@ -289,8 +289,7 @@ define('setup_manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
         vocabulary: vocab
       };
       var vocabularies = this.state.vocabularies;
-      vocabularies.push(newVocab);
-      this.setState({vocabularies: vocabularies});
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -289,8 +289,7 @@ define('setup_manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
         vocabulary: vocab
       };
       var vocabularies = this.state.vocabularies;
-      vocabularies.push(newVocab);
-      this.setState({vocabularies: vocabularies});
+      this.setState({vocabularies: vocabularies.concat([newVocab])});
```
carsongee, line 292: I guess I didn't realize this was mutating the state, since I'm calling set state anyway, how is this "forbidden"? (I'm just honestly curious why this is wrong)

noisecapella, line 292: It's discouraged in [their documentation](https://facebook.github.io/react/docs/component-api.html). It may give other pieces of code (like unit tests) a false impression on what the state is since setState may have been batched for later execution, but the change here is happening immediately.

PR #371, opened 2015-07-16
Added shopping cart for export

### LINKED ISSUE #357: Export: user can add resources to a shopping cart for export
Shopping carts contents are per-user and can be stored in a cookie.

noisecapella:
Fixes #357

### DIFF
--- a/rest/permissions.py
+++ b/rest/permissions.py
@@ -200,3 +200,16 @@ class ViewStaticAssetPermission(ViewLearningResourcePermission):
             raise Http404()
 
         return True
+
+
+class ViewLearningResourceExportPermission(BasePermission):
+    """
+    Checks that username mentioned is the same one as the request.
+    In the future we may want to let users see each other's exports
+    but for now this just enforces the user seeing their own exports.
+    """
+
+    def has_permission(self, request, view):
+        username = view.kwargs['username']
+
+        return request.user.username == username
--- a/rest/tests/test_learning_resource.py
+++ b/rest/tests/test_learning_resource.py
@@ -1,5 +1,5 @@
 """
-REST tests for learning resources and static assets
+REST tests for LearningResources and static assets
 """
 
 from __future__ import unicode_literals
@@ -24,15 +24,17 @@ from learningresources.api import get_resources
 from learningresources.models import (
     LearningResource,
     LearningResourceType,
+    Repository,
 )
 from importer.tasks import import_file
 from taxonomy.models import Vocabulary
+from roles.permissions import GroupTypes
+from roles.api import assign_user_to_repo_group
 
 
-# pylint: disable=invalid-name
 class TestLearningResource(RESTTestCase):
     """
-    REST tests for learning resources and static assets
+    REST tests for LearningResources and StaticAssets
     """
 
     def test_immutable_fields_learning_resource(self):
@@ -81,7 +83,7 @@ class TestLearningResource(RESTTestCase):
                                        skip_assert=True))
 
     def test_missing_learning_resource(self):
-        """Test for an invalid learning resource id"""
+        """Test for an invalid LearningResource id."""
         repo_slug1 = self.repo.slug
         resource1 = self.import_course_tarball(self.repo)
         lr1_id = resource1.id
@@ -104,8 +106,37 @@ class TestLearningResource(RESTTestCase):
                                    expected_status=HTTP_404_NOT_FOUND)
         self.get_learning_resource(repo_slug2, lr2_id)
 
+    def test_learning_resource_filter(self):
+        """Test for filtering LearningResources by ids."""
+
+        def get_filtered(ids, expected_status=HTTP_200_OK):
+            """Return list of LearningResources in shopping cart."""
+            url_base = "{repo_base}{repo_slug}/learning_resources/".format(
+                repo_base=REPO_BASE,
+                repo_slug=self.repo.slug,
+            )
+            resp = self.client.get("{url_base}?id={ids}".format(
+                url_base=url_base,
+                ids=",".join([str(s) for s in ids])
+            ))
+            self.assertEqual(expected_status, resp.status_code)
+            if expected_status == HTTP_200_OK:
+                return sorted([x['id'] for x in as_json(resp)['results']])
+
+        self.import_course_tarball(self.repo)
+        self.assertEqual([], get_filtered([]))
+        get_filtered(["not-a-number"],
+                     expected_status=HTTP_400_BAD_REQUEST)
+        self.assertEqual([], get_filtered([-1]))
+
+        all_ids = list(get_resources(
+            self.repo.id).values_list('id', flat=True))
+        self.assertEqual(sorted(all_ids), get_filtered(all_ids))
+        self.assertEqual(
+            sorted(all_ids[:5]), get_filtered(all_ids[:5]))
+
     def test_filefield_serialization(self):
-        """Make sure that URL output is turned on in settings"""
+        """Make sure that URL output is turned on in settings."""
         resource = self.import_course_tarball(self.repo)
         static_assets = self.get_static_assets(
             self.repo.slug, resource.id)['results']
@@ -113,7 +144,7 @@ class TestLearningResource(RESTTestCase):
 
     def test_add_term_to_learning_resource(self):
         """
-        Add a term to a learning resource via PATCH
+        Add a term to a LearningResource via PATCH.
         """
 
         resource = self.import_course_tarball(self.repo)
@@ -149,7 +180,7 @@ class TestLearningResource(RESTTestCase):
 
     def test_learning_resource_types(self):
         """
-        Get from learning_resource_types
+        Get from learning_resource_types.
         """
         base_url = "{}learning_resource_types/".format(API_BASE)
 
@@ -190,7 +221,7 @@ class TestLearningResource(RESTTestCase):
 
     def test_preview_url(self):
         """
-        Assert preview url behavior for learning resources
+        Assert preview url behavior for LearningResources.
         """
         learning_resource = LearningResource.objects.first()
         expected_jump_to_id_url = (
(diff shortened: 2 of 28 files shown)


noisecapella:
This PR looks brutal but most of this is just tests that were moved around. You only need to pay attention to tests which have `learning_resource_exports` somewhere in the title, and to the related utility functions `get_learning_resource_exports`, etc.

pwilkins:
All the functionality of the UI works as described.  (The ability to export the selected learning resources is not yet implemented.)

pwilkins:
There is an issue with the Apiary Blueprint document (`apiary.apib`) that, while not related to this PR, needs fixing.  The second line of the file:

`HOST: http://lore.odl.mit.edu/`

is incomplete.  It should be:

`HOST: http://lore.odl.mit.edu/api/v1/`

This is needed for the Apiary web app to generate the correct API endpoints.

rest/tests/test_learning_resource.py
```diff
@@ -104,6 +107,34 @@ def test_missing_learning_resource(self):
                                    expected_status=HTTP_404_NOT_FOUND)
         self.get_learning_resource(repo_slug2, lr2_id)
 
+    def test_learning_resource_filter(self):
+        """Test for filtering learning resources by ids"""
+
+        def get_filtered(ids, expected_status=HTTP_200_OK):
+            """Return list of learning resourceswhich are in shopping cart"""
+            url_base = "{repo_base}{repo_slug}/learning_resources/".format(
```
pwilkins: Does ``{repo_base}`` always have a trailing slash?  I presume it does.

rest/views.py
```diff
@@ -465,3 +482,109 @@ def get_queryset(self):
...
+    serializer_class = LearningResourceExportSerializer
+    permission_classes = (
+        ViewLearningResourcePermission,
+        IsAuthenticated,
+    )
+    lookup_field = 'id'
+    lookup_url_kwarg = 'lr_id'
+
+    def get_object(self):
+        try:
```
pwilkins, line 565: does this need a docstring?

rest/views.py
```diff
@@ -465,3 +482,109 @@ def get_queryset(self):
...
+        except KeyError:
+            raise Http404
+
+        if lr_id not in exports:
+            raise Http404
+
+        return {"id": lr_id}
+
+    def perform_destroy(self, instance):
+        lr_id = instance['id']
```
pwilkins, line 586: does this need a docstring?

ui/static/ui/js/lr_exports.jsx
```diff
@@ -0,0 +1,77 @@
+define("lr_exports",
+  ['reactaddons', 'jquery', 'lodash', 'utils'], function (React, $, _, Utils) {
+  'use strict';
+
+  var ExportsComponent = React.createClass({
+    componentDidMount: function() {
```
pwilkins, line 10: I wonder if there is a way to document ReactJS component classes that doesn't bloat the code that goes over the wire?  It doesn't look like JSX strips comments when it compiles to javascript.

noisecapella, line 10: I believe we do minimize our javascript using compressor when it's in production, which should strip out all comments. You can verify this on lore-ci

pwilkins:
@noisecapella I am prepared to merge this unless you want a more rigorous code review.

noisecapella:
I think a rigorous code review is important here given the size of the PR. Any volunteers?

rest/views.py
noisecapella, line 586: Pylint doesn't require it since it overrides a method but in this case it's probably a good idea anyway

ShawnMilo:
General comment: `LearningResource` is often written as "learning resource." I think we've tried to standardize on the former. Here is a patch you can apply if you like:

http://hastebin.com/behodanifi

rest/tests/test_learning_resource.py
```diff
@@ -214,6 +245,71 @@ def test_preview_url(self):
...
+        self.get_learning_resource_export(
+            self.repo.slug, lr_id, expected_status=HTTP_404_NOT_FOUND)
+        self.create_learning_resource_export(
+            self.repo.slug, {"id": lr_id},
+            expected_status=HTTP_400_BAD_REQUEST)
+        self.delete_learning_resource_export(
+            self.repo.slug, lr_id, expected_status=HTTP_404_NOT_FOUND)
+
 
 # pylint: disable=too-many-ancestors
```
ShawnMilo: We disable `too-many-ancestors` and `invalid-name` a fair number of times. I think we should disable them globally, because they're not the kind of things we tend to violate out of laziness or oversight, like line length and missing docstrings. Opinions, @carsongee, @pdpinch?

ShawnMilo:
General comment: Lots of docstrings are sentences that are missing the period at the end.

rest/tests/test_learning_resource.py
```diff
@@ -104,6 +107,34 @@ def test_missing_learning_resource(self):
                                    expected_status=HTTP_404_NOT_FOUND)
         self.get_learning_resource(repo_slug2, lr2_id)
 
+    def test_learning_resource_filter(self):
+        """Test for filtering learning resources by ids"""
+
+        def get_filtered(ids, expected_status=HTTP_200_OK):
+            """Return list of learning resourceswhich are in shopping cart"""
```
ShawnMilo: Typo: "resourceswhich" 

Also, maybe sentence should just be:

"Return list of LearningResources in shopping cart."

rest/tests/test_learning_resource.py
```diff
@@ -104,6 +107,34 @@ def test_missing_learning_resource(self):
...
+            if expected_status == HTTP_200_OK:
+                return [x['id'] for x in as_json(resp)['results']]
+
+        self.import_course_tarball(self.repo)
+        self.assertEqual([], get_filtered([]))
+        get_filtered(["not-a-number"],
+                     expected_status=HTTP_400_BAD_REQUEST)
+        self.assertEqual([], get_filtered([-1]))
+
+        all_ids = [r.id for r in get_resources(self.repo.id).all()]
```
ShawnMilo: This might be better as:

``` python
all_ids = list(get_resources(self.repo.id).values_list('id', flat=True))
```

rest/tests/test_learning_resource.py
```diff
@@ -104,6 +107,34 @@ def test_missing_learning_resource(self):
...
+                repo_base=REPO_BASE,
+                repo_slug=self.repo.slug,
+            )
+            resp = self.client.get("{url_base}?id={ids}".format(
+                url_base=url_base,
+                ids=",".join([str(s) for s in ids])
+            ))
+            self.assertEqual(expected_status, resp.status_code)
+            if expected_status == HTTP_200_OK:
+                return [x['id'] for x in as_json(resp)['results']]
```
ShawnMilo: Please return this list comprehension `sorted()` -- since it's being sorted multiple times in comparisons below.

rest/tests/test_learning_resource.py
```diff
@@ -451,3 +547,123 @@ def test_static_assets_delete(self):
             self.DEFAULT_LR_DICT
         )
         self.assertEqual(resp.status_code, HTTP_405_METHOD_NOT_ALLOWED)
+
+    def test_learning_resource_exports(self):
+        """Test for creating learning resource ids in session shopping cart"""
+
+        # setup our data
```
ShawnMilo: "setup" is a noun, "set up" is a verb. This typo is duplicated elsewhere in this diff. Please update.

rest/tests/test_learning_resource.py
```diff
@@ -451,3 +547,123 @@ def test_static_assets_delete(self):
...
+        repo_slug1 = self.repo.slug
+        repo_slug2 = self.create_repository()['slug']
+        repo2 = Repository.objects.get(slug=repo_slug2)
+
+        # 'add_repo_user' will be able to see both repos
+        assign_user_to_repo_group(
+            self.add_repo_user, self.repo, GroupTypes.REPO_ADMINISTRATOR)
+        self.import_course_tarball(repo2)
+
+        repo1_lrid1 = get_resources(self.repo.id).all()[0].id
```
ShawnMilo, line 575: This is a nit. These two lines can use `.first()` instead of `.all()[0]`. Given the `.all()[1]` below an argument could be made for uniformity. Feel free to ignore this if you think it's more readable this way.

rest/tests/test_learning_resource.py
```diff
@@ -451,3 +547,123 @@ def test_static_assets_delete(self):
...
+            repo_slug1, repo2_lrid1,
+            expected_status=HTTP_404_NOT_FOUND)
+        self.get_learning_resource_export(
+            repo_slug2, repo1_lrid1,
+            expected_status=HTTP_404_NOT_FOUND)
+        self.assertEqual(repo2_lrid1, self.get_learning_resource_export(
+            repo_slug2, repo2_lrid1)['id'])
+
+        # repo1_lrid1 belongs to repo1, can't add it to repo2
+        self.create_learning_resource_export(repo_slug2, {
```
ShawnMilo: This test is a duplicate of the test above on line 621.

rest/tests/test_learning_resource.py
```diff
@@ -451,3 +547,123 @@ def test_static_assets_delete(self):
...
+            1, self.get_learning_resource_exports(repo_slug1)['count'])
+        self.delete_learning_resource_export(repo_slug1, repo1_lrid1)
+        self.get_learning_resource_export(
+            repo_slug1, repo1_lrid1, expected_status=HTTP_404_NOT_FOUND)
+
+        self.logout()
+        self.login(self.user)
+
+        # After logout and login session was cleared
+        self.assertEqual(
```
ShawnMilo, line 677: Does this test anything? It looks like it might if line 660 didn't delete the cart. Right now, it looks like this `assertEqual` will always be true, even if there's a bug.

rest/views.py
```diff
@@ -411,9 +417,20 @@ class LearningResourceList(ListAPIView):
 
     def get_queryset(self):
         """Get queryset for a learning resource"""
-        return LearningResource.objects.filter(
+        queryset = LearningResource.objects.filter(
             course__repository__slug=self.kwargs['repo_slug']
         )
+        id_value = self.request.query_params.get('id', None)
+        if id_value is not None:
+            if len(id_value) == 0:
```
ShawnMilo: This check for zero is unnecessary. The list comprehension in the `else` will return an empty list if `if len(x) > 0` is added and will handle both cases.

rest/views.py
```diff
@@ -465,3 +482,109 @@ def get_queryset(self):
...
+                              'lr_id': lr_id})
+        headers = {"Location": url}
+        return Response(
+            {"id": lr_id}, status=status.HTTP_201_CREATED, headers=headers)
+
+    # pylint: disable=unused-argument
+    def delete(self, request, *args, **kwargs):
+        """Delete all ids in session for this repository and user"""
+        try:
+            repo_slug = self.kwargs['repo_slug']
```
ShawnMilo: Pretty minor: These two lines could just be:

``` python
self.request.session[EXPORTS_KEY].pop(self.kwargs['repo_slug'], None)
```

I like to avoid a variable that's only used once. Also, this will prevent a `KeyError` (as long as `self.request.session[EXPORTS_KEY]` always exists).

rest/tests/test_learning_resource.py
noisecapella: I'm fine either way. I'll move it to pylintrc if nobody else comments on this

ui/static/ui/js/listing.js
```diff
@@ -213,6 +229,72 @@ define(
...
+      /**
+       * Update export count badge
+       * @param {number} direction Add this value to
+       * the existing count before rendering
+       */
+      var updateExportCount = function(direction) {
+        var oldCount = $("#export_button").data("export-count");
+        var newCount = oldCount + direction;
+
+        $("#export_button .badge").remove();
```
ShawnMilo, line 252: This deleting and recreating HTML strings sounds like what React does. Is there a reason not to use it here?

Same comment for lines 35 and 46 in `ui/static/ui/js/lr_exports.jsx`. It seems like React should be used to manipulate the DOM.

ui/static/ui/js/listing.js
```diff
@@ -213,6 +229,72 @@ define(
...
+        $(this).data("selected", !selected);
+
+        if (selected) {
+          // remove an item from export cart
+          $.ajax({
+            type: "DELETE",
+            url: "/api/v1/repositories/" + repoSlug +
+            "/learning_resource_exports/" + learningResourceId + "/"
+          }).fail(function() {
+            // TODO: where should this go?
```
ShawnMilo: A `TODO` was left in. The options to me seem to be `alert();` or ignore. Either seem fine, since it shouldn't happen anyway. If it does happen, it means the item isn't there, which is what the user wanted anyway, so it seems like ignoring is slightly better.

If there is a real problem, alerting here isn't the best solution anyway; logging from the Django application (possibly with notifications) is better.

ui/static/ui/js/listing.js
```diff
@@ -213,6 +229,72 @@ define(
...
+            console.error("Unable to delete learning resource " +
+              learningResourceId + " from export list");
+          });
+          updateExportCount(-1);
```
ShawnMilo, line 284: This line and line 292 below (incrementing and decrementing) the counter look like they always fire, even if there's a failure. Could this cause a UI bug?

noisecapella, line 284: Possibly. I'll add a `updateExportCount(1)` to balance it out in `fail`. If we did it in a success callback instead there'd be a bit of UI lag which may be undesirable, but I'm open to opinions here

ShawnMilo, line 284: Okay. Don't forget to add a `-1` below as well.

rest/tests/test_learning_resource.py
noisecapella, line 677: Good catch, I'll add something to the shopping cart first

ui/static/ui/js/listing.js
noisecapella, line 252: I was avoiding React here for consistency with the rest of the listing page.

`errorText` and `messageText` in `lr_exports.jsx` are text strings and don't manipulate the DOM, they're escaped and included as text. Is that what you're referring to? The line numbers may have slipped since last push

ShawnMilo, line 252: I was referring to building the `<ul>` elements (search for `var exports`), and `errorBox` and `messageBox`. If there's a reason not to use React that's fine; I just wanted to mention it.

noisecapella, line 252: I see. `var exports` is actually using React. It's receiving a list of React elements which will be provided to the div a few lines down from that. It looks like a template but it's just some syntactic sugar. Same for `errorBox` and `messageBox`

noisecapella:
@ShawnMilo Hold off on merging, I'm making a last minute change to add the username to the API endpoint so that we can theoretically see other users shopping carts without breaking API compatibility

noisecapella:
I added a username to each `learning_resource_exports` endpoint so that we can maybe look into different shopping carts in the future without making backwards incompatible changes. There are also tests to verify that only the logged in user can do this.

noisecapella:
I also realized that there is no Javascript testing included here. I'll add it for this PR if you want or I'll just do it in the next one where I extend the UI further

ShawnMilo:
Okay. I'll take a look. I don't know about the JavaScript testing part. Do we have a coverage report for that?

noisecapella:
Coveralls should include it. The hook is not working but I believe you can navigate to their site and see latest coverage of our builds there. You can also look at HTML reports at coverage/lcov-report/index.html or see a percent summary at the bottom of Travis's output.

ShawnMilo:
I think using the user's `id` instead of `username` would be better. You never have to worry about encoding issues, and there are places in this commit that have the user ID and then have to do an extra lookup to get the username.

noisecapella:
The members API uses username so using it would be consistent with that API.

rest/permissions.py
```diff
@@ -200,3 +200,16 @@ def has_permission(self, request, view):
             raise Http404()
 
         return True
+
+
+class ViewLearningResourceExportPermission(BasePermission):
+    """
+    Checks that username mentioned is the same one as the request.
+    In the future we may want to let users see each other's exports
+    but for now this just enforces
```
ShawnMilo: This comment looks like it forgot to finish.

ShawnMilo:
There are JavaScript comments with no space after the `//`. Also, many JavaScript comments aren't sentences. I don't know what style guide we're using for JS, so maybe that's not important.

ShawnMilo:
Overall looks good and tests pass. Minor comments, mostly on the comments.

I had some other thoughts about maybe using `get_or_404` for User lookups instead of `get`, and wondering if it would be convenient to add username to the session in addition to the ID, but they were in the tests so it doesn't matter.

rest/tests/test_learning_resource.py
pdpinch: Ok by me to globally disable too-many-ancestors and invalid-name

noisecapella:
We're using the Google style guide: http://google.github.io/styleguide/javascriptguide.xml minus some things we thought were annoying (see `.jscsrc`)
It's enforced by jscs which will fail the build, however it isn't as picky as PEP8.

All JavaScript comments in the diff have one space after the `//` already but I'll add punctuation and make the `get_or_404` change

pdpinch:
@ShawnMilo can this be merged?

ShawnMilo:
Checking now. I missed the label update.

PR #377, opened 2015-07-17
Fixed comparison of FileFields to strings.

### LINKED ISSUE #376: test failure comparing assets to list
`TestImportToy.test_parse_static` fails under some conditions. This is because it's comparing a list of `models.FileField` objects to a list of strings. This evidently works sometimes, probably due to the`__unicode__` representation of a `FileField`, but can fail.

ShawnMilo:
fixes #376

Also explicitly sorts instead of depending on chance.

### DIFF
--- a/importer/tests/test_import.py
+++ b/importer/tests/test_import.py
@@ -286,9 +286,10 @@ class TestImportToy(LoreTestCase):
         ).distinct()
         self.assertEqual(htmls.count(), 1)
         self.assertEqual(
-            [asset.asset for asset in htmls[0].static_assets.all()],
-            [
+            sorted([
+                asset.asset.name for asset in htmls[0].static_assets.all()]),
+            sorted([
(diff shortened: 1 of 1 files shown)


importer/tests/test_import.py
```diff
@@ -286,9 +286,10 @@ def get_counts():
         ).distinct()
         self.assertEqual(htmls.count(), 1)
         self.assertEqual(
-            [asset.asset for asset in htmls[0].static_assets.all()],
-            [
+            sorted([
```
ShawnMilo, line 289: The important part of this change is the addition of `.name` to the asset.

noisecapella, line 289: What was it doing before? Coercing FileField to a string before comparison?

ShawnMilo, line 289: I don't know. I assume it was, and maybe still does on some machines. It just broke for me on a fresh installation. In any case, this is more explicit.

PR #378, opened 2015-07-17
Implemented learning resource panel updating on every panel open

### LINKED ISSUE #338: Learning Resources Panel needs to refresh every time a user opens it
Currently the Learning Resources Panel refreshes the content only when the user clicks it for the first time or after clicking on another learning resource.

This is not a problem unless the user creates a new taxonomy. In this case the new taxonomy will not show up int he panel unless it refreshes.

noisecapella:
Fixes #338 

This causes the panel to always update which means existing text or errors will be forgotten if the user closes the panel.

### DIFF
--- a/ui/static/ui/js/learning_resources.jsx
+++ b/ui/static/ui/js/learning_resources.jsx
@@ -222,9 +222,11 @@ define('learning_resources', [
     VocabularyOption: VocabularyOption,
     LearningResourcePanel: LearningResourcePanel,
     loader: function (repoSlug, learningResourceId, container) {
+      // Unmount and remount the component to ensure that its state
+      // is always up to date with the rest of the app.
+      React.unmountComponentAtNode(container);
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/learning_resources.jsx
```diff
@@ -222,9 +222,9 @@ define('learning_resources', [
     VocabularyOption: VocabularyOption,
     LearningResourcePanel: LearningResourcePanel,
     loader: function (repoSlug, learningResourceId, container) {
+      React.unmountComponentAtNode(container);
```
giocalitri, line 227: Can you please add a comment explaining what this fixes? it would be nice to be able to search for it in future :-)

giocalitri:
Functionality is fine: if you add the comment can be merged after code freeze

giocalitri:
perfect

PR #382, opened 2015-07-17
Fixed display of vocabulary terms containing spaces.

### LINKED ISSUE #367: Facet term text is broken up by spaces
@Ferdi pointed this out to me earlier. To reproduce, create terms with spaces in them. They show up on the listing page as separate words split by space

ShawnMilo:
```
Fixed display of vocabulary terms containing spaces.

fixes #367

I spent some (too much) time on the command line using the elasticsearch
module trying to figure out how to fix the search to work with spaces.

In the end, I went for the obvious, lame solution. Eventually it seems
like we'll want to ditch Haystack and use Elasticsearch directly.
```

### DIFF
--- a/importer/tests/test_import.py
+++ b/importer/tests/test_import.py
@@ -286,9 +286,10 @@ class TestImportToy(LoreTestCase):
         ).distinct()
         self.assertEqual(htmls.count(), 1)
         self.assertEqual(
-            [asset.asset for asset in htmls[0].static_assets.all()],
-            [
+            sorted([
+                asset.asset.name for asset in htmls[0].static_assets.all()]),
+            sorted([
                 "assets/edX/toy/TT_2012_Fall/essays_x250.png",
                 "assets/edX/toy/TT_2012_Fall/webGLDemo.css",
-            ]
+            ])
         )
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -87,9 +87,12 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
         """
         prepared = super(LearningResourceIndex, self).prepare(obj)
         for vocab in Vocabulary.objects.all():
-            # Values with spaces do not work, so we use the slug.
+            # Values with spaces do not work, so replace them with underscores.
+            # Slugify doesn't work because it adds hypens, which are also
+            # split by Elasticsearch.
             terms = [
-                term.label for term in obj.terms.filter(vocabulary_id=vocab.id)
+                term.label.replace(" ", "_")
+                for term in obj.terms.filter(vocabulary_id=vocab.id)
             ]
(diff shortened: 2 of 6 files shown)


importer/tests/test_import.py
```diff
@@ -286,9 +286,10 @@ def get_counts():
         ).distinct()
         self.assertEqual(htmls.count(), 1)
         self.assertEqual(
-            [asset.asset for asset in htmls[0].static_assets.all()],
-            [
+            sorted([
```
ShawnMilo, line 289: These changes belong to commit #377; hopefully it'll be merged before this one.

search/search_indexes.py
```diff
@@ -88,8 +88,11 @@ def prepare(self, obj):
         prepared = super(LearningResourceIndex, self).prepare(obj)
         for vocab in Vocabulary.objects.all():
             # Values with spaces do not work, so we use the slug.
+            # However, slugs use a hypen, which is still split, so
```
noisecapella: hypen -> hyphen :)

search/tests/test_search_view.py
```diff
@@ -28,3 +28,14 @@ def test_single_repo(self):
         )
         resp = self.client.get(reverse("repositories", args=(new_repo.slug,)))
         self.assertNotContains(resp, self.resource.title)
+
+    def test_terms_with_spaces(self):
+        """
+        Terms with spaces should show up in facet list correctly.
+        """
+        resource = self.resource
```
giocalitri: I guess you can avoid this and reuse `self.resource`

ui/views.py
```diff
@@ -168,7 +168,7 @@ def extra_context(self):
                 "repo": self.repo,
                 "perms_on_cur_repo": get_perms(self.request.user, self.repo),
                 "vocabularies": {
-                    k: v
+                    k: [x + (x[0].replace("_", " "),) for x in v]
```
giocalitri, line 173: Are you trying to reconstruct the therm here?

ShawnMilo, line 173: This replaces `("slug", "count")` with `("slug", "count", "pretty name")`.

For example, `("very_difficult", "3")` would become `("very_difficult", "3", "very difficult")`.

giocalitri:
This look good to me.
I tested the functionality and it solves the space problem, but it does not solve the case where the therm is something like `anc@ra`: it splits on the `@` and there are 2 facets `anc` and `ra`

@pdpinch agreed to merge as it is and open another issue

ShawnMilo:
@giocalitri #384 added. Perhaps base64 encoding will be the silver bullet.

Ferdi:
Without complicating this further, I suggest we only allow users to provide alphanumeric term names

giocalitri:
@Ferdi better to continue the discussion on #384

PR #383, opened 2015-07-17
Added sorting to search results

### LINKED ISSUE #349: Search: add ui for choosing sort order
Options:
should include analytics fields

Default: 
number of views (from analytics)

This needs further description.

giocalitri:
Closes #349 

[#98962342]

### DIFF
--- a/learningresources/tests/base.py
+++ b/learningresources/tests/base.py
@@ -57,6 +57,22 @@ class LoreTestCase(TestCase):
         if return_body:
             return resp.content.decode("utf-8")
 
+    def create_resource(self, **kwargs):
+        """Creates a learning resource with extra fields"""
+        learn_res = create_resource(
+            course=self.course,
+            parent=None,
+            resource_type=kwargs.get('resource_type') or "example",
+            title=kwargs.get('title') or "other silly example",
+            content_xml=kwargs.get('content_xml') or "<blah>other blah</blah>",
+            mpath=kwargs.get('mpath') or "/otherblah"
+        )
+        learn_res.xa_nr_views = kwargs.get('xa_nr_views', 0)
+        learn_res.xa_nr_attempts = kwargs.get('xa_nr_attempts', 0)
+        learn_res.xa_avg_grade = kwargs.get('xa_avg_grade', 0)
+        learn_res.save()
+        return learn_res
+
     def setUp(self):
         """set up"""
         super(LoreTestCase, self).setUp()
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -38,6 +38,10 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     # page because that page is always for a single repository.
     repository = indexes.CharField(faceted=True)
 
+    nr_views = indexes.IntegerField(model_attr="xa_nr_views")
+    nr_attempts = indexes.IntegerField(model_attr="xa_nr_attempts")
+    avg_grade = indexes.FloatField(model_attr="xa_avg_grade")
+
     def get_model(self):
         """Return the model for which this configures indexing."""
         return LearningResource
--- a/search/tests/test_indexing.py
+++ b/search/tests/test_indexing.py
@@ -1,5 +1,7 @@
 """Tests for search engine indexing."""
+from __future__ import unicode_literals
 
+from search.sorting import LoreSortingFields
 from search.tests.base import SearchTestCase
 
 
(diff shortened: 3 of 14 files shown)


search/tests/test_sorting.py
```diff
@@ -0,0 +1,117 @@
+"""
```
ShawnMilo, line 1: This whole file just tests some constants and isn't necessary. It doesn't improve test coverage except for a single `if` statement, `search_sorting.py` line 42. It would be better to add a call to `get_sorting_option` with a missing field and delete this file.

learningresources/tests/base.py
```diff
@@ -57,6 +57,22 @@ def assert_status_code(self, url, code, return_body=False):
         if return_body:
             return resp.content.decode("utf-8")
 
+    def create_resource(self, **kwargs):
+        """Creates a learning resource with extra fields"""
+        learn_res = create_resource(
+            course=self.course,
+            parent=None,
+            resource_type=kwargs.get('resource_type') or "example",
+            title=kwargs.get('title') or "othet silly example",
```
ShawnMilo: Very minor, but "other" is spelled "othet" on two lines here.

learningresources/tests/base.py
```diff
@@ -57,6 +57,22 @@ def assert_status_code(self, url, code, return_body=False):
...
+            content_xml=kwargs.get('content_xml') or "<blah>othet blah</blah>",
+            mpath=kwargs.get('mpath') or "/otherblah"
+        )
+        learn_res.xa_nr_views = kwargs.get('xa_nr_views') or 0
```
ShawnMilo: These can all be `.get(thing, 0)`, instead of `or 0`.

search/search_indexes.py
```diff
@@ -38,6 +38,19 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     # page because that page is always for a single repository.
     repository = indexes.CharField(faceted=True)
 
+    nr_views = indexes.IntegerField(
+        model_attr="xa_nr_views",
+        faceted=True,
```
ShawnMilo: You don't need `faceted=True` on these. I deleted them all locally and tested and it still works the same.

search/tests/test_indexing.py
```diff
@@ -36,3 +38,59 @@ def test_strip_xml(self):
         self.resource.save()
         self.assertTrue(self.count_results("you're it") == 1)
         self.assertTrue(self.count_results(xml) == 0)
+
+    def test_sorting(self):
```
ShawnMilo, line 42: This test would be much easier to read (and shorter) if there were just as many resources created as sort options, with values that ensured each one would be first for a particular sort, then sort each way, asserting the order expected.

The test here takes more effort to understand than it has to.

ui/views.py
```diff
@@ -136,6 +137,16 @@ def __call__(self, request, repo_slug):
         if repo_slug not in set([x.slug for x in repos]):
             return HttpResponseForbidden("unauthorized")
         self.repo = [x for x in repos if x.slug == repo_slug][0]
+        # get sorting from params if it's there
+        params = dict(request.GET.copy())
+        if ('sortby' in params and
```
ShawnMilo: The `sortby` can just be popped here, which would make this `if` statement shorter, and ensure it wasn't there so the lines below that pop it just to make sure it isn't there can be deleted.

search/search_indexes.py
giocalitri: I actually added it to support future faceting, but I guess they are useless now

search/tests/test_sorting.py
giocalitri, line 1: The meaning of this is to test the constants, since the strings need to match the index names, and the content of the lists returned.
If I do it in the interface, I need to check the content of the page and that would be much harder.

ShawnMilo, line 1: I don't see why constants have to be tested at all, when it doesn't do anything for test coverage.

giocalitri, line 1: I disagree: the meaning is actually to test that they do not change, because other part of the code can be impacted.
For example the default sorting is used in many places. 
The fact that the sorting is a tuple of 2 elements and the first is the index name is important.
The behavior of the functions needs to be consistent with what I defined here.
How do I test this stuff at high level?

giocalitri, line 1: And I totally disagree with  "it doesn't do anything for test coverage."
coverage.py just tests that a piece of code is actually called. The fact that it works as expected is not tested.

ui/views.py
giocalitri: but to do this I need to pop it from the `request.GET` that makes it a bit weird (and I have to check if I can actually modify the request object)

giocalitri: It can be done: just checked and `request.GET` has a method `pop`.
It still feels weird to me to modify the request object, but I have no strong opinions about this.

Anyway, I agree that my code could be better and I modified it to:

```
        # get sorting from params if it's there
        sortby = dict(request.GET.copy()).get('sortby', [])
        if (len(sortby) > 0 and
                sortby[0] in LoreSortingFields.all_sorting_fields()):
            self.sortby = sortby[0]
        else:
            # default value
            self.sortby = LoreSortingFields.DEFAUT_SORTING_FIELD
```

Please let me know if this is enough or you prefer the `pop`

noisecapella: DEFAUT looks like a typo

noisecapella: Also, `dict` will create a copy itself so wouldn't `request.GET.copy` be redundant? If the `pop` were done to `params` it wouldn't modify the original request (which I agree would be sketchy)

giocalitri: @noisecapella I initially agreed that removing dict was a good idea, but when I do it, my code does not work properly.
I guess the `type(request.GET)` even if it resembles a `dict`, is slightly different. So I will leave it in this way

search/tests/test_indexing.py
```diff
@@ -36,3 +50,79 @@ def test_strip_xml(self):
...
+        res1 = self.create_resource(**dict(
+            resource_type="example",
+            title="silly example 1",
+            content_xml="<blah>blah 1</blah>",
+            mpath="/blah1",
+            xa_nr_views=1001,
+            xa_nr_attempts=99,
+            xa_avg_grade=9.9
+        ))
+        res2 = self.create_resource(**dict(
```
ShawnMilo, line 56: There are no tests where `res2` comes first. `res1` comes first once, and `res3` comes first twice.

search/tests/test_indexing.py
```diff
@@ -36,3 +50,79 @@ def test_strip_xml(self):
...
+            avg_grades=(
+                res3.xa_avg_grade, res2.xa_avg_grade, res1.xa_avg_grade)
+        )
+        # sorting by number of attempts
+        results = self.search(
+            '',
+            sorting=LoreSortingFields.SORT_BY_NR_ATTEMPTS[0]
+        )
+        # expected position (res3, res1, res2)
+        self.check_resources_sorting(
```
ShawnMilo: I think these tests should be much simpler; just assert that a specific resource is first, instead of passing and comparing all these values.

giocalitri: I think that a test of the whole sequence is needed: this is the only place where we will ever do it.

ShawnMilo: You have one resource per sort type. If you make each one the "top" result for a sort option, you will have a complete test.

search/tests/test_indexing.py
ShawnMilo, line 42: Note: Comment above has been addressed, but it didn't require changing line 42, so it didn't get hidden.

ShawnMilo:
Looks okay to me. Will merge after Travis says it's okay.

PR #392, opened 2015-07-20
Refactored permission check for listing view

giocalitri:
Changed the way we check the repo permission in the listing view and fixed a test that was not testing the unauthorized listing view in the right way.

### DIFF
--- a/ui/tests/test_learningresources_views.py
+++ b/ui/tests/test_learningresources_views.py
@@ -86,15 +86,23 @@ class TestViews(LoreTestCase):
         body = resp.content.decode("utf-8")
         self.assertTrue(repo_name in body)
 
-    def test_listing_unauthorized(self):
-        """View listing page."""
-        # Not authorized to view this repository...
-        body = self.assert_status_code(
+    def test_listing_not_found(self):
+        """View listing page, but repo does not exist."""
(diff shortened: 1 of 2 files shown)


ui/tests/test_learningresources_views.py
```diff
@@ -84,15 +84,23 @@ def test_create_repo_post(self):
...
+    def test_listing_unauthorized(self):
+        """View listing page, but not authorized to view this repository."""
+        self.logout()
+        self.login(self.USERNAME_NO_REPO)
+        self.assert_status_code(
+            self.repository_url,
             UNAUTHORIZED,
             return_body=True
         )
-        self.assertTrue("unauthorized" in body)
```
noisecapella, line 97: Is this line still useful?

giocalitri, line 97: The body does not contain the string any more because now I raise an exception instead of returning a response object

noisecapella, line 97: Makes sense

noisecapella:
This is good to merge, however it looks like it needs some rebasing before this happens

giocalitri:
Rebased and pushed again

noisecapella:
Ok, I'll merge after tests pass

PR #393, opened 2015-07-20
Used different haystack index for tests to prevent conflict with web application

noisecapella:
The environment variable tells haystack to use a different name for its index


giocalitri:
looks good to me: will tag "merge after freeze" after tests pass

giocalitri:
tested functionality as well and it's fine

PR #396, opened 2015-07-20
Grouped REST tests by common endpoint

noisecapella:


noisecapella:
None of the tests should be changed, just reorganized. This must be merged before #371 can be looked at

noisecapella:
There is one change actually in base.py at the top of the PR. It's a fix where `self.repo` was used mistakenly in place of `repo` passed as an argument

rest/tests/test_learning_resource.py
```diff
@@ -0,0 +1,425 @@
...
+        resp = self.client.get(base_url)
+        self.assertEqual(HTTP_200_OK, resp.status_code)
+        types = as_json(resp)
+
+        self.assertEqual(sorted([lrt.name for lrt
+                                 in LearningResourceType.objects.all()]),
+                         sorted([t['name'] for t in types['results']]))
+
+
+# pylint: disable=too-many-ancestors
```
giocalitri, line 218: if this is not to disable it globally in the entire file, you should move it inside the class after the docstring

rest/tests/test_members.py
```diff
@@ -579,3 +583,170 @@ def test_members_delete(self):
             group_type=BaseGroupTypes.ADMINISTRATORS
         )
         self.assert_users_count(admins=1)
+
+
+# pylint: disable=too-many-ancestors
```
giocalitri, line 588: same as before

rest/tests/test_repository.py
```diff
@@ -0,0 +1,322 @@
...
+                repo_base=REPO_BASE,
+            ))
+        self.assertEqual(HTTP_200_OK, resp.status_code)
+        repositories = as_json(resp)
+        self.assertEqual(41, repositories['count'])
+        self.assertEqual(
+            [RepositorySerializer(Repository.objects.get(id=x['id'])).data
+             for x in expected[19:39]], repositories['results'])
+
+    # for some reason Pylint has issues with this name
```
giocalitri: the issue is the length of the name (I guess), but in the tests we can even disable this globally

rest/tests/test_repository.py
```diff
@@ -0,0 +1,322 @@
...
+
+            # created_by is set internally and should not show up in output.
+            self.assertNotIn('created_by', new_dict)
+            repository = Repository.objects.get(slug=new_dict['slug'])
+            self.assertEqual(repository.created_by.id, self.user.id)
+
+        assert_not_changed(self.create_repository(repo_dict, skip_assert=True))
+
+
+# pylint: disable=too-many-ancestors
```
giocalitri, line 168: same as before

rest/tests/test_rest.py
```diff
@@ -4,31 +4,12 @@
 from __future__ import unicode_literals
```
giocalitri, line 5: I guess this file can become `test_rest_root.py`

giocalitri, line 5: unless you have something else in mind

rest/tests/test_learning_resource.py
noisecapella, line 218: This doesn't work, I believe it needs to be before the class to have an effect on it

rest/tests/test_vocabulary.py
```diff
@@ -0,0 +1,889 @@
...
+        vocab_slug = self.create_vocabulary(
+            self.repo.slug, vocab_copy)['slug']
+
+        self.assertEqual(
+            ['example'],
+            self.get_vocabulary(
+                self.repo.slug, vocab_slug)['learning_resource_types'])
+
+
+# pylint: disable=too-many-ancestors
```
giocalitri, line 604: same as before

rest/tests/test_learning_resource.py
giocalitri, line 218: if it is for the class it's fine inside (but feel free to forget the comment if it does not work)

rest/tests/test_rest.py
noisecapella, line 5: I've renamed it `test_misc` since it's intended for REST-related tests to go there if they don't fit elsewhere

giocalitri:
only the comment about pylint (I assumed you kept all the tests, but I did not check them one by one)

noisecapella:
I tried moving the `too-many-ancestors` directives after the class docstring but then it doesn't apply to the class anymore. I've moved all `invalid-name` directives to the global level

giocalitri:
ok, it looks good: will merge after tests pass

PR #374, opened 2015-07-17
Added vocabulary name and slug to facets

amir-qayyum-khan:
Hi

In this PR i added vocabulary name to facets along with slug.

https://github.com/mitodl/lore/issues/318
@carsongee @pdpinch @pwilkins

PR #397, opened 2015-07-21
Changed vocabulary term indexing from string to integer.

### LINKED ISSUE #384: Handle non-alphanumeric characters in terms.
Similar to #367, except other characters, including `-` and `@` will cause the same problem.
#367 was fixed by #382, but we should handle in a more universal manner. The current solution uses `string.replace()`, because Django's `slugify` adds hypens, which have the same problem.
-  Find out which characters have the issue (hopefully from Elasticsearch's documentation.
-  Create a function similar to slugify, but which handles this case.

Preferably have a reverse function for displaying the term in the template as well; the current solution just replaces all `_` characters with spaces. It would be nice to return the "special" character instead. This would mean somehow encoding the original value. Perhaps even something as simple as base64 encoding and decoding. If base64 encoding is selected, that eliminates the need to research and replace specific characters, as long as `=` isn't a problem for Elasticsearch.

ShawnMilo:
Prevents any funny business with special characters. Incidentally saves
space in Elasticsearch.

Fixes #384
Replaces and closes #374.
Replaces and closes #316.

### DIFF
--- a/search/__init__.py
+++ b/search/__init__.py
@@ -20,8 +20,8 @@ def get_sqs():
     # Add hard-coded facets.
     for facet in ("course", "run", "resource_type"):
         sqs = sqs.facet(facet)
-    # Add dynamic facets (from taxonomy). Values with spaces do not work,
-    # so use the slug.
-    for slug in Vocabulary.objects.all().values_list("slug", flat=True):
-        sqs = sqs.facet(slug)
+    # Add dynamic facets (from taxonomy). Certain characters cause problems,
+    # so use the primary key.
+    for vocabulary_id in Vocabulary.objects.all().values_list("id", flat=True):
+        sqs = sqs.facet(vocabulary_id)
     return sqs
--- a/ui/templates/includes/facet_panel.html
+++ b/ui/templates/includes/facet_panel.html
@@ -108,28 +108,28 @@
     </div>
   </div>
 </div>
-{% for name, vocab in vocabularies.items %}
-{% if vocab %}
+{% for vocab, terms in vocabularies.items %}
+{% if terms %}
 <div class="panel panel-default">
   <div class="panel-heading">
     <h4 class="panel-title">
       <a class="accordion-toggle" data-toggle="collapse" data-parent="#accordion" href="#collapse-vocab-{{ forloop.counter }}">
-        {{ name|title }}
+        {{ vocab.1|title }}
       </a>
     </h4>
   </div>
   <div id="collapse-vocab-{{ forloop.counter }}" class="panel-collapse collapse in">
     <div class="panel-body">
       <ul class="icheck-list">
-        {% for term in vocab %}
+        {% for term in terms %}
         <li>
-          <input id="check-{{ name }}-{{ forloop.counter }}"
+          <input id="check-{{ vocab.0 }}-{{ forloop.counter }}"
                  tabindex="add" type="checkbox" class="icheck-11"
-                 data-facet-name="{{ name|urlencode }}_exact"
+                 data-facet-name="{{ vocab.0 }}_exact"
                  data-facet-value="{{ term.0 }}"
           >
-          <label for="check-{{ name }}-{{ forloop.counter }}">{{ term.2 }}</label>
-          <span class="badge">{{ term.1 }}</span>
+          <label for="check-{{ vocab.0 }}-{{ forloop.counter }}">{{ term.1 }}</label>
+          <span class="badge">{{ term.2 }}</span>
         </li>
         {% endfor %}
       </ul>
--- a/ui/views.py
+++ b/ui/views.py
@@ -31,7 +31,7 @@ from roles.api import assign_user_to_repo_group
 from roles.permissions import GroupTypes, RepoPermission
 from search import get_sqs
 from search.sorting import LoreSortingFields
-from taxonomy.models import Vocabulary
+from taxonomy.models import Vocabulary, Term
 from ui.forms import UploadForm, RepositoryForm
 
 log = logging.getLogger(__name__)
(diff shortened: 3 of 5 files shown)


search/__init__.py
```diff
@@ -20,8 +20,8 @@ def get_sqs():
...
     for facet in ("course", "run", "resource_type"):
         sqs = sqs.facet(facet)
-    # Add dynamic facets (from taxonomy). Values with spaces do not work,
-    # so use the slug.
-    for slug in Vocabulary.objects.all().values_list("slug", flat=True):
-        sqs = sqs.facet(slug)
+    # Add dynamic facets (from taxonomy). Certain characters cause problems,
+    # so use the primary key.
+    for v_id in Vocabulary.objects.all().values_list("id", flat=True):
+        sqs = sqs.facet(v_id)
```
noisecapella: How does the SearchQuerySet know that `v_id` is a vocabulary id and not a term id or other number?

ShawnMilo: It can only be a vocabulary. It so happens it's an integer, but if you look at the lines 25-26 this replaces, the logic has not changed at all.

It can never be a term because, in the key/value store of Elasticsearch, the vocabulary integers are keys and the term integers are values.

ShawnMilo: To more fully elaborate, unless we add hard-coded values to course, run, and resource_type that are integers for some reason, there can be no collision.

ui/views.py
```diff
@@ -123,6 +123,53 @@ def create_repo(request):
     )
 
 
+def get_vocabularies(facets):
+    """
+    Parse facet information for the template.
```
noisecapella, line 128: This is great documentation but can you reformat into a napoleon docstring?

ui/views.py
```diff
@@ -123,6 +123,53 @@ def create_repo(request):
...
+    This is for ease-of-use in the template, where the integer primary keys
+    are used for search links and the names/labels are used for display.
+    """
+
+    if "fields" not in facets:
+        return {}
+    vocab_ids = []
+    term_ids = []
+    for vocab_id, counts in facets["fields"].items():
+        if not vocab_id.isdigit():
```
noisecapella, line 153: Wouldn't we want to raise an exception if this caused a problem?

ShawnMilo, line 153: No, this is to filter out course, run, and resource_type.

ui/templates/includes/facet_panel.html
```diff
@@ -108,28 +108,28 @@ <h4 class="panel-title">
...
     <div class="panel-body">
       <ul class="icheck-list">
-        {% for term in vocab %}
+        {% for term in terms %}
         <li>
-          <input id="check-{{ name }}-{{ forloop.counter }}"
+          <input id="check-{{ vocab.0 }}-{{ forloop.counter }}"
                  tabindex="add" type="checkbox" class="icheck-11"
-                 data-facet-name="{{ name|urlencode }}_exact"
+                 data-facet-name="{{ vocab.0|urlencode }}_exact"
```
noisecapella: Can you remove the `urlencode` here or add it above and below to make it consistent with the rest of this file?

ui/views.py
ShawnMilo, line 128: Docstring updated.

ui/templates/includes/facet_panel.html
ShawnMilo: Since they'll be integers, I'll remove the `urlencode`.

ui/views.py
```diff
@@ -123,6 +123,58 @@ def create_repo(request):
...
+    vocabs = {
+        x: y for x, y in Vocabulary.objects.filter(
+            id__in=vocab_ids).values_list('id', 'name')
+    }
+    terms = {
+        x: y for x, y in Term.objects.filter(
+            id__in=term_ids).values_list('id', 'label')
+    }
+    vocabularies = {}
+    for v_id, t_data in facets["fields"].items():
```
noisecapella: Can you rename to vocab_id and term_data for consistency with the above for loop?

noisecapella: Or whatever names make sense in this context, as long as they're the same

ShawnMilo: As discussed in HipChat, I've decided not to do this. It adds duplicated code in multiple places to add `vocab_` and `term_`, then requires more code to strip them out again for database lookups.

Less code is always better, and it's more efficient as-is.

noisecapella: I won't argue further on changing the facet key, but please change v_id to something more descriptive. If it can contain `resource_type` or `run` it shouldn't be named `v_id`

ShawnMilo: Renamed:

``` diff
diff --git search/__init__.py search/__init__.py
index b4067e7..e4915b9 100644
--- search/__init__.py
+++ search/__init__.py
@@ -22,6 +22,6 @@ def get_sqs():
         sqs = sqs.facet(facet)
     # Add dynamic facets (from taxonomy). Certain characters cause problems,
     # so use the primary key.
-    for v_id in Vocabulary.objects.all().values_list("id", flat=True):
-        sqs = sqs.facet(v_id)
+    for vocabulary_id in Vocabulary.objects.all().values_list("id", flat=True):
+        sqs = sqs.facet(vocabulary_id)
     return sqs
diff --git ui/views.py ui/views.py
index ba914a2..8080449 100644
--- ui/views.py
+++ ui/views.py
@@ -165,12 +165,12 @@ def get_vocabularies(facets):
             id__in=term_ids).values_list('id', 'label')
     }
     vocabularies = {}
-    for v_id, t_data in facets["fields"].items():
-        if not v_id.isdigit():
+    for vocabulary_id, term_data in facets["fields"].items():
+        if not vocabulary_id.isdigit():
             continue
-        vocab = (v_id, vocabs[int(v_id)])
+        vocab = (vocabulary_id, vocabs[int(vocabulary_id)])
         vocabularies[vocab] = []
-        for t_id, count in t_data:
+        for t_id, count in term_data:
             vocabularies[vocab].append((t_id, terms[int(t_id)], count))
     return vocabularies
```

noisecapella:
I will merge this after tests pass

ShawnMilo:
Cool, thanks. :smiley_cat:

PR #399, opened 2015-07-21
Fixed preview link not showing up in list view

noisecapella:
When `preview_url` was changed from a property to a method I forgot to update the template accordingly. There's also a test to verify this


carsongee:
:goat:

PR #400, opened 2015-07-21
Updated Haystack to 2.4.0

carsongee:
- Removed automatic index update from deployment

The deprecated class views we are using were not removed, so is working fine with 2.4.0.  Also, after having learned from @giocalitri and @ShawnMilo that ES isn't used in the listing page it occurred to me that the only time we would ever need to rebuild the index is if we added a new facet type, or changed what is searchable in the text field.  Since those will likely be very rare events, I think we can remove the index update from the build process and just do them manually as needed since I don't see this happening much at all.  I am definitely open for counter suggestions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,4 @@ deploy:
   run:
     - "sleep 15"
     - "python manage.py migrate --noinput"
-    - "python manage.py update_index"
     - "python manage.py sync_permissions"
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Django==1.8.3
 django-appconf==1.0.1
 django-bootstrap3==6.1.0
 django-compressor==1.5
-django-haystack==2.3.1
+django-haystack==2.4.0
 djangorestframework==3.1.3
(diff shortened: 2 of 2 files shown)


Ferdi:
I'm surprised that listing page is not powered by  ES. Can we change that?

carsongee:
It is using the results to return database objects, so it is a bit of a hybrid.  They do have suggestions on reducing those DB calls with: https://django-haystack.readthedocs.org/en/v2.4.0/best_practices.html#avoid-hitting-the-database

giocalitri:
@carsongee the first link you sent me this afternoon gives a clearer explanation of what we need to do to avoid to hit the DB (but not sure how up to date it is):
https://www.dominicrodger.com/order-n-haystack.html

noisecapella:
Looked over it, I have no problem with reindexing on an as-needed basis

ShawnMilo:
I agree with @giocalitri's comments in the stand-up. The plan sounds good to me.

carsongee:
Looks like consensus, merging.

PR #403, opened 2015-07-22
import xanalytics data

### LINKED ISSUE #166: XAnalytics
Xanalytics app will be a layer between the LORE system and MITx analytics system (xanalytics). It will connect with the xanalytics web service and get and update the relevant metatada in LORE database. 
- [x] Define the protocol of communication between LORE and Xanalytics. Xanalytics team will provide the apis for us to integrate
- [x] Until the API is ready, we will apply some fake data at the time of course import
- [ ] Integrate with xanalytics system using the protocol above.

ShawnMilo:
closes #166

Uses [responses](https://pypi.python.org/pypi/responses/0.4.0)
for testing API calls, based on
[this blog post](http://cramer.io/2014/05/20/mocking-requests-with-responses/).

### DIFF
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -17,6 +17,7 @@ from django.core.files.storage import default_storage
 from lxml import etree
 from xbundle import XBundle, DESCRIPTOR_TAGS
 
+from importer.tasks import populate_xanalytics_fields
 from learningresources.api import (
     create_course,
     create_resource,
@@ -128,6 +129,7 @@ def import_course(bundle, repo_id, user_id, static_dir):
     )
     import_static_assets(course, static_dir)
     import_children(course, src, None, '')
+    populate_xanalytics_fields.delay(course.id)
     return course
 
 
--- a/importer/tasks.py
+++ b/importer/tasks.py
@@ -3,11 +3,63 @@ Celery tasks for import module.
 """
 from __future__ import unicode_literals
 
+import json
+import logging
+
+from django.conf import settings
+import requests
+
 from lore.celery import async
-from importer.api import import_course_from_file
+from learningresources.api import update_xanalytics
+from xanalytics import send_request, get_result
+
+log = logging.getLogger(__name__)
+
+RETRY_LIMIT = 10
 
 
 @async.task
 def import_file(path, repo_id, user_id):
     """Asynchronously import a course."""
+    from importer.api import import_course_from_file
     import_course_from_file(path, repo_id, user_id)
+
+
+@async.task
+def populate_xanalytics_fields(course_id):
+    """
+    Initiate request to xanalytics API to get stats for a course,
+    then trigger async job to retrieve results when they become available.
+    Args:
+        course_id (int): primary key of a Course
+    """
+    if settings.XANALYTICS_URL != "":
+        token = send_request(settings.XANALYTICS_URL + "/create", course_id)
+        check_for_results.apply_async(
+            kwargs={"token": token, "wait": 5, "attempt": 1}, countdown=5)
+
+
+@async.task
+def check_for_results(token, wait, attempt):
+    """
+    Check for xanalytics results for a course.
+    Args:
+        token (string): Token received from xanalytics server.
+        wait (int): Seconds to wait before repository.
+        attempt (int): Attempt number, so we don't try forever.
+    """
+    resp = get_result(settings.XANALYTICS_URL + "/status", token)
+    if resp["status"] == "still busy" and attempt < RETRY_LIMIT:
+        attempt += 1
+        wait *= 2
+        check_for_results.apply_async(
+            kwargs={"token": token, "wait": wait, "attempt": attempt},
+            countdown=wait,
+        )
+    if resp["status"] == "complete":
+        content = requests.get(resp["url"]).content
+        try:
+            data = json.loads(content)
+            update_xanalytics(data)
+        except (ValueError, TypeError):
+            log.error("Unable to parse xanalytics response: %s", content)
--- a/importer/tests/test_xanalytics.py
+++ b/importer/tests/test_xanalytics.py
@@ -1,0 +1,121 @@
+"""
+Tests for hitting the xanalytics Celery tasks.
+"""
+
+from __future__ import unicode_literals
+
+import json
+import logging
+
+from django.conf import settings
+import responses
+
+from importer.tasks import populate_xanalytics_fields, check_for_results
+from learningresources.tests.base import LoreTestCase
+
+log = logging.getLogger(__name__)
+# pylint: disable=no-self-use
+
+
+class TestAPI(LoreTestCase):
+    """
+    Tests using mocks to avoid actual HTTP requests.
+    """
+    def setUp(self):
+        """Override setting."""
+        super(TestAPI, self).setUp()
+        self.original_url = settings.XANALYTICS_URL
+        settings.XANALYTICS_URL = "http://example.com"
+
+    def tearDown(self):
+        """Restore original setting."""
+        super(TestAPI, self).tearDown()
+        settings.XANALYTICS_URL = self.original_url
+
+    @responses.activate
+    def test_populate_xanalytics_fields(self):
+        """Test kicking off a request."""
+        responses.add(
+            responses.POST,
+            'http://example.com/create',
+            body='{"token": "abcde"}',
+            content_type="application/json"
+        )
+        # We must mock the /status endpoint as well,
+        # because a status check is queued when /create is called.
+        reply = {"status": "still busy"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        populate_xanalytics_fields(1234)
+
+    @responses.activate
+    def test_check_for_results_waiting(self):
+        """Test kicking off a request."""
+        reply = {"status": "still busy"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        check_for_results("abcde", 10, 1)
+
+    @responses.activate
+    def test_check_for_results_done(self):
+        """Test kicking off a request."""
+        file_reply = {
+            "course_id": "123",
+            "module_medata": [
+                {
+                    "module_id": "1",
+                    "xa_nr_views": "3",
+                    "xa_nr_attempts": "25"
+                },
+                {
+                    "module_id": "2",
+                    "xa_nr_views": "7",
+                    "xa_nr_attempts": "99"
+
+                }
+            ]
+        }
+        responses.add(
+            responses.GET,
+            'http://example.com/foo.json',
+            body=json.dumps(file_reply),
+            content_type="application/json"
+        )
+
+        reply = {"status": "complete", "url": "http://example.com/foo.json"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        check_for_results("abcde", 10, 1)
+
+    @responses.activate
+    def test_check_for_results_bad(self):
+        """Test kicking off a request."""
+        responses.add(
+            responses.GET,
+            'http://example.com/foo.json',
+            body='{"invalid": "JSON", "because": "developer comma",}',
+            content_type="application/json"
+        )
+        reply = {"status": "complete", "url": "http://example.com/foo.json"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        self.assertRaises(
+            ValueError,
+            check_for_results("abcde", 10, 1)
+        )
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -307,6 +307,29 @@ def import_static_assets(course, path):
                 create_static_asset(course.id, django_file)
 
 
+def update_xanalytics(data):
+    """
+    Update xanalytics fields for a LearningResource.
+    Args:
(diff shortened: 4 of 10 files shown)


carsongee:
We had previously evaluated responses vs httppretty and decided to go with httppretty (https://github.com/mitodl/orcoursetrion, https://github.com/mitodl/pylti, https://github.com/mitodl/PyLmod).  I know you didn't know that, but I am curious why you picked responses?  We chose httpretty because it supports requests as well as several other http libs. I was under the impression that responses only supported requests (similar to betamax only supporting requests, whereas vcrpy supports more), is that right?  None of this is to say you need to change that here, I'm just interested in starting a conversation about it.

importer/api/__init__.py
```diff
@@ -124,6 +125,7 @@ def import_course(bundle, repo_id, user_id, static_dir):
     )
     import_static_assets(course, static_dir)
     import_children(course, src, None)
+    populate_xanalytics_fields.delay(course.id)
```
carsongee, line 132: Isn't this already inside an asynchronous task, why are we asyncing on top of asyncing, ouch my brain ;)

importer/tasks.py
```diff
@@ -3,11 +3,55 @@
...
+def populate_xanalytics_fields(course_id):
+    """
+    Initiate request to xanalytics API to get stats for a course,
+    then trigger async job to retrieve results when they become available.
+    Args:
+        course_id (int): primary key of a Course
+    """
+    if settings.XANALYTICS_URL != "":
+        token = send_request(settings.XANALYTICS_URL + "/create", course_id)
+        check_for_results.apply_async(
```
carsongee, line 38: oh, and then we are asyncing again, so three levels of async calls

ShawnMilo:
I chose it based on the blog post in the pull request message. I looked at both options, and based on the activity being more recent, the project being smaller, and supported by a company instead of an individual, I chose responses.

I'm not married to it, but I really like it.

importer/api/__init__.py
ShawnMilo, line 132: Yes, because even if we trap all "expected" errors when accessing an external API (such as timeouts), it could fail in a new and wonderful way. That shouldn't break the course import.

importer/tasks.py
ShawnMilo, line 38: Yes, because of the back-off retry delay.

importer/api/__init__.py
carsongee, line 132: You can just catch those exceptions here with either a custom exception class you raise in the called functions, or the set of exception classes that will occur (`requests.exceptions.RequestException`, and `ValueError` for json stuff), log the exception, and move on since we don't have any reason to be in a rush anyway since we are already offloaded from the Web thread.  I don't see how we could raise an exception we can't plan for and should be checking for individually anyway.  Layering async on async on async is going to make debugging extremely difficult.

importer/tasks.py
carsongee, line 38: Why not just use the request builtin mechanism for retries: i.e. use session and do `session.mount(urlbase, HTTPAdapter(max_retries=self.RETRIES))` and just handle `requests.RequestException`, it would simplify this code quite a bit I think, and should mean we can unwrap another layer of async calls.

importer/api/__init__.py
ShawnMilo, line 132: There is no layering of async on async; there's a single queue. Throwing a task onto the queue from a worker doesn't involve any nesting or recursion. Given that, I don't see any functional difference between the current version and the suggested version, and the current version is simpler (less code).

I think it's a bad idea to list out all "expected" exception types for the reason I stated above. A custom exception would work, but would require a "bare Except" to ensure everything was caught.

I think the main difference is that you think there's something inherently wrong about scheduling a Celery task from within a Celery task, and I don't. Unless you have a reason it's bad that I'm not seeing, I'd prefer to leave this as it is.

importer/tasks.py
ShawnMilo, line 38: I don't see anything in the `requests` docs for using a backoff. Am I missing it?

carsongee, line 38: oh, should have read further, this is waiting for a task to complete, will the other app be okay with us firing multiple jobs off at the same time?

importer/tasks.py
```diff
@@ -3,11 +3,55 @@
...
+@async.task
+def populate_xanalytics_fields(course_id):
+    """
+    Initiate request to xanalytics API to get stats for a course,
+    then trigger async job to retrieve results when they become available.
+    Args:
+        course_id (int): primary key of a Course
+    """
+    if settings.XANALYTICS_URL != "":
+        token = send_request(settings.XANALYTICS_URL + "/create", course_id)
```
carsongee, line 37: I'm sure this is probably out of scope if they have already started the API, but why do the RESTful way of doing jobs on the remote end by returning a 202 with a location header on where to check for status, and then the status page on the remote end returns a 200 until it is done ala http://restcookbook.com/Resources/asynchroneous-operations/ or https://www.adayinthelifeof.nl/2011/06/02/asynchronous-operations-in-rest/

importer/tasks.py
ShawnMilo, line 38: If you mean multiple calls to the `/create` endpoint (one per course), we'd have to clarify that with Ike. Since they're low-priority, asynchronous jobs on his end (as discussed in an e-mail thread), and return a token to check the status, I assume it's okay to kick off more than one simultaneously. But I'll check.

importer/tasks.py
ShawnMilo, line 37: Yeah, that depends on how they implement it. If they choose to do that we can adapt. I'll mention it while I'm asking about the multiple concurrent course requests.

importer/tasks.py
```diff
@@ -3,11 +3,55 @@
...
+def check_for_results(token, wait, attempt):
+    """
+    Check for xanalytics results for a course.
+    Args:
+        token (string): Token received from xanalytics server.
+        wait (int): Seconds to wait before repository.
+        attempt (int): Attempt number, so we don't try forever.
+    """
+    resp = get_result(settings.XANALYTICS_URL + "/status", token)
+    if resp["status"] == "still busy" and attempt < 10:
```
carsongee: Since we are async why not wait based on  status code and potentially forever?  At the very least the retry should be an all caps variable at the top of the file so it is easy to find, adjust, and should make testing easier since you can mock it to be whatever you need.

xanalytics/tests/test_api.py
```diff
@@ -0,0 +1,87 @@
...
+    def test_get_result(self):
+        """Check status of request."""
+        reply = {"status": "success", "url": "http://example.com/foo.json"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        resp = get_result(settings.XANALYTICS_URL + "/status", 1234)
```
carsongee: this is only called once and is just one line, why not inline it?

xanalytics/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_xanalytics_file(url):
+    """
+    Retrieve file and save locally.
+    Args:
+        URL (unicode): URL to file location.
+    Returns:
+        filename (string): Path to file on local filesystem.
+    """
+    resp = requests.get(url)
```
carsongee: Why are we making this a file at all?  We delete it as soon as we process it, and the requests.get means it is already going to be completely stored in memory, so we aren't saving mem at all, seems to me we should just get rid of this function, do the request above and have the update function work off the in memory structure.

lore/settings.py
```diff
@@ -348,3 +348,4 @@ def get_var(name, default):
     }
 }
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
+XANALYTICS_URL = get_var('XANALYTICS_URL', "")
```
carsongee, line 357: Have we pinned down the auth scheme at all?  This will be fine if we end up using basic auth, but if we need token auth, we are going to need a second var to store the security token in it.

importer/tasks.py
ShawnMilo: I think the "potentially forever" thing would be a problem; we would never want that to happen. It _should_ log if it gives up, though. Preferably sending an e-mail alert. If it went on forever it would increasingly tax the remote API, and also just consume resources on our servers. Also, this is something we could easily trigger from outside the import task, so it would be easy to "backfill" these once deployed, and also easy to make another attempt if we were alerted it had failed.

The retry count being a "constant" is definitely the way to go here.

xanalytics/tests/test_api.py
ShawnMilo: Do you mean rolling 55 and 56 into one line and not making `resp` a variable? If so, that would work. It's probably that way because, during development, I was doing some logging there.

learningresources/api.py
```diff
@@ -301,3 +302,20 @@ def import_static_assets(course, path):
...
+
+
+def update_xanalytics(filename):
+    """
+    Update xanalytics fields for a LearningResource.
+    Args:
+        filename (string): filename of JSON file from xanalytics
+    """
+    with open(filename) as raw:
+        data = json.load(raw)
```
carsongee: you may want to do some exception handling here, or at least in the calling function.  At the very least you should have a Raises section in the doc string that includes IOError and ValueError

xanalytics/__init__.py
ShawnMilo: Yeah, I think you're right. I doubt it would be big enough for this to be a problem. Hopefully those aren't famous last words.

lore/settings.py
ShawnMilo, line 357: I don't know yet. We haven't received any details of the other end. I assume there will be tweaking necessary in any case once the API is released.

learningresources/api.py
ShawnMilo: If we're getting rid of the file, then this will already be getting a valid dict. But yeah, there should be exception handling and logging in the part where we retrieve and parse the JSON from the API.

importer/tests/test_xanalytics.py
```diff
@@ -0,0 +1,89 @@
...
+import responses
+
+from importer.tasks import populate_xanalytics_fields, check_for_results
+from learningresources.tests.base import LoreTestCase
+
+log = logging.getLogger(__name__)
+# pylint: disable=no-self-use
+
+
+class TestAPI(LoreTestCase):
```
carsongee, line 20: There aren't any unit tests for xanalytics/send_request, get_result, get_xanalytics_file, or learningresources/api.py:update_xanalytics, just integration level tests for check_for_results and populate_xanalytics_fields, they should be added

carsongee, line 20: Or pointed out to me if I missed them (more than possible)

importer/tests/test_xanalytics.py
```diff
@@ -0,0 +1,89 @@
...
+        reply = {"status": "still busy"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        check_for_results("abcde", 10, 1)
+
+    @responses.activate
```
carsongee, line 67: So this API seems identical to httppretty, just swap `@responses.activate` for `@httpretty.activate,` `responses.add` with `httpretty.register_uri`, and `responses.GET` with `httpretty.GET` and I think all of these would work without any other changes.

importer/tests/test_xanalytics.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        reply = {"status": "complete", "url": "http://example.com/foo.json"}
+        responses.add(
+            responses.POST,
+            'http://example.com/status',
+            body=json.dumps(reply),
+            content_type="application/json"
+        )
+        check_for_results("abcde", 10, 1)
```
carsongee, line 100: I would like to see one vcrpy test here as a full integration test that we could replace with live server data when the API becomes available (and fake it for now).

importer/tests/test_xanalytics.py
ShawnMilo, line 20: I think you missed them a little bit, although you did leave code review notes in that file. :o)

https://github.com/mitodl/lore/blob/feature/skm/166_xanalytics/xanalytics/tests/test_api.py

importer/tests/test_xanalytics.py
ShawnMilo, line 100: As I understand it, we can't create a vcr test unless we have a server to pull from. So that either means waiting for the server to happen, or making our own. We can't make our own because we don't know how the real one is going to work until we get it. Or am I misunderstanding?

importer/tests/test_xanalytics.py
ShawnMilo, line 67: @carsongee: I made the other changes, but when I tried to swap in httpretty, my Docker environment won't build. Please give it a try in case there's just something screwy locally:

Here's a [patch](http://hastebin.com/raw/lurasateqo) to do all the responses -> httpretty replacements.

importer/tasks.py
ShawnMilo, line 38: Yes, multiple simultaneous calls are okay. Ike confirmed this via e-mail.

importer/tasks.py
ShawnMilo, line 37: Ike has replied to this question. It will probably be RESTful. Development has not begun.

pwilkins:
@ShawnMilo How do I test this from the UI?  How can I tell if it's working?

ShawnMilo:
Right now you can't test it from the UI because the API is unavailable. We'd have to write an API and put the URL in the LORE settings to test it in the UI.

I contacted Ike earlier in the week for a status update; I'll post in this thread when there's an update. For now we just have the test suite to judge by.

ShawnMilo:
@pdpinch @carsongee: The merge conflict has been fixed and the tests have passed.

PR #404, opened 2015-07-22
Refactored code for reloading module into a function

noisecapella:


carsongee:
Why not just `from six.moves import reload_module`?

noisecapella:
I didn't know about that. I've updated the PR for that function

carsongee:
:+1:

PR #405, opened 2015-07-23
Added description path to listing page

### LINKED ISSUE #388: Display path on search results view
As a user, I would like to see the context for the learning resources in my search results. 

Ideally, the items in the path would be links to open the learning resource panel for the corresponding container resource. 

May need to add path field (in model) to elastic search index.

giocalitri:
Implements the path as sequence of titles of the learning resources " / " separated.
An important part of this PR is the migration that populates the new field of the learningresource model.

Closes #388 
[#99424610]

### DIFF
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -79,7 +79,8 @@ def create_course(org, repo_id, course_number, run, user_id):
 
 # pylint: disable=too-many-arguments
 def create_resource(
-        course, parent, resource_type, title, content_xml, mpath, url_name
+        course, parent, resource_type, title, content_xml, mpath, url_name,
+        dpath
 ):
     """
     Create a learning resource.
@@ -93,6 +94,8 @@ def create_resource(
         content_xml (unicode): XML
         mpath (unicode): Materialized path
         url_name (unicode): Resource identifier
+        dpath (unicode): Description path
+
     Returns:
         resource (learningresources.models.LearningResource):
             New LearningResource
@@ -104,6 +107,7 @@ def create_resource(
         "content_xml": content_xml,
         "materialized_path": mpath,
         "url_name": url_name,
+        "description_path": dpath,
     }
     if parent is not None:
         params["parent_id"] = parent.id
@@ -301,3 +305,41 @@ def import_static_assets(course, path):
                 name = join(root, name).replace(path + sep, '', 1)
                 django_file.name = name
                 create_static_asset(course.id, django_file)
+
+
+def join_description_paths(*args):
+    """
+    Helper function to format the description path.
+    Args:
+        args (unicode): description path
+    Returns:
+        unicode: Formatted dpath
+    """
+    return ' / '.join([dpath for dpath in args if dpath != ''])
+
+
+def update_description_path(resource, force_parent_update=False):
+    """
+    Updates the specified learning resource description path
+    based on the current title and the parent's description path
+    Args:
+        resource (learningresources.models.LearningResource): LearningResource
+        force_parent_update (boolean): force parent update
+    Returns:
+        None
+    """
+    description_path = ''
+    if resource.parent is None:
+        description_path = join_description_paths(resource.title)
+    else:
+        # if the parent doesn't have a description_path update first the parent
+        if resource.parent.description_path == '' or force_parent_update:
+            update_description_path(resource.parent, force_parent_update)
+        # the current description path is
+        # the parent's one plus the current title
+        description_path = join_description_paths(
+            resource.parent.description_path,
+            resource.title
+        )
+    resource.description_path = description_path
+    resource.save()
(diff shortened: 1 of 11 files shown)


learningresources/api.py
```diff
@@ -301,3 +305,44 @@ def import_static_assets(course, path):
...
                 create_static_asset(course.id, django_file)
+
+
+def join_description_paths(*args):
+    """
+    Helper function to format the description path.
+    Args:
+        list of dpath (unicode): description path
+    Returns:
+        formatted dpath (unicode)
```
noisecapella: For the argument you should use the format `args (type here): description path`. Whatever you think is sensible for the type is fine with me

The return section should be `unicode: Formatted dpath`

learningresources/management/commands/update_description_paths.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Command(BaseCommand):
+    """
+    Command for sync_permissions
+    """
+    help = "Synchronizes and updates the description_path"
+
+    def handle(self, *args, **options):
+        """Command handler"""
+        for learning_resourse in LearningResource.objects.all():
```
noisecapella: Change to `learning_resource`

learningresources/api.py
```diff
@@ -301,3 +305,44 @@ def import_static_assets(course, path):
...
+    """
+    formatted_dpath = join(
+        *[' {0} '.format(dpath) for dpath in args if dpath != '']
+    )
+    return formatted_dpath.strip()
+
+
+def update_description_path(resource, force_parent_update=False):
```
noisecapella, line 321: Would it make sense to move this function to a method in `LearningResource`? The only required argument is the `resource` model, it doesn't return anything and mutates the `LearningResource`

noisecapella, line 321: Oh, it looks like that may interfere with your migration

learningresources/tests/test_api.py
```diff
@@ -190,5 +184,105 @@ def test_create_static_asset(self):
...
+    def test_update_description_path(self):
+        """Tests for update_description_path"""
+        # after created a resource without parent has the description path
+        # equal to the title
+        self.assertIsNone(self.resource.parent)
+        self.assertEqual(
+            self.resource.title,
+            self.resource.description_path
+        )
+        # changing the title does not update the descriptio path automatically
```
noisecapella: Typo in `descriptio`

learningresources/models.py
```diff
@@ -125,6 +125,7 @@ class LearningResource(BaseModel):
     description = models.TextField(blank=True)
     content_xml = models.TextField()
     materialized_path = models.TextField()
+    description_path = models.TextField(blank=True)
```
noisecapella, line 128: Why are we making this a field and not just creating this value on the fly?

learningresources/tests/test_api.py
```diff
@@ -190,5 +184,105 @@ def test_create_static_asset(self):
...
+        )
+        # create a child resource
+        child_res = self.create_resource(
+            parent=self.resource
+        )
+        # the description path is the combination of the child resource title
+        # and the parent description path
+        self.assertEqual(
+            child_res.description_path,
+            api.join_description_paths(
```
noisecapella, line 248: In any of the invocations of `join_description_paths` are we passing in an empty string? If not we should assert that the empty string is skipped

giocalitri, line 248: no, we are almost never passing an empty string, but I added tests anyway and I changed the way I join the strings

learningresources/api.py
```diff
@@ -301,3 +305,41 @@ def import_static_assets(course, path):
...
+    """
+    Updates the specified learning resource description path
+    based on the current title and the parent's description path
+    Args:
+        resource (learningresources.models.LearningResource): LearningResource
+        force_parent_update (boolean): force parent update
+    Returns:
+        None
+    """
+    description_path = ''
```
noisecapella, line 331: Since we are setting this in both paths we can remove this line

learningresources/tests/base.py
```diff
@@ -61,17 +66,19 @@ def create_resource(self, **kwargs):
...
+            content_xml=kwargs.get('content_xml', "<blah>other blah</blah>"),
+            mpath=kwargs.get('mpath', "/otherblah"),
+            url_name=kwargs.get('url_name'),
+            dpath=''
         )
         learn_res.xa_nr_views = kwargs.get('xa_nr_views', 0)
         learn_res.xa_nr_attempts = kwargs.get('xa_nr_attempts', 0)
         learn_res.xa_avg_grade = kwargs.get('xa_avg_grade', 0)
         learn_res.save()
+        update_description_path(learn_res)
```
noisecapella, line 81: I think this means we have two saves happening unnecessarily. No strong opinion on this though, it's probably not often called

PR #406, opened 2015-07-23
Removed export view which isn't used anymore

### LINKED ISSUE #408: Remove export view
The export view was replaced with the REST api (`content_xml` within repositories/repo/learning_resources/#/) so we should remove it.

noisecapella:
Fixes #408

### DIFF
--- a/ui/templates/includes/learning_resource.html
+++ b/ui/templates/includes/learning_resource.html
@@ -26,7 +26,7 @@
       </div>
       <div class="tile-content">
         <h2>
-          <a href="{% url 'export' repo.slug result.object.id %}" class="cd-btn" data-learningresource-id="{{ result.object.id }}">
+          <a href="#" class="cd-btn" data-learningresource-id="{{ result.object.id }}">
             {% if result.object.title %}
             {{ result.object.title }}
             {% else %}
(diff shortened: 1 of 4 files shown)


giocalitri:
Following what @pdpinch said yesterday, you should also open an issue for what you are solving here

giocalitri:
other than that, if the tests pass, it can be merged

noisecapella:
Ok, I added an issue and the 'Fixes' thing above

ui/templates/includes/learning_resource.html
```diff
@@ -26,7 +26,7 @@
       </div>
       <div class="tile-content">
         <h2>
-          <a href="{% url 'export' repo.slug result.object.id %}" class="cd-btn" data-learningresource-id="{{ result.object.id }}">
+          <a href="#" class="cd-btn" data-learningresource-id="{{ result.object.id }}">
```
giocalitri, line 29: not sure it's a good practice, but for debugging reasons (at least) it would be a good idea to have the learning resource id somewhere, so what about having the `href="#<lr_id>"`?

noisecapella, line 29: It's stored in the `data-learningresource-id` attribute

giocalitri:
The functionality is intact so looks good to me

Issue #387, opened 2015-07-20
Export: builds gzip'ed file containing XML and static assets


pdpinch:
```
 go for the simplest possible solution: one XML file for each learning resource, using the url_name
```
- [x] export is an asynchronous task
- [x] user needs a way to know the status of their export
- [x] use Amazon temporary S3 URLs to limit access
- [x] need way to manage collisions of learning resources. Can we use database ids? asset ids
- [x] need way to manage collisions of static assets. 
    This is a little harder than XML collisions because the XML has references to the static files. For now, we will got with last asset wins.

pdpinch:
XML should be organized into folders, with a folder for each XBlock type, e.g. problem, video, vertical, sequential, chapters -- in order to match the structure of OLX. Static files should go into a folder names /static

noisecapella:
Can we change this to first asset wins instead of last asset wins? It makes the coding slightly easier. Current plan is to have first asset keep its name and all other assets append a `_1`, `_2` etc before the extension.

Is it ok if resource XML files are formatted `{id}_{name}.xml`, or is `url_name` very much desired? What if the `url_name` doesn't exist?

ShawnMilo:
Considering that the export order is essentially random, I don't see any reason why not.

pdpinch:
- last asset wins is just as good.
- is {name} equivalent to the display_name attribute in the XML or url_name? I would prefer the later, i.e. {id}_{url_name} with a failover to just {id} when there is no url_name. 
  
  Display names make poor file names (spaces, etc.) and they are more likely to collide (not an issue with the {id} but still confusing for humans)

pdpinch:
This looks done to me. We will probably have some refinements to the ui/ux but those can go into new stories.

PR #407, opened 2015-07-23
Added export functionality for learning resources

noisecapella:
This is a piece of #387 but certain requirements of it are not fulfilled yet in this PR

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,7 @@ web:
     LORE_ADMIN_EMAIL:
     LORE_CAS_URL:
     CELERY_ALWAYS_EAGER: False
+    CELERY_RESULT_BACKEND: redis://redis:6379/4
     BROKER_URL: redis://redis:6379/4
     HAYSTACK_URL: elastic:9200
   ports:
--- a/exporter/api.py
+++ b/exporter/api.py
@@ -1,0 +1,84 @@
+"""
+Functions for exporting learning resources
+"""
+
+from __future__ import unicode_literals
+
+from tempfile import mkdtemp, mkstemp
+import tarfile
+import os
+from shutil import rmtree
+
+from django.utils.text import slugify
+from django.core.files.storage import default_storage
+from lore.settings import EXPORT_PATH_PREFIX
+
+
+def export_resources_to_directory(learning_resources):
+    """
+    Create files of LearningResource and StaticAsset contents inside directory.
+
+    Args:
+        learning_resources (list of learningresources.models.LearningResource):
+            LearningResources to export in tarball
+    Returns:
+        unicode: Newly created temp directory with files inside of it.
+    """
+    def sanitize(title):
+        """Sanitize title for use in filename."""
+        # Limit filename to 200 characters since limit is 256
+        # and we have a number and extension too.
+        return slugify(title)[:200]
+
+    tempdir = mkdtemp()
+    try:
+        for learning_resource in learning_resources:
+            filename = "{id}_{title}.xml".format(
+                id=learning_resource.id,
+                title=sanitize(learning_resource.title),
+            )
+            with open(os.path.join(tempdir, filename), 'w') as f:
+                f.write(learning_resource.content_xml)
+    except:
+        # Clean up temporary directory since we can't return it anymore.
+        rmtree(tempdir)
+        raise
+    return tempdir
+
+
+def export_resources_to_tarball(learning_resources, username):
+    """
+    Create tarball and put LearningResource and StaticAsset contents in it.
+
+    Args:
+        learning_resources (list of learningresources.models.LearningResource):
+            LearningResources to export in tarball
+        username (unicode): Name of user
+    Returns:
+        unicode: path of tarball within django-storage
+    """
+    handle, tempfilepath = mkstemp()
+    os.close(handle)
+    try:
+        archive = tarfile.open(tempfilepath, mode='w:gz')
+
+        tempdir = export_resources_to_directory(learning_resources)
+        try:
+            for name in os.listdir(tempdir):
+                abs_path = os.path.join(tempdir, name)
+                if os.path.isfile(abs_path):
+                    archive.add(abs_path, arcname=name)
+
+            archive.close()
+            output_path = '{prefix}{username}_exports.tar.gz'.format(
+                prefix=EXPORT_PATH_PREFIX,
+                username=username
+            )
+
+            # Remove any old paths.
+            default_storage.delete(output_path)
+            return default_storage.save(output_path, open(tempfilepath, 'rb'))
+        finally:
+            rmtree(tempdir)
+    finally:
+        os.remove(tempfilepath)
--- a/exporter/tests/test_export.py
+++ b/exporter/tests/test_export.py
@@ -1,0 +1,127 @@
+"""
+Tests for export
+"""
+
+from __future__ import unicode_literals
+from tempfile import mkdtemp
+from shutil import rmtree
+import os
+
+from archive import Archive
+from django.core.files.storage import default_storage
+from django.utils.text import slugify
+
+from learningresources.tests.base import LoreTestCase
+from learningresources.models import (
+    LearningResource,
+    LearningResourceType,
+)
+from learningresources.api import update_description_path
+from exporter.api import (
+    export_resources_to_directory,
+    export_resources_to_tarball,
+)
+from exporter.tasks import export_resources
+from learningresources.api import create_resource
+from importer.tasks import import_file
+
+
+class TestExport(LoreTestCase):
+    """
+    Tests for export
+    """
+
+    def setUp(self):
+        super(TestExport, self).setUp()
+
+        # Add some LearningResources on top of the default to make things
+        # interesting.
+        tarball_file = self.get_course_single_tarball()
+        import_file(
+            tarball_file, self.repo.id, self.user.id)
+
+        # Add a resource with a '/' in the title and too many characters.
+        resource = create_resource(
+            course=self.repo.course_set.first(),
+            resource_type=LearningResourceType.objects.first().name,
+            title="//x"*300,
+            content_xml="",
+            mpath="",
+            url_name=None,
+            parent=None,
+            dpath=''
+        )
+        update_description_path(resource)
+
+    def assert_resource_directory(self, resources, tempdir):
+        """Assert that files are present with correct content."""
+        def make_name(resource):
+            """Format expected filename."""
+            return "{id}_{title}.xml".format(
+                id=resource.id,
+                title=slugify(resource.title)[:200],
+            )
+
+        for resource in resources:
+            abs_path = os.path.join(tempdir, make_name(resource))
+            self.assertTrue(os.path.exists(abs_path))
+            with open(abs_path) as f:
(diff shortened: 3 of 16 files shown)


ui/views.py
```diff
@@ -261,12 +261,13 @@ def build_form(self, form_kwargs=None):
...
 
 @login_required
-def serve_media(request, media_path):
+def serve_static_assets(request, path):
     """
     View to serve media files in case settings.DEFAULT_FILE_STORAGE
     is django.core.files.storage.FileSystemStorage
     """
     # first check if the user has access to the file
+    media_path = os.path.join('assets', path)
```
ShawnMilo: Where does this hard-coded `'assets`' come from?

ui/views.py
```diff
@@ -280,3 +281,33 @@ def serve_media(request, media_path):
     response['Content-Length'] = os.path.getsize(file_path)
     response['Content-Disposition'] = "attachment; filename=%s" % filename
     return response
+
+
+@login_required
```
ShawnMilo, line 290: Does this have (or need) any additional permissions beyond login, to prevent users from downloading others' exports?

ui/views.py
```diff
@@ -280,3 +281,33 @@ def serve_media(request, media_path):
...
+def serve_resource_exports(request, path):
+    """
+    View to serve media files in case settings.DEFAULT_FILE_STORAGE
+    is django.core.files.storage.FileSystemStorage
+    """
+    media_path = os.path.join(settings.EXPORT_PATH_PREFIX, path)
+    file_path = os.path.join(settings.MEDIA_ROOT, media_path)
+
+    # There is only one path that will work here, make sure it matches exactly.
```
ShawnMilo, line 299: Why are we hard-coding a specific path? This prevents a user from doing multiple exports from the same repository. Is it a technical limitation, or a design decision?

exporter/api.py
```diff
@@ -0,0 +1,74 @@
...
+from lore.settings import EXPORT_PATH_PREFIX
+
+
+def export_resources_to_directory(learning_resources, tempdir):
+    """
+    Create XML files of learning resource contents inside directory.
+
+    Args:
+        learning_resources (list of learningresources.models.LearningResource):
+            LearningResources to export in tarball
```
ShawnMilo, line 23: `tempdir` is missing from the docstring.

However, I think this function should create the tempdir and return it. That prevents the multiple callers from having to call `mkdtemp` themselves.

exporter/api.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    def sanitize(title):
```
ShawnMilo, line 27: There are multiple nested `sanitize` functions. Why not use `slugify`? It handles characters other than `/` by default, and also covers whitespace.

exporter/tests/test_export.py
```diff
@@ -0,0 +1,133 @@
...
+            content_xml="",
+            mpath="",
+            url_name=None,
+            parent=None,
+            dpath=''
+        )
+        update_description_path(resource)
+
+    def assert_resource_directory(self, resources, tempdir):
+        """Assert that files are present with correct content."""
```
ShawnMilo, line 57: Missing args, etc. Same for `make_name` below. This might be unimportant, though.

test_requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Core project/common application testing requirements
 
+django-celery
```
ShawnMilo: I don't think this is required. I'm running the test suite locally without it and tests are passing.

The [django-celery](https://pypi.python.org/pypi/django-celery) docs say it's deprecated, and the [celery docs](http://celery.readthedocs.org/en/latest/django/first-steps-with-django.html) say you only need it if you want to use the Django ORM as the results backend, but we're using Redis as the results backend.

docker-compose.yml
```diff
@@ -33,6 +33,7 @@ web:
     LORE_ADMIN_EMAIL:
     LORE_CAS_URL:
     CELERY_ALWAYS_EAGER: False
+    CELERY_RESULT_BACKEND: redis://redis:6379/4
```
ShawnMilo, line 36: Are these results being purged or expiring automatically?

ui/views.py
```diff
@@ -280,3 +281,33 @@ def serve_media(request, media_path):
...
+    if not os.path.exists(file_path):
+        raise Http404()
+
+    filename = os.path.basename(file_path)
+    response = HttpResponse(
+        FileWrapper(open(file_path, 'rb')),
+        content_type=mimetypes.guess_type(file_path)[0]
+    )
+    response['Content-Length'] = os.path.getsize(file_path)
+    response['Content-Disposition'] = "attachment; filename=%s" % filename
```
ShawnMilo: Why use `%` interpolation instead of `.format()` here?

ui/views.py
noisecapella: I suppose from STATIC_ASSET_BASEPATH in `learningresources/models.py` originally. I'll separate out the 'assets' prefix and reference it here and elsewhere I use it

rest/tests/test_export.py
```diff
@@ -0,0 +1,72 @@
...
+
+class TestExport(RESTTestCase):
+    """Test for export."""
+
+    @override_settings(
+        DEFAULT_FILE_STORAGE='django.core.files.storage.FileSystemStorage'
+    )
+    def test_create_new_task(self):
+        """Test a basic export."""
+        resource = self.repo.course_set.first().learningresource_set.first()
```
ShawnMilo: Why not just use `self.resource`?

rest/tests/test_export.py
```diff
@@ -0,0 +1,72 @@
...
+        self.create_learning_resource_export(self.repo.slug, {
+            "id": resource.id
+        })
+
+        task_info = self.create_learning_resource_export_task(self.repo.slug)
```
ShawnMilo: The variable `task_info` isn't really necessary. It's only used once.

rest/tests/test_export.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertEqual("success", result['status'])
+        self.assertTrue(result['url'].startswith(
+            "/media/resource_exports/test_exports.tar"))
+
+        with self.settings(
+            DEFAULT_FILE_STORAGE='storages.backends.s3boto.S3BotoStorage'
+        ):
+            # change the default file storage to S3
+            reload_module(ui.urls)
+            # the view is not available any more
```
ShawnMilo, line 50: The comments here are not sentences. Also, "any more" should be "anymore" in this case.

exporter/api.py
```diff
@@ -0,0 +1,74 @@
...
+        output_path = '{prefix}{username}_exports.tar.gz'.format(
+            prefix=EXPORT_PATH_PREFIX,
+            username=username
+        )
+
+        # Remove any old paths.
+        default_storage.delete(output_path)
+        return default_storage.save(output_path, open(tempfilepath, 'rb'))
+    finally:
+        os.unlink(tempfilepath)
```
ShawnMilo: `os.remove` and `os.unlink` are synonyms, so it seems to always make more sense to use `remove`, as it's more understandable.

ui/views.py
noisecapella: I copied this from the above function, but I can make this change here

rest/views.py
```diff
@@ -593,3 +599,145 @@ def perform_destroy(self, instance):
...
+    task_id = task['id']
+
+    state = "processing"
+    url = ""
+    # initial_state is a workaround for EagerResult used in testing.
+    # In production initial_state should usually be pending.
+    if initial_state == SUCCESS:
+        state = "success"
+        url = task['url']
+    elif initial_state == FAILURE or initial_state == REVOKED:
```
ShawnMilo: This could be:

``` python
elif initial_state in (FAILURE, REVOKED):
```

rest/views.py
```diff
@@ -593,3 +599,145 @@ def perform_destroy(self, instance):
...
+    def get_object(self):
+        """
+        Retrieve current information about an export task.
+        """
+        try:
+            task_id = self.kwargs['task_id']
+        except ValueError:
+            raise Http404
+
+        repo_slug = self.kwargs['repo_slug']
```
ShawnMilo: This variable isn't really necessary, since it's only used once, and wouldn't make the line too long.

ui/views.py
noisecapella, line 290: It matches the filename to the user inside the function to restrict users downloading others exports

ui/views.py
noisecapella, line 299: I talked with @carsongee and it seems like we don't need to keep more than one copy at a time. It does overwrite the user's previous exports but these are temporary files anyway. What do you think?

ShawnMilo, line 299: I would expect to be able to not only have multiple exports, but go back a week later and re-download an old export.

However, it's fine if it's intentional for practical reasons. My only worry is that someone will eventually dictate that we have to support multiple downloads (or access to the download history), and some of this design will make that harder than it must be.

It's probably YAGNI for now if we don't yet _know_ we're going to need it.

test_requirements.txt
noisecapella: At one point I was trying to set up async Celery testing which I think required that library, however I went a different route and this can be removed

exporter/tests/test_export.py
noisecapella, line 57: Since these functions are only used in testing I'm going to skip over these

exporter/api.py
noisecapella, line 27: We still need to truncate the filename below the file system limit. Does slugify do that?

ShawnMilo, line 27: Slugify doesn't truncate, but you can always keep the `[:200]`.

exporter/tests/test_export.py
ShawnMilo, line 57: ACK

test_requirements.txt
ShawnMilo: I'm trying it locally. I had an error when exporting, so I'm rebuilding my whole environment now. If it works, I'll remove that requirement and rebuild it again to try.

The reason I haven't changed the label on this PR is that I have a couple of things I want to check with the actual tarball.

docker-compose.yml
noisecapella, line 36: I'm not sure how long these results last. @carsongee ?

This was added to allow our docker containers to keep track of Celery tasks so that the front end can watch their progress.

ui/views.py
noisecapella, line 299: Yeah, I think YAGNI seems like the best course here

ShawnMilo, line 299: ACK

rest/tests/test_export.py
noisecapella, line 50: This is copied/pasted from another test elsewhere. In general I don't really care about strict grammar and capitalization, at least where the end user won't ever see it, so I'll leave this alone

docker-compose.yml
noisecapella, line 36: 1 day by default: from `CELERY_TASK_RESULT_EXPIRES` from [here](http://celery.readthedocs.org/en/latest/configuration.html)

ShawnMilo, line 36: Cool.

PR #417, opened 2015-07-27
Avoided hitting the database for the search page.

### LINKED ISSUE #402: Make the listing page elasticsearch powered ONLY
Make sure to avoid hitting the DB.

Bonus points if we enforce this with exceptions if in the templates someone writes `result.object.<model_field>` that will force Django to hit the DB anyway.

References:
https://django-haystack.readthedocs.org/en/v2.4.0/best_practices.html#avoid-hitting-the-database
https://www.dominicrodger.com/order-n-haystack.html

ShawnMilo:
Load data elements from LearningResources into the Elasticsearch index,
so they can be inserted into the template from the query results, not
requiring a DB lookup.

Also added a test which should catch regressions. There are still queries
due to the checks to authentication and session management, but there
are no queries to learningresources_learningresource.

Closes #402.

Used documentation from [Haystack](https://django-haystack.readthedocs.org/en/v2.4.0/best_practices.html#avoid-hitting-the-database) and [a blog](https://www.dominicrodger.com/order-n-haystack.html) as suggested in #402. In testing, the `load_all` argument to the view didn't make a difference as suggested in the Haystack documentation, but I left it in there anyway just in case it helps an edge case.

### DIFF
--- a/ui/templates/includes/learning_resource.html
+++ b/ui/templates/includes/learning_resource.html
@@ -26,36 +26,36 @@
       </div>
       <div class="tile-content">
         <h2>
-          <a href="#" class="cd-btn" data-learningresource-id="{{ result.object.id }}">
-            {% if result.object.title %}
-            {{ result.object.title }}
+          <a href="#" class="cd-btn" data-learningresource-id="{{ result.lid }}">
+            {% if result.title %}
+            {{ result.title }}
             {% else %}
             No Title
             {% endif %}
           </a>
         </h2>
-        
+
         <div class="tile-meta">
-          <span class="meta-item">{{ result.object.description_path }}</span>
+          <span class="meta-item">{{ result.description_path }}</span>
         </div>
         <div class="tile-blurb">
-          {% if result.object.description %}
-          {{ result.object.description }}
+          {% if result.description %}
+          {{ result.description }}
           {% else %}
           No description provided.
           {% endif %}
         </div>
         <div class="tile-meta">
-          <span class="meta-item">{{ result.object.course.course_number }}</span>
-          <span class="meta-item">{{ result.object.course.run }}</span>
-          <span class="meta-item"><a href="{{ result.object.get_preview_url }}" target="_blank">Preview</a></span>
+          <span class="meta-item">{{ result.course }}</span>
+          <span class="meta-item">{{ result.run }}</span>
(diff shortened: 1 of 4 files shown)


Ferdi:
Nice.  Is there a simple way to check if we have any performance gain here. It would be great to know.

ShawnMilo:
Evidently all learning resources are pulled in a single query by Haystack, so it may only be saving us one query. It does allow us to throw whatever we want into the index, so things that refer to related objects could be more efficient. It does bloat the Elasticsearch database and increase its storage requirements, though.

Also, since all the usual checks are happening against auth and the session table, a simple load of the search page takes 10 queries. We could improve that by using another session backend, such as Redis.

ui/templates/includes/learning_resource.html
```diff
@@ -26,36 +26,36 @@
...
+          <span class="meta-item">{{ result.run }}</span>
+          <span class="meta-item"><a href="{{ result.preview_url }}" target="_blank">Preview</a></span>
         </div>
       </div>
     </div>
   </div>
   <div class="col-md-2 col-export">
-    <a class="link-export" href="#" data-learningresource-id="{{ result.object.id }}"
-       {% if result.object.id in exports %}
+    <a class="link-export" href="#" data-learningresource-id="{{ result.id }}"
```
noisecapella: Why is this `result.id` but above it's `result.lid`?

ShawnMilo: It's a mistake. It should always be `.lid`. Thanks for catching it.

I named the field in `LearningResourceIndex` `lid` instead of `id` because `id` is special. Then I replaced `result.object.id` with `result.id` by just deleting `object.`. This could have caused a subtle bug. Will fix.

noisecapella: Sounds good, also update the next line

noisecapella:
:+1: feel free to merge after tests pass

ShawnMilo:
Merged master into this branch to make tests pass (had the same `logiclab-common/asteroid` error).

ShawnMilo:
Tests passed, will squash, wait for Travis, then merge.

noisecapella:
It looks like the test is some random `npm install` failure, nothing to be concerned about. I'd repush just to be safe though, up to you

ShawnMilo:
Yeah, I merged the latest `master` in and re-pushed.

PR #420, opened 2015-07-27
Fix test failures due to pylint dependency chain

bdero:
logilab-common is only necessary because pylint's dependency chain
will result in the latest logilab-common being fetched, but astroid
(another pylint dependency) depends on logilab-common<=0.63.0

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -2,6 +2,10 @@
 
 pytest
 pytest-pep8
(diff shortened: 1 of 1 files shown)


giocalitri:
Looks good to me, will merge as soon as tests pass

carsongee:
That is strange, we were running on 1.0.1 and then 1.0.2 until now, did astroid change something?

PR #423, opened 2015-07-27
Fixed bug where export checkboxes were not updated in sync with export count

### LINKED ISSUE #422: Export links sometimes don't update export count

noisecapella:
Fixes #422

### DIFF
--- a/ui/static/ui/js/listing.js
+++ b/ui/static/ui/js/listing.js
@@ -265,11 +265,11 @@ define('listing',
       // Set up click handlers for export links.
       $(".link-export").click(function(event) {
         event.preventDefault();
+        var node = this;
+        var $node = $(node);
 
-        var learningResourceId = $(this).data("learningresource-id");
-        var selected = $(this).data("selected") === true;
-        $(this).data("selected", !selected);
-
+        var learningResourceId = $node.data("learningresource-id");
+        var selected = $node.data("selected") === true;
         if (selected) {
           // Remove an item from export cart.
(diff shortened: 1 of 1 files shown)


ShawnMilo:
Does this fix mean that removing an item from the cart failed, but we fix the UI anyway? Wouldn't that result in it still being there when they refreshed the page? What is the failure?

Also, don't lines 270 and 273 make the extra checks of `if ($(this).data("selected")` redundant?

Right at the top (after 270), why not just immediately call `updateExportCount()`? It looks like we want it called no matter what.

noisecapella:
I think you were missing that some of the code was in callbacks so we would need to read from the attribute again, `selected` might have been out of date.

In any case I've changed this to just update on successful AJAX calls to keep things simple and avoid UI quirks if DELETE and POST get out of order.

ShawnMilo:
This solution makes sense. Is it possible to write tests for this, or do we not have tests that execute JavaScript live on the page?

noisecapella:
We don't have tests that cover `listing.js` yet, it's not broken into its component pieces so it's very hard to debug.

Technically all our tests are live on PhantomJS's testing page but most tests operate on DOM elements which are disconnected from the window, or just verify changes in React state. At some point we need to refactor this, either by making the listing page use React or just breaking it into components in a better way, but since the tests weren't there to begin with I think it's out of the scope of this issue to do that work.

ShawnMilo:
LGTM and tested in browser locally.

PR #427, opened 2015-07-28
Output static assets in export

noisecapella:


exporter/api.py
```diff
@@ -0,0 +1,111 @@
...
+
+            for static_asset in learning_resource.static_assets.all():
+                # Output to static directory.
+                static_path = re.sub(r"^assets/", "static/",
+                                     static_asset.asset.name)
+
+                relative_path, filename = os.path.split(static_path)
+                try:
+                    os.makedirs(os.path.join(tempdir, relative_path))
+                except OSError:
```
ShawnMilo: `ex.args` should be checked to make sure it's actually failure to create a duplicate directory.

[example](https://github.com/mitodl/lore/blob/master/ui/views.py#L66) from our codebase

Working example of this use-case:

``` python
import os

try:
    os.mkdir("/tmp/butter")
    print "yay"
except OSError as ex:
    if "File exists" not in ex.args:
        raise ex
    print "It's okay, it's already there."
```

exporter/tests/test_export.py
```diff
@@ -0,0 +1,138 @@
...
+                self.assertEqual(f.read(), resource.content_xml)
+
+            # Verify static asset placement.
+            for static_asset in resource.static_assets.all():
+                relative_path = static_asset.asset.name.replace(
+                    "assets/", "static/")
+                abs_path = os.path.join(tempdir, relative_path)
+                self.assertTrue(os.path.exists(abs_path))
+                with open(abs_path, 'rb') as f:
+                    self.assertEqual(f.read(), static_asset.asset.read())
```
ShawnMilo: This optional, since the files in the test suite _should_ always be tiny, but it's technically better to not read in the files all at once. This _could_ cause problems running tests on restricted machines, like the Travis VMs.

Something like this could be used to limit the amount of memory used to read a file:

``` python
from hashlib import sha256

def get_hash(filename):
    """
    Get a file's hash without reading it all
    into memory at once.
    """
    read_size = 2**20
    signature = sha256()
    with open(filename, "rb") as handle:
        chunk = handle.read(read_size)
        while len(chunk) > 0:
            signature.update(chunk)
            chunk = handle.read(read_size)
    return signature.hexdigest()
```

ShawnMilo:
The only issues I found this morning were covered by your latest push; the pylint errors.

I just added one about the `OSError`.

exporter/api.py
noisecapella: Seems reasonable. It's unfortunate that `FileExistsError` only exists in Python 3

ShawnMilo:
Closed by mistake.

ShawnMilo:
Note: When flattening later, please change "outputted" to "output.'

exporter/tests/test_export.py
noisecapella: I'll keep it in mind. It looks like the test files are all in the kilobyte range or less so I'll leave this alone for now.

ShawnMilo: ACK

noisecapella:
@ShawnMilo This needs another look due to changes in filestructure and tests for collisions, everything new is in the second commit.

exporter/api.py
```diff
@@ -22,28 +23,78 @@ def export_resources_to_directory(learning_resources):
...
+                os.mkdir(os.path.join(tempdir, type_name))
+            except OSError as ex:
+                # Directory already exists
+                if "File exists" not in ex.args:
+                    raise
+            with open(os.path.join(tempdir, type_name, filename), 'w') as f:
                 f.write(learning_resource.content_xml)
+
+            for static_asset in learning_resource.static_assets.all():
+                try:
```
ShawnMilo: I think this `mkdir` attempt should happen above the outer loop -- before `for learning_resource in learning_resources:`. It's being done many times. I think it's even okay if you create a folder you never put anything in, but if you disagree you can check if `learningresources.filter(static_asset__id__insull=False)` above to find out if there are going to be any.

exporter/api.py
```diff
@@ -22,28 +23,78 @@ def export_resources_to_directory(learning_resources):
...
+                    """Convenience function for making a static path."""
+                    return os.path.join(tempdir, "static", name)
+
+                if os.path.exists(static_file_path(static_filename)):
+                    collision = True
+
+                    base, ext = os.path.splitext(static_filename)
+
+                    i = 1
+                    while True:
```
ShawnMilo: Minor style difference; I might make the loop `while os.path.exists(static_file_path(static_filename)):` instead. It's a little simpler, and `while True` makes me nervous.

exporter/api.py
```diff
@@ -55,19 +106,24 @@ def export_resources_to_tarball(learning_resources, username):
...
-                abs_path = os.path.join(tempdir, name)
-                if os.path.isfile(abs_path):
-                    archive.add(abs_path, arcname=name)
+            os.chdir(tempdir)
+            archive = tarfile.open(tempfilepath, mode='w:gz')
+
+            for root, _, files in os.walk("."):
+                for name in files:
+                    abs_path = os.path.join(root, name)
+                    abs_path = re.sub(r'^\./', '', abs_path)
```
ShawnMilo, line 116: Is there a difference between these two lines and `abs_path = os.path.join(root, name)[2:]`?

noisecapella, line 116: I feel a bit uneasy about string slicing without a check for what is being removed, even if in this situation it's guaranteed to exist.

ShawnMilo, line 116: ACK. I'm not one of those people who has a problem with a liberal use of regular expressions. I just thought it might be a tad clearer.

Maybe compile the regex at the top of the function with a comment that it strips off the leading `./`, then use the compiled regex in the loop.

exporter/api.py
```diff
@@ -22,28 +24,70 @@ def export_resources_to_directory(learning_resources):
...
+                filename = "{id}_{url_name}.xml".format(
+                    id=learning_resource.id,
+                    url_name=sanitize(learning_resource.url_name),
+                )
+            else:
+                filename = "{id}.xml".format(id=learning_resource.id)
+
+            type_name = learning_resource.learning_resource_type.name
+            try:
+                os.mkdir(os.path.join(tempdir, type_name))
```
ShawnMilo, line 53: This mkdir is still in a loop. I recommend keeping track with a set, like this (tested locally):

``` python
def mkdir(path):
    print "made directory {0}".format(path)

def save(filename):
    pass


files = (
    "a/one.png",
    "a/two.png",
    "a/three.png",
    "a/four.png",
    "a/five.png",
    "a/six.png",
    "a/seven.png",
    "b/one.png",
    "b/two.png",
    "b/three.png",
    "b/four.png",
    "b/five.png",
    "b/six.png",
    "b/seven.png",
)

seen = set()
for filename in files:
    path, name = filename.split("/")
    if path not in seen:
        mkdir(path)
        seen.add(path)
    save(file)
```

ShawnMilo:
It looks fine and works locally. I just added one more comment about a `mkdir` in a loop.

exporter/api.py
noisecapella, line 53: Will do

noisecapella:
I squashed the change in but I didn't change anything except the mkdir you mentioned

PR #433, opened 2015-07-29
Pinned all versions

carsongee:
Follow up fix to the xbundle requirement issue.  From now on we need to pin all requirements, even if we make up a version for the github ones to allow for pip to correctly upgrade them.


giocalitri:
I think this can be merged

PR #437, opened 2015-07-29
Created ICheckbox React component

### LINKED ISSUE #436: Create React component for iCheck checkboxes
This is necessary to match the UI mocks

noisecapella:
Closes #436


carsongee:
Can we swap this control into the Add vocabulary form to have a way to manually check that it is working?  That would help me verify it.  Also, this might be cool to turn into an NPM package if we have some time/ever need to use it in another project (obviously out of scope here).

noisecapella:
I'm open to an NPM package at some point in the future, not right now though. iCheck 2.0 is being worked on so whatever we release probably won't be useful for long.

Checkboxes are now in add vocabulary

Issue #429, opened 2015-07-28
the XML preview box for a LearningResource should be read only


giocalitri:
Currently it allows changes but they are not saved and this can be misleading even if the current implementation is consistent with the mockup.

PR #441, opened 2015-07-30
Made XML preview box for a LearningResource should be read only

amir-qayyum-khan:
Hi
In this PR I made XML preview box for a LearningResource should be read only

issue https://github.com/mitodl/lore/issues/429
@pdpinch @carsongee @giocalitri

screen shot 2015-07-30 at 3 03 09 pm

### DIFF
--- a/ui/static/ui/js/learning_resources.jsx
+++ b/ui/static/ui/js/learning_resources.jsx
@@ -70,6 +70,7 @@ define('learning_resources', [
         <form className="form-horizontal">
           <StatusBox message={this.state.message} />
           <textarea className="form-control textarea-xml"
+                    readOnly="true"
                     valueLink={this.linkState('contentXml')}
             />
 


ui/static/ui/js/learning_resources.jsx
```diff
@@ -70,6 +70,7 @@ define('learning_resources', [
         <form className="form-horizontal">
           <StatusBox message={this.state.message} />
           <textarea className="form-control textarea-xml"
+                    readOnly="true"
```
pdpinch, line 73: mixed case attributes look so strange to me, but I'll learn to live with it.

pdpinch:
:+1:

ui/static/ui/js/learning_resources.jsx
amir-qayyum-khan, line 73: yeah, but this is react syntax :smile:

noisecapella, line 73: That's correct, it's specifically mentioned in the style guide

Issue #372, opened 2015-07-16
User menu doesn't work on welcome page


noisecapella:
It works when viewing a repository but not on the welcome page. I inspected the element and it looks like the event is bound correctly for one and not the other, so I don't think it's simply a CSS issue

pdpinch:
Closed by #442

PR #442, opened 2015-07-30
Fixed user menu display on welcome page LORE

amir-qayyum-khan:
Hi

In this PR i fixed user menu display issue on welcome page

issue https://github.com/mitodl/lore/issues/372

@pdpinch @carsongee  @noisecapella

screen shot 2015-07-30 at 4 24 10 pm

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -115,6 +115,7 @@
         shim: SHIMS,
         paths: REQUIRE_PATHS
       });
+      require(["jquery", "bootstrap"]);
     </script>
     {% endcompress %}
     {% block postcontent %}


ui/templates/welcome.html
```diff
@@ -29,3 +30,14 @@
         </div>
     {% endif %}
 {% endblock %}
+{% block postcontent %}
+{{ block.super }}
+{% compress js %}
+  <script type="text/javascript"
```
noisecapella: Thanks for investigating this, I can confirm this causes the problem to go away. However you should replace these lines with:

```
<script type="text/javascript">
require(["jquery", "bootstrap"]);
</script>
```

On `repository.html` the problem doesn't appear because we require `listing.js` which in turn requires `jquery` and `bootstrap` among other things. It should be sufficient to require them on this page in the same way

noisecapella: Actually it may be better to put the `require(["jquery", "bootstrap"])` in `base.html` instead. Thoughts @carsongee ?

noisecapella: @amir-qayyum-khan: Can you remove the code from this page and just add `require(["jquery", "bootstrap"])` to `base.html` right after the `require.config({...})` statement?

carsongee: Agreed @noisecapella. We noticed that this also occurs on the create repository page, and since these two are required for any page we create, I think it makes sense to put them in base so this works on all pages without duplicating code everywhere.

amir-qayyum-khan:
Done @noisecapella

ui/templates/base.html
```diff
@@ -115,6 +115,7 @@
         shim: SHIMS,
         paths: REQUIRE_PATHS
       });
+      require(["jquery", "bootstrap"])
```
noisecapella: Please add a semicolon here

amir-qayyum-khan:
ah, sorry , fixed @noisecapella

noisecapella:
It looks like the test failed due to elasticsearch timing out. I've repushed to run the tests again.

PR #443, opened 2015-07-30
Added checkboxes to allow user to uncheck items in export panel

### LINKED ISSUE #425: Show export checkboxes on export panel and update API to input this list
This would let the user check and uncheck certain resources at the export panel and would also make it fit the UI mockups better.

noisecapella:
Fixes #425

### DIFF
--- a/exporter/api.py
+++ b/exporter/api.py
@@ -4,11 +4,11 @@ Functions for exporting learning resources
 
 from __future__ import unicode_literals
 
-from tempfile import mkdtemp, mkstemp
-import tarfile
 import os
-from shutil import rmtree
 import re
+import tarfile
+from shutil import rmtree
+from tempfile import mkdtemp, mkstemp
 
 from django.utils.text import slugify
 from django.core.files.storage import default_storage
--- a/rest/tests/base.py
+++ b/rest/tests/base.py
@@ -759,7 +759,7 @@ class RESTTestCase(LoreTestCase):
         if expected_status == HTTP_200_OK:
             return as_json(resp)
 
-    def create_learning_resource_export_task(self, repo_slug,
+    def create_learning_resource_export_task(self, repo_slug, input_dict,
                                              expected_status=HTTP_201_CREATED):
         """
         Helper method to create a task for the user to export a tarball of
@@ -769,7 +769,9 @@ class RESTTestCase(LoreTestCase):
             '/api/v1/repositories/{repo_slug}/'
             'learning_resource_export_tasks/'.format(
                 repo_slug=repo_slug,
-            )
+            ),
+            json.dumps(input_dict),
+            content_type='application/json',
         )
         self.assertEqual(expected_status, resp.status_code)
         if expected_status == HTTP_201_CREATED:
--- a/rest/tests/test_export.py
+++ b/rest/tests/test_export.py
@@ -6,7 +6,9 @@ from __future__ import unicode_literals
 from tempfile import mkdtemp
 from shutil import rmtree
 import tarfile
+import os
 
+from django.utils.text import slugify
 from django.test import override_settings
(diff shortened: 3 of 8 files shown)


noisecapella:
This includes chunks of the other PR, it may be easiest just to look at the commit diff rather than the PR diff until the other PR is merged into master

exporter/api.py
```diff
@@ -8,6 +8,7 @@
 import tarfile
 import os
 from shutil import rmtree
+import re
```
giocalitri, line 8: you should move this before the `'from` and sort the `import`s

exporter/api.py
```diff
@@ -37,8 +38,30 @@ def sanitize(title):
                 id=learning_resource.id,
                 title=sanitize(learning_resource.title),
             )
-            with open(os.path.join(tempdir, filename), 'w') as f:
+            type_name = learning_resource.learning_resource_type.name
+            try:
+                os.mkdir(os.path.join(tempdir, type_name))
+            except OSError:
+                pass
```
giocalitri: how can you be sure that the directory exists if you just pass here?
if you do not have rights to create the directory, it will still give you `OSError`:

```
In [1]: import os

In [2]: os.mkdir('/usr/bin/foogiovanni')
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-bdef8fc69048> in <module>()
----> 1 os.mkdir('/usr/bin/foogiovanni')

OSError: [Errno 13] Permission denied: '/usr/bin/foogiovanni'
```

noisecapella: That's true, I'll check for 'File exists' in the argument string

exporter/api.py
```diff
@@ -37,8 +38,30 @@ def sanitize(title):
...
+                static_path = re.sub(r"^assets/", "static/",
+                                     static_asset.asset.name)
+
+                relative_path, filename = os.path.split(static_path)
+                try:
+                    os.makedirs(os.path.join(tempdir, relative_path))
+                except OSError as ex:
+                    # Directory already exists
+                    if "File exists" not in ex.args:
+                        raise
```
giocalitri: Not sure I understood what are you are doing here

noisecapella: If the directory already exists, it's OK to pass, otherwise we should reraise the exception

exporter/api.py
```diff
@@ -37,8 +38,30 @@ def sanitize(title):
...
+                with open(os.path.join(
+                    tempdir, static_path
+                ), 'wb') as outfile:
```
giocalitri: nit: maybe this looks better as:

```
with open(
    os.path.join(tempdir, static_path),
    'wb'
) as outfile:
```

but feel free to ignore

exporter/api.py
giocalitri: makes sense

exporter/api.py
giocalitri: why don't you just check if the directory exists?

ShawnMilo: It's more Pythonic (and less verbose) to "ask for forgiveness" instead of "asking for permission."

noisecapella: It prevents against the race condition where you check if the directory exists, it does not at the time, then another process creates the directory and then you get an exception during `mkdir` anyway.

giocalitri: I was talking about in the except, not before

noisecapella: Both ways would work, I'm going to leave it the way it is since the change was made in the other PR (this PR includes bits of the other PR)

giocalitri: ack

pwilkins:
@noisecapella You mention the "other PR" in a early comment here.  Which PR is that?

noisecapella:
#427 which is merged now

noisecapella:
@pwilkins says he will do a quick run through of functionality and merge this. @giocalitri has already looked over the code

PR #445, opened 2015-07-30
Added test for StaticAsset.loader

noisecapella:
This kind of test is useful because it fails if we add arguments to the loader functions and forget to update its uses.


carsongee:
:+1:

Issue #346, opened 2015-07-13
Collapse vocabs/terms ui into a double drop-down


pdpinch:
This UI change is a solution to the problem that there may be many, many vocabularies in a repository. 

First column allows user to choose vocabulary, second column would be used to choose a term/terms.

noisecapella:
Currently when the Save button is clicked the UI will update each term to whatever is selected at the time. This UI change will limit the user to changing only one term selection at a time. That's not a big deal but it makes the UX clunkier IMHO

Ferdi:
@noisecapella are we talking about the same thing ?  Here is what I have in mind : 

![vocab_assign](https://cloud.githubusercontent.com/assets/125561/8702819/18ffbfce-2ae8-11e5-85be-d143244c3b3d.png)

noisecapella:
I think it would be a clunkier UX since you need to set two selection boxes instead of one. The second box would save its state when the value changes but the first box is just showing you which set of terms you're operating on. What does the '+' mean in your diagram?

Ferdi:
Let say you have 100 vocabularies and user wants to assign 2 terms to the given resource
vocab1: someterm
vocab99: someotherterm

Left diagram:
We list all vocabularies in the page when user actually wants to only use 2 of them. She has to scrooool down to find vocab99

Right diagram:
- User selects vocab1 : someterm  
- clicks +
- Selects vocab99: someotherterm

Make sense ?

noisecapella:
Oh I think I understand now, the user would have two lines for the two vocabularies they want to select, the rest would be left out

Ferdi:
Right.

pdpinch:
Also, we want to provide a tag-style ui for vocabs that support multiple values. Perhaps using https://select2.github.io/examples.html -- see "Tagging support"

giocalitri:
other useful link
https://github.com/JedWatson/react-select

noisecapella:
![18ffbfce-2ae8-11e5-85be-d143244c3b3d](https://cloud.githubusercontent.com/assets/863262/9087760/957a9272-3b5a-11e5-9986-c6915d62dd17.png)

After talking with Jamie about it yesterday I was wondering if this made more sense for a UI mock (excuse the badly thrown together image).

Ferdi:
I had in mind something like this

![vocab_assign](https://cloud.githubusercontent.com/assets/125561/9089133/b4e9fda8-3b61-11e5-818c-40b3d97a7077.png)

[...]

PR #447, opened 2015-07-30
Modified learningresource panel to include multi select fields

giocalitri:
Second part of story #346

### DIFF
--- a/ui/jstests/test-learning-resource.jsx
+++ b/ui/jstests/test-learning-resource.jsx
@@ -1,6 +1,6 @@
-define(['QUnit', 'jquery', 'learning_resources', 'reactaddons',
+define(['QUnit', 'jquery', 'reactaddons', 'lodash', 'learning_resources',
   'test_utils', 'jquery_mockjax'], function(
-  QUnit, $, LearningResources, React, TestUtils) {
+  QUnit, $, React, _, LearningResources, TestUtils) {
   'use strict';
 
   var VocabularyOption = LearningResources.VocabularyOption;
@@ -41,6 +41,12 @@ define(['QUnit', 'jquery', 'learning_resources', 'reactaddons',
     "label": "required",
     "weight": 1
   };
+  var prereqTermsResponseNotRequired = {
+    "id": 4,
+    "slug": "notrequired",
+    "label": "notrequired",
+    "weight": 1
+  };
   var vocabularyResponsePrereq = {
     "id": 1,
     "slug": "prerequisite",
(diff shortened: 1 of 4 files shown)


noisecapella:
I can confirm the functionality, looking over code now

ui/jstests/test-learning-resource.jsx
```diff
@@ -158,28 +164,46 @@ define(['QUnit', 'jquery', 'learning_resources', 'reactaddons',
...
+          // make sure second vocab (which has a default value) is set properly
+          var $terms2Select = $($vocabSelect[1]).find("option");
+          assert.equal($terms2Select.size(), 2);
+          assert.equal($terms2Select[0].selected, true);
+          assert.equal($terms2Select[1].selected, false);
+          //the second vocabulary can be a multi select
+          $($vocabSelect[1]).val(["notrequired", "required"]).trigger('change');
+          React.addons.TestUtils.Simulate.change($vocabSelect[1]);
+          component.forceUpdate(function() {
+            assert.equal($terms2Select[0].selected, true);
```
noisecapella, line 181: Can you move the rest of this test inside this callback to ensure that it all happens after the component was updated?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -1,34 +1,63 @@
...
+        multiple: isMultiTerms,
+        data: options,
+        placeholder: "Select a value for " + this.props.vocabulary.name,
+        allowClear: allowClear,
+        theme: "bootstrap",
+      })
+      .val(this.props.selectedTerms)
+      .trigger('change')
+      .on('change', function () {
+        var selectValues = [].concat($(this).val());
```
noisecapella, line 32: What is `[].concat` used for here? If you're just making a defensive copy, that already happens in the _.filter below

ui/jstests/test-learning-resource.jsx
giocalitri, line 181: sure

ui/static/ui/js/learning_resources.jsx
giocalitri, line 32: it's used because the `$(this).val()` on a select with single value does not return a list, so I want a generic way to transform the result to a list in both cases

noisecapella, line 32: Makes sense

noisecapella:
Just the thing about the test callback, everything else looks good

ui/jstests/test-learning-resource.jsx
noisecapella, line 181: `done` must be the very last thing called in the test. Otherwise the test completes without executing the callbacks that are queued. Can you move that within the callback too?

giocalitri:
I pushed the change, but the diff still shows the old code: not sure what's going on.
If you open the whole file you can see I modified the code.

giocalitri:
Done

giocalitri:
Tests passed: merging

Issue #348, opened 2015-07-13
Refresh search results after editing a learning resource


pdpinch:
So that facets and descriptions will be added to the search results.

pdpinch:
This was fixed incidentally when we started using react for the listing and the learning resource panel.

PR #453, opened 2015-07-31
Refreshed search results after editing a learning resource

amir-qayyum-khan:
Hi

In this PR i am refreshing repository/search results page after learning resource form save successfully.

Issue https://github.com/mitodl/lore/issues/348

@pdpinch @carsongee  @noisecapella

### DIFF
--- a/ui/static/ui/js/learning_resources.jsx
+++ b/ui/static/ui/js/learning_resources.jsx
@@ -156,6 +156,7 @@ define('learning_resources', [
         thiz.setState({
           message: "Form saved successfully!"
         });
+        location.reload();
       }).fail(function () {
         thiz.setState({
           message: {error: "Unable to save form"}


giocalitri:
The tests do not pass because the page is fully reloaded.

I guess we should implement this solution with another approach, like (hack solution until the listing is reimplemented in React):
- changing another variable in the state in React
- updating an hidden input in main page
- hook an event handlar to that input to reload the page in case that value changes

(or something like this)
@noisecapella ?

noisecapella:
Depending on how urgent #348 is we may want to just convert the listing page to React first (#455) 

My preferred solution is to have a prop called 'reload' take a callback which will execute. This would allow us to leave out the page refresh from the test but keep it in the listing page.

pdpinch:
Closing this in favor of the work started in #474

PR #460, opened 2015-08-03
Sped up Elasticsearch indexing.

ShawnMilo:
Cache course information so it doesn't do extra database
queries per LearningResource during import.

### DIFF
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -136,12 +136,25 @@ class LearningResource(BaseModel):
     xa_histogram_grade = models.FloatField(default=0)
     url_name = models.TextField(null=True)
 
-    def get_preview_url(self):
-        """Create a preview URL."""
+    def get_preview_url(self, org=None, course_number=None, run=None):
+        """
+        Create a preview URL. Accepts optional kwargs to prevent
+        database lookups, especially for during search engine indexing.
+        Args:
+            org (unicode): self.course.org
+            run (unicode): self.course.run
+            course_number (unicode): self.course.course_number
+        """
+        if org is None:
+            org = self.course.org
+        if course_number is None:
+            course_number = self.course.course_number
+        if run is None:
+            run = self.course.run
         key = "{org}/{course}/{run}".format(
-            org=self.course.org,
-            course=self.course.course_number,
-            run=self.course.run,
+            org=org,
+            course=course_number,
+            run=run,
         )
 
         if self.url_name is not None:
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -14,11 +14,82 @@ the index. They can not be mixed.
 
 from __future__ import unicode_literals
 
-from lxml import etree
-from haystack import indexes
+from collections import defaultdict
+from datetime import datetime, timedelta
+import logging
 
-from learningresources.models import LearningResource
-from taxonomy.models import Vocabulary
+from django.conf import settings
+from haystack import indexes
+from lxml import etree
+
+from learningresources.models import Course, LearningResource
+
+INDEX_CACHE = {
+    "course": {},
+    "term": {},
+    "born": datetime.now(),
+}
+MAX_INDEX_AGE = timedelta(minutes=1)
+
+log = logging.getLogger(__name__)
+
+
+def check_cache_age():
+    """
+    Keep the cache from getting stale.
+
+    Must not cache during tests or development, or it will
+    cause confusion. Must manually enable caching in production.
+    """
+    # pylint: disable=global-statement
+    global INDEX_CACHE
+    if (datetime.now() - INDEX_CACHE["born"] > MAX_INDEX_AGE) \
+            or settings.ALLOW_CACHING is False:
+        INDEX_CACHE = {
+            "course": {},
+            "term": {},
+            "born": datetime.now(),
+        }
+
+
+def get_course_metadata(course_id):
+    """
+    Caches and returns course metadata.
+    Args:
+        course_id (int): Primary key of learningresources.models.Course
+    Returns:
+        data (dict): Metadata about course.
+    """
+    check_cache_age()
+    data = INDEX_CACHE["course"].get(course_id, {})
+    if data == {}:
+        course = Course.objects.get(id=course_id)
+        data["run"] = course.run
+        data["course_number"] = course.course_number
+        data["org"] = course.org
+        data["repo_slug"] = course.repository.slug
+        INDEX_CACHE["course"][course_id] = data
+    return data
+
+
+def get_vocabs(course_id, lid):
+    """
+    Caches and returns taxonomy metadata for a course.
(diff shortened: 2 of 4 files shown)


learningresources/models.py
```diff
@@ -136,12 +136,25 @@ class LearningResource(BaseModel):
     xa_histogram_grade = models.FloatField(default=0)
     url_name = models.TextField(null=True)
 
-    def get_preview_url(self):
-        """Create a preview URL."""
+    def get_preview_url(self, org=None, course_number=None, run=None):
```
noisecapella, line 139: I wonder if this would be cleaner as a standalone function since we're now taking in all the arguments we need explicitly.

search/search_indexes.py
```diff
@@ -14,12 +14,75 @@
...
 
+INDEX_CACHE = {
+    "course": {},
+    "term": {},
+    "born": datetime.now(),
+}
+MAX_INDEX_AGE = timedelta(minutes=1)
+
+
+def reset_cache():
```
noisecapella: `reset_cache` makes it sound like the function always resets the cache, is it possible to change the name to be clearer about its purpose? If I saw a `Cache().reset()` in the wild I would expect it to always throw everything away

search/search_indexes.py
```diff
@@ -14,12 +14,75 @@
...
+        data["course_number"] = course.course_number
+        data["org"] = course.org
+        data["repo_slug"] = course.repository.slug
+        INDEX_CACHE["course"][course_id] = data
+    return data
+
+
+def get_vocabs(course_id, lid):
+    """
+    Caches and returns metadata for a course.
```
noisecapella: Can you update the description? It looks copied from above function

search/search_indexes.py
```diff
@@ -14,12 +14,75 @@
...
+    Args:
+        course_id (int): Primary key of learningresources.models.Course
+        lid (int): Primary key of learningresources.models.LearningResource
+    Returns:
+        data (dict): Vocab/term data for course.
+    """
+    reset_cache()
+    if INDEX_CACHE["term"].get(course_id) == None:
```
noisecapella: `is None` is cleaner here, no strong opinion however

learningresources/models.py
ShawnMilo, line 139: Right now all the kwargs are optional. This is because I didn't want to force any other code which calls this to change.

We could make it a separate function, but I think the bigger win would be to not have it implicitly do database lookups when it runs. I don't know if that's practical, though.

search/search_indexes.py
ShawnMilo: `is None` is more correct than `== None`. I'll change it.

noisecapella:
It would be nice if the cache were in a separate object so that it could be easily unit tested. I know there's functional testing done already of search indexing but unit tests highlight problems more specifically which allows for quicker diagnosis of problems.

noisecapella:
Looks good overall, some minor comments and tests are failing which you're probably already aware of.

ShawnMilo:
Yep, I'm working on fixing those now. I did it really quick one morning before stand-up, and didn't clean it up to push. This morning's stand-up prompted me to throw it up for people to look at.

ShawnMilo:
Note: When this passes code review, I still need to flatten it.

noisecapella:
Looks good except for the typo, feel free to merge

ShawnMilo:
Thanks, typo fixed and flattened. Will merge after tests pass.

@carsongee: This is disabled by default, so if we want to use it in Heroku, an environment variable has to be set. `ALLOW_CACHING`

search/search_indexes.py
```diff
@@ -14,11 +14,82 @@
...
+    if (datetime.now() - INDEX_CACHE["born"] > MAX_INDEX_AGE) \
+            or settings.ALLOW_CACHING is False:
+        INDEX_CACHE = {
+            "course": {},
+            "term": {},
+            "born": datetime.now(),
+        }
+
+
+def get_course_metadata(course_id):
```
carsongee, line 55: I think this would all be cleaner using a memoize decorator: https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize

carsongee, line 55: We could also use a django local mem cache too

carsongee:
It would be nice if there was a performance test we could add that would assert we wouldn't go slower, like assert number of DB queries for import of a course and make sure that doesn't go up with the flag enabled. Also, I don't know if we want the feature flag, if it makes things faster and doesn't break things, it should just always be on right? Or do it with a function parameter so the caller can determine if it is on or not, and not the sys admin?

search/search_indexes.py
ShawnMilo, line 55: The problem with memoize is that there's no concept of expiration. I do think Django's caching would be an upgrade, since they [added the ability](https://docs.djangoproject.com/en/1.8/topics/cache/#cache-arguments) to set an expiration timeout in Django 1.7. I'll open an issue for it.

ShawnMilo:
The main practical reason the flag is there is that it breaks all kinds of tests if it's on by default. Since it's called implicitly when various objects are saved, an optional kwarg won't help; it has to be something set in a larger scope.

The other reason it should be off by default is that otherwise it's going to cause a lot of confusion during development, as we save something and refresh and don't see the change -- unless we always keep caching in the front of our minds or remember to disable it locally. Since cache invalidation is one of the hard problems in programming, I think it's best to leave it off by default to prevent surprises. I also know this from bitter experience.

carsongee:
We shouldn't write it in such a way that it will break the tests if the feature is on, that would invalidate much of our testing (as it relates to verifying we are production ready), so part of adding this would be to make the code those tests use smarter (and probably some of the test smarter as well) to deal with this, otherwise we end with all those problems you have likely run into.
For this specifically, you can bypass that harder problem because cache invalidation is actually very easy to handle since we aren't using threads, we just need to turn cache on at start of import and turn it off afterwards, that way we are limiting the scope of the change and thus wouldn't need a flag to turn it on and off.  We can even limit the scope of the caching to inside the import pipeline

carsongee:
Have you looked at the speed up this provides for large courses? i.e. 8.01 or the MechCX course? i.e. from 5mins to 4mins or similar?

ShawnMilo:
I ran it against all the courses I had locally, and overall it cut the time in half. I didn't test any courses in isolation.

Your comments above (regarding cache invalidation) will take some more thinking to respond to, so a response will be later. Just so you didn't think I missed or ignored that part.

search/search_indexes.py
carsongee, line 55: It would be pretty trivial to add expiration to memoize, just make a parameter of the decorator the expiration time or add your `datetime.now() - INDEX_CACHE["born"] > MAX_INDEX_AGE)` inside the memoize wrapper if it is constant.

PR #466, opened 2015-08-04
Added loader and fixed scroll issue on save button press

### LINKED ISSUE #439: Successful LearningResource "save" may not be obvious to the user.
If the user is editing the description of a LearningResource and scroll down to the bottom to click the "Save" button, they may have no indication it worked and continue clicking. This is because the confirmation is at the top of the screen and may be invisible.

Possibilities:
- Move the confirmation to the bottom, where the user must be to click "Save."
- Scroll to the top upon click so the confirmation will be in view.

amir-qayyum-khan:
Hi

In this PR I have 
- Added a loader to both taxonomy panel and learning resource panel on save button click.
- After ajax call success/fail i scrolled up page to the **response message**. **Note** Only in case of success call in taxonomy  we switch tabs and scroll to new added vocabulary at bottom . 
- Fixed scroll issue on taxonomy and learning resource panels where previously we could scroll panels and background at same time (scroll overlapping issue)

fixes https://github.com/mitodl/lore/issues/439
@pdpinch @ShawnMilo @giocalitri 

screen shot 2015-08-04 at 4 58 30 pm

screen shot 2015-08-04 at 6 20 27 pm

screen shot 2015-08-04 at 6 19 59 pm

### DIFF
--- a/ui/jstests/test-learning-resource.jsx
+++ b/ui/jstests/test-learning-resource.jsx
@@ -176,17 +176,128 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     function (assert) {
       var done = assert.async();
 
+      var vocab = $.extend({}, selectedVocabulary);
+      var appendTermSelectedVocabulary = function(term) {
+        vocab.terms = vocab.terms.concat(term);
+      };
+
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
+
       var afterMount = function(component) {
         var $node = $(React.findDOMNode(component));
 
         var $termSelect = $node.find("select");
         assert.equal($termSelect.size(), 1);
 
-        done();
+        var $options = $termSelect.find("option");
+        var values = _.map($options, function(option) {
+          return $(option).val();
+        });
+        assert.deepEqual(values, ["easy", "hard"]);
+        // Add a term via free tagging
+
+        $termSelect
+          .append($('<option />', {value: 'medium'}))
+          .val(["hard", "easy", "medium"])
+          .trigger('change');
+
+        component.forceUpdate(function() {
+          assert.equal(loadedState, false);
+          waitForAjax(1, function() {
+            assert.equal(loadedState, true);
+            // term successfully added
+            assert.equal(vocab.terms.length, 3);
+            done();
+          });
+        });
       };
 
       React.addons.TestUtils.renderIntoDocument(
         <TermSelect
+          setValues={function() {}}
+          appendTermSelectedVocabulary={appendTermSelectedVocabulary}
+          repoSlug="repo"
+          setLoadedState={setLoadedState}
+          reportMessage={function() {}}
+          vocabs={vocabulariesAndTerms}
+          selectedVocabulary={selectedVocabulary}
+          ref={afterMount}
+        />
+      );
+    }
+  );
+
+  QUnit.test(
+    'Assert that TermSelect handles failure to add term gracefully',
+    function (assert) {
+      var done = assert.async();
+
+      TestUtils.replaceMockjax({
+        url: '/api/v1/repositories/repo/vocabularies/difficulty/terms/',
+        type: 'POST',
+        responseText: termResponseMedium,
+        status: 400
+      });
+
+      var vocab = $.extend({}, selectedVocabulary);
+      var appendTermSelectedVocabulary = function(term) {
+        vocab.terms = vocab.terms.concat(term);
+      };
+
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
+
+      var message;
+      var reportMessage = function(m) {
+        message = m;
+      };
+
+      var afterMount = function(component) {
+        var $node = $(React.findDOMNode(component));
+
+        var $termSelect = $node.find("select");
+        assert.equal($termSelect.size(), 1);
+
+        var $options = $termSelect.find("option");
+        var values = _.map($options, function(option) {
+          return $(option).val();
+        });
+        assert.deepEqual(values, ["easy", "hard"]);
+        // Add a term via free tagging
+
+        $termSelect
+          .append($('<option />', {value: 'medium'}))
+          .val(["hard", "easy", "medium"])
+          .trigger('change');
+
+        component.forceUpdate(function() {
+          assert.equal(loadedState, false);
+          waitForAjax(1, function() {
+            assert.equal(loadedState, true);
+
+            // Didn't change
+            assert.equal(vocab.terms.length, 2);
+            assert.deepEqual(
+              message,
+              {error: 'Error occurred while adding new term "medium".'}
+            );
+            done();
+          });
+        });
+      };
+
+      React.addons.TestUtils.renderIntoDocument(
+        <TermSelect
+          setValues={function() {}}
+          appendTermSelectedVocabulary={appendTermSelectedVocabulary}
+          repoSlug="repo"
+          setLoadedState={setLoadedState}
+          reportMessage={reportMessage}
           vocabs={vocabulariesAndTerms}
           selectedVocabulary={selectedVocabulary}
           ref={afterMount}
@@ -364,7 +475,9 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
             assert.equal(closeLearningResourcePanelCount, 0);
             React.addons.TestUtils.Simulate.click(saveAndCloseButton);
             component.forceUpdate(function() {
+              assert.equal(component.state.loaded, false);
               waitForAjax(1, function() {
+                assert.equal(component.state.loaded, true);
                 assert.equal(component.state.message,
                   "Form saved successfully!");
                 assert.equal(refreshCount, 2);
@@ -413,16 +526,20 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
           });
           var saveButton = $node.find("button")[0];
           React.addons.TestUtils.Simulate.click(saveButton);
-          waitForAjax(1, function() {
-            assert.deepEqual(
-              component.state.message,
-              {error: "Unable to save form"}
-            );
+          component.forceUpdate(function() {
+            assert.equal(component.state.loaded, false);
+            waitForAjax(1, function () {
+              assert.equal(component.state.loaded, true);
+              assert.deepEqual(
+                component.state.message,
+                {error: "Unable to save form"}
+              );
 
-            assert.equal(0, refreshCount);
-            //assert that panel does not close on ajax fail
-            assert.equal(closeLearningResourcePanelCount, 0);
-            done();
+              assert.equal(0, refreshCount);
+              //assert that panel does not close on ajax fail
+              assert.equal(closeLearningResourcePanelCount, 0);
+              done();
+            });
           });
         });
       };
--- a/ui/jstests/test_manage_taxonomies.jsx
+++ b/ui/jstests/test_manage_taxonomies.jsx
@@ -211,6 +211,11 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
         errorMessage = message;
       };
 
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
+
       var afterMount = function(component) {
         var labels = React.addons.TestUtils.
         scryRenderedDOMComponentsWithTag(
@@ -260,8 +265,10 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             assert.equal(refreshCount, 0);
             React.addons.TestUtils.Simulate.click(saveButton);
             component.forceUpdate(function() {
+              assert.equal(loadedState, false);
               //after saved term using api
               waitForAjax(1, function() {
+                assert.equal(loadedState, true);
                 //term state reset
                 assert.equal(component.state.formatActionState, 'show');
                 // term is update in parent
@@ -392,6 +399,7 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             vocabulary={vocabulary}
             refreshFromAPI={refreshFromAPI}
             reportMessage={reportMessage}
+            setLoadedState={setLoadedState}
             ref={afterMount}
           />
         );
@@ -422,6 +430,11 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
       var refreshCount = 0;
       var refreshFromAPI = function() {
         refreshCount++;
+      };
+
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
       };
 
       var afterMount = function(component) {
@@ -481,7 +494,10 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
                 assert.equal(refreshCount, 0);
                 React.addons.TestUtils.Simulate.click(saveButton);
                 component.forceUpdate(function() {
+                  // Loader is spinning...
+                  assert.equal(loadedState, false);
                   waitForAjax(1, function() {
+                    assert.equal(loadedState, true);
                     assert.equal(component.state.label, "TestB");
                     assert.equal(
                       component.state.errorMessage, 'Unable to update term'
@@ -530,6 +546,7 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             updateTerm={updateTerm}
             vocabulary={vocabulary}
             refreshFromAPI={refreshFromAPI}
+            setLoadedState={setLoadedState}
             ref={afterMount}
           />
         );
@@ -564,6 +581,11 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
         refreshCount++;
       };
 
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
+
       var afterMount = function(component) {
         var deleteButton = React.addons.TestUtils.
           findRenderedDOMComponentWithClass(
@@ -573,8 +595,10 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
         //select delete
         React.addons.TestUtils.Simulate.click(deleteButton);
         component.forceUpdate(function() {
+          assert.equal(loadedState, false);
           assert.equal(component.state.formatActionState, 'show');
           waitForAjax(1, function() {
+            assert.equal(loadedState, true);
             // term is delete in parent
             assert.equal(parentUpdateCount, 1);
             // listing was asked to refresh
@@ -593,6 +617,7 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             deleteTerm={deleteTerm}
             vocabulary={vocabulary}
             refreshFromAPI={refreshFromAPI}
+            setLoadedState={setLoadedState}
             ref={afterMount}
           />
         );
@@ -629,6 +654,11 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
         refreshCount++;
       };
 
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
+
       var afterMount = function(component) {
         var deleteButton = React.addons.TestUtils.
           findRenderedDOMComponentWithClass(
@@ -642,7 +672,9 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             component.state.errorMessage, ''
           );
           assert.equal(component.state.formatActionState, 'show');
+          assert.equal(loadedState, false);
           waitForAjax(1, function() {
+            assert.equal(loadedState, true);
             // check state of error message
             assert.equal(
               component.state.errorMessage, 'Unable to delete term.'
@@ -661,6 +693,7 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
             deleteTerm={deleteTerm}
             vocabulary={vocabulary}
             refreshFromAPI={refreshFromAPI}
+            setLoadedState={setLoadedState}
             ref={afterMount}
           />
         );
@@ -693,6 +726,11 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
       };
 
       var reportMessage = function() {};
+
+      var loadedState;
+      var setLoadedState = function(loaded) {
+        loadedState = loaded;
+      };
 
       var refreshCount = 0;
       var refreshFromAPI = function() {
@@ -783,7 +821,9 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
                   React.addons.TestUtils.Simulate.keyUp(
                     textbox, {key: "Enter"}
                   );
+                  assert.equal(loadedState, false);
                   waitForAjax(1, function () {
+                    assert.equal(loadedState, true);
                     assert.equal(addTermCalled, 1);
                     done();
                   });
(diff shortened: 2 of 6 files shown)


amir-qayyum-khan:
FYI it need some tests

pdpinch:
Just to confirm: does it have the needed tests now? (coverage seems good)

pdpinch:
@ferdi, is this what you had in mind for #439 ?

amir-qayyum-khan:
@pdpinch done with writing tests.

amir-qayyum-khan:
@pdpinch fixed coverage of https://coveralls.io/builds/3236569/source?filename=importer%2Fapi%2F__init__.py#L58

Thought it was out of scope of this PR. but I fixed it anyway, to fix CI

noisecapella:
The importer coverage change is out of the scope of this PR, so it should go in another PR. We will get fluctuations in code coverage from coveralls all the time which will display the X, we should only make that a blocker if some functionality touched by the PR (or removed by it) is left uncovered.

importer/api/__init__.py
```diff
@@ -84,6 +84,9 @@ def seekable():
         default_storage.delete(filename)
         rmtree(tempdir)
 
+    import_course_from_file.seekable = seekable
```
noisecapella: We should not be adding properties to methods for the sake of code coverage

importer/api/__init__.py
```diff
@@ -84,6 +84,9 @@ def seekable():
...
+    return import_course_from_file
```
noisecapella: Now the function is attaching a method to itself and returning itself. This is fraught with potential problems. It changes the function on import vs run. You're also returning something completely unrelated to the purpose of the function

ui/jstests/test-learning-resource.jsx
```diff
@@ -228,7 +228,15 @@ define(['QUnit', 'jquery', 'reactaddons', 'lodash', 'learning_resources',
 
           var saveButton = $node.find("button")[0];
           React.addons.TestUtils.Simulate.click(saveButton);
+          assert.equal(
```
noisecapella: Please only assert a change in state here. We can add a separate test to assert that the state change causes a difference in the rendered output, and so we have two smaller tests rather than one big test.

ui/jstests/test-learning-resource.jsx
```diff
@@ -261,7 +269,15 @@ define(['QUnit', 'jquery', 'reactaddons', 'lodash', 'learning_resources',
           });
           var saveButton = $node.find("button")[0];
           React.addons.TestUtils.Simulate.click(saveButton);
+          assert.equal(
```
noisecapella: Same comment about asserting state change here.

ui/jstests/test_lr_exports.jsx
```diff
@@ -824,7 +824,7 @@ define(['QUnit', 'jquery', 'lr_exports', 'reactaddons',
 
     assert.equal(0, $(container).find("div").size());
     Exports.loader("repo", "user", function() {}, container);
-    assert.equal(1, $(container).find("div").size());
+    assert.equal(2, $(container).find("div").size());
```
noisecapella: Where did this change come from? It doesn't look like we are editing `lr_exports.jsx` here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -624,7 +625,15 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'reactaddons',
                     //and the multi terms state is consistent
                     assert.equal(component.state.multiTerms, false);
                     React.addons.TestUtils.Simulate.submit(formNode);
+                    assert.equal(
```
noisecapella: Please only assert state changes here as commented earlier

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -709,7 +719,15 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'reactaddons',
           {target: {value: 'course', checked: true}}
         );
         React.addons.TestUtils.Simulate.submit(formNode);
+        assert.equal(
```
noisecapella: Same comment about asserting state changes

ui/static/ui/css/slide-drawer.css
```diff
@@ -81,6 +81,7 @@ html, body {
   -webkit-transition: visibility 0s 0s;
   -moz-transition: visibility 0s 0s;
   transition: visibility 0s 0s;
+  overflow: scroll;
```
noisecapella: This leaves a horizontal scrollbar at the bottom of the page after the panel closes. What is this for?

ui/static/ui/css/mit-lore.css
```diff
@@ -494,3 +494,13 @@ ul.repos {
     padding: 0px;
     margin: 0px;
 }
+
+i.add-vocabulary-spinner {
```
noisecapella: Please reindent to match rest of CSS (this was merged into master just recently so you may need to rebase first)

ui/static/ui/js/learning_resources.jsx
```diff
@@ -123,6 +124,7 @@ define('learning_resources', [
         terms: terms,
         description: this.state.description
       };
+      $(this.getDOMNode()).find('.lr-spinner').css('display', 'inline');
```
noisecapella: Use React.findDOMNode(this)

noisecapella: You must not mess with the DOM in React. The only exceptions occur in certain lifecycle functions and this doesn't apply. You have to alter the state here and have React rerender with the spinner enabled or disabled.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -135,12 +137,27 @@ define('learning_resources', [
...
+        thiz.hideLoaderAndScrollUp();
       }).fail(function () {
         thiz.setState({
           message: {error: "Unable to save form"}
         });
+        thiz.hideLoaderAndScrollUp();
       });
     },
+    hideLoaderAndScrollUp: function() {
+      var scrollableDiv = $(this.getDOMNode()).closest(
```
noisecapella: React components should not touch anything outside of the component hierarchy. You should not read from the DOM in a React component and you certainly shouldn't touch any of a DOM element's ancestors.

If you need a React component to make a modification outside the component you should pass it a callback in the props and execute that. That callback should not access React's DOM either, but here you shouldn't need to.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -135,12 +137,27 @@ define('learning_resources', [
...
+        ".cd-panel .cd-panel-container .cd-panel-content"
+      );
+      var messageBox = $(this.getDOMNode()).find('.status-container');
+      $(this.getDOMNode()).find('.lr-spinner').css('display', 'none');
+      if (typeof messageBox !== 'undefined' && $(messageBox).text()) {
```
noisecapella: Why not just `messageBox !== undefined`?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -135,12 +137,27 @@ define('learning_resources', [
...
+        scrollableDiv.animate(
+          {scrollTop: messageBox.offset().top},
```
noisecapella: You shouldn't need to scroll to the messageBox exactly, why not just scroll all the way to the top?

ui/static/ui/js/listing.js
```diff
@@ -88,6 +88,7 @@ define('listing',
         LearningResources.loader(repoSlug, learningResourceId, $("#tab-1")[0]);
         $('.cd-panel').addClass('is-visible');
         StaticAssets.loader(repoSlug, learningResourceId, $("#tab-3")[0]);
+        $('body').css('overflow', 'hidden');
```
noisecapella: What is the purpose of setting `overflow` on `body in all of these places?

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -177,6 +177,9 @@ define('manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
         learning_resource_types: this.state.learningResourceTypes,
         multi_terms: this.state.multiTerms
       };
+      $(this.getDOMNode()).find(
```
noisecapella: We should not touch the DOM directly

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -206,20 +209,38 @@ define('manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
...
       }).done(function(data) {
         // Reset state (and eventually update the vocab tab
         thiz.props.updateParent(data);
         thiz.replaceState(thiz.getInitialState());
         // Switch to taxonomy panel and scroll to new vocab (bottom)
+        thiz.hideLoaderAndScrollUp(true);
+      });
+    },
+    hideLoaderAndScrollUp: function(switchTab) {
+      switchTab = switchTab || false;
```
noisecapella: Please just pass in `false` above and remove this line

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -206,20 +209,38 @@ define('manage_taxonomies', ['reactaddons', 'lodash', 'jquery', 'utils'],
...
-          ".cd-panel-2 .cd-panel-container .cd-panel-content"
-        );
         scrollableDiv.animate(
           {scrollTop: scrollableDiv.prop('scrollHeight')},
           500
         );
-      });
+      } else {
+        var messageBox = $(this.getDOMNode()).find('.alert-dismissible');
+        if (typeof messageBox !== 'undefined' && $(messageBox).text()) {
```
noisecapella: Same comments as above

ui/static/ui/js/utils.jsx
```diff
@@ -32,23 +32,25 @@ define("utils", ["jquery", "lodash", "reactaddons", "select2"],
...
-      if (!this.props || !this.props.message) {
+      if (!this.props) {
         return null;
       }
 
-      if (this.props.message.error !== undefined) {
-        return <div className="alert alert-danger alert-dismissible">
+      if (this.props.message && this.props.message.error !== undefined) {
+        return <div
+          className="alert alert-danger alert-dismissible status-container">
```
noisecapella: Since the only reason you're defining `status-container` is to get its position, and you can just scroll the panel to the top it seems like you don't need these changes anymore

importer/api/__init__.py
amir-qayyum-khan: Revert code to previous state (removed this code)

importer/api/__init__.py
amir-qayyum-khan: Revert code to previous state (removed this code)

ui/jstests/test-learning-resource.jsx
amir-qayyum-khan: Done, made a new test to check that loader show and hides properly.

ui/jstests/test-learning-resource.jsx
amir-qayyum-khan: It is ok to test loader in save LR panel only, because the show/hide mechanism is same in both cases success and failure.

ui/jstests/test_lr_exports.jsx
amir-qayyum-khan: Because now there will be always a message `<div>` in panel with no text, That , made div size 2.

see react class StatusBox

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Created a new test to monitor loader

ui/jstests/test_manage_taxonomies.jsx
amir-qayyum-khan: Created a new test to monitor loader

ui/static/ui/css/slide-drawer.css
amir-qayyum-khan: This is for handling scroll overlap between panels and background. Before this fix you can vertically scroll panel and background same time, which case issue on animate scroll up in case of error message or success.

Now at the time one scroll will be enable i.e if you open a panel i will disable background page scroll

ui/static/ui/js/listing.js
amir-qayyum-khan: To avoid the scroll overlapping between panel and background. This overlapping case failure in scroll up animation. if you scroll down and scroll background as well then press save it will not scroll up with animation.

ui/jstests/test-learning-resource.jsx
```diff
@@ -228,6 +229,7 @@ define(['QUnit', 'jquery', 'reactaddons', 'lodash', 'learning_resources',
 
           var saveButton = $node.find("button")[0];
           React.addons.TestUtils.Simulate.click(saveButton);
+          assert.equal(component.state.message, undefined);
```
noisecapella: Whenever `setState` is called, even if triggered by an event, you must always assert the state change within the `forceUpdate` callback to ensure the state change was made.

ui/jstests/test-learning-resource.jsx
```diff
@@ -237,7 +239,43 @@ define(['QUnit', 'jquery', 'reactaddons', 'lodash', 'learning_resources',
...
+        repoSlug="repo"
+        learningResourceId="1"
+        ref={afterMount} />);
+    }
+  );
+  QUnit.test(
+    'Assert that loader on LearningResourcePanel show and hides properly',
+    function(assert) {
+      var done = assert.async();
+      var scrollUpToStatusBox = function() {};
```
noisecapella: Can you keep track of when this function is called and add an assertion that it was called?

jamiefolsom:
@amir-qayyum-khan It would appear as though there are now merge conflicts that need to be resolved. There are a few items like the background color when the panel is open that have been merged, for example. Might be a good time to rebase against master?

jamiefolsom:
The scroll to top functionality works, and I do see the loading spinner next to the save button briefly on save. The loading spinner appears to change position while it's visible, however.

jamiefolsom:
All tests are passing; coverage is good, and there are no js errors in the console.

amir-qayyum-khan:
@jamiefolsom done with rebase

amir-qayyum-khan:
Done with rebase @pdpinch

jamiefolsom:
OK, post-rebase (which is again 4 commits behind master, unfortunately):
- Functionality works (there was some debate at the planning meeting today about UX, but that's outside the scope of this review)
- Tests pass
- No js errors in the console.

I will look at code next.

@pdpinch @noisecapella @amir-qayyum-khan

jamiefolsom:
There's a slight jag when the spinner shows/hides. Hard coding it to show in the Chrome dev tools panel, it's possible to see that it's because in addition to "spinning", the whole gif is moving in a circular pattern.

spinner

ui/static/ui/css/mit-lore.css
```diff
@@ -554,3 +554,28 @@ input.repo-page-status {
...
+    margin-left: 10px;
+}
+
+i.lr-spinner-hide {
+    display: none;
+    margin-left: 10px;
+}
+
+i.lr-spinner-show {
+    display: inline;
```
jamiefolsom, line 604: Make this `display: inline-block;` instead.

jamiefolsom:
Appears to be due `display: inline;`, so instead add `display:inline-block;`:

spinner-inline-block

And just for fun, have a look at what happens when you apply `display:block;`:

spinner-block

See: https://github.com/mitodl/lore/pull/466/files#diff-83189b8b88a4be317bb11393848b74eeR578

jamiefolsom:
@amir-qayyum-khan, apart from the "spinning spinner" issue,  this PR looks good to me :+1:  (although it does need to be rebased again). @noisecapella do you want to have a look?

amir-qayyum-khan:
Fixed and rebased.
Thanks @jamiefolsom

jamiefolsom:
Thanks, @amir-qayyum-khan. Github is reporting conflicts, though?

noisecapella:
I think it's important that @ferdi or @pdpinch run this locally and weigh in on whether this functionality works as desired.

pdpinch:
Amir, we'd like to change the messaging behavior for this, to avoid the scrolling altogether. See https://github.com/mitodl/lore/issues/439#issuecomment-138642892

amir-qayyum-khan:
sure, let me do it @pdpinch

amir-qayyum-khan:
Done @pdpinch

ui/jstests/test-learning-resource.jsx
```diff
@@ -209,6 +209,7 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     'Assert that LearningResourcePanel changes state properly',
     function(assert) {
       var done = assert.async();
+      var scrollUpToStatusBox = function() {};
```
noisecapella: Can you remove this since we don't do scrolling here anymore?

pdpinch:
After some further discussion, instead of placing the confirmation banner next to the save button, we'd like to have it floated at the top of the viewport (still within the panel). This will make it appear in a consistent, easy-to-see location. 

The banner also needs a close/dismiss button.

amir-qayyum-khan:
Done @pdpinch

noisecapella:
The status message covers up the tabs in the panel
screenshot from 2015-09-16 16 41 17

amir-qayyum-khan:
@noisecapella any suggestion on where should i put this status box? 
Because top most header of panel has css property `position: fixed`  thats why i am not able to put status-box  on the top of panel :(

noisecapella:
That's up to @pdpinch, I would be happy with it near the top like it was before

pdpinch:
Can the top most header of the panel be floated, so it moves down when the the status box is visible?

pdpinch:
Amir and I spoke and he's going to push another version of this

amir-qayyum-khan:
Done with updating status message.
@pdpinch

pdpinch:
I'm having some trouble with this when I dismiss the message box. It seems like whenever I update the state of the component, the message reappears -- even if the error has been removed. 

Let me know if you need more detail.

amir-qayyum-khan:
Done with rebasing and bug fixed.
Please have a look now @pdpinch.

jamiefolsom:
Yes, I think you're right -- this is the same issue. I will close the other one as a dupe. Thanks!

ui/static/ui/css/mit-lore.css
```diff
@@ -622,3 +622,35 @@ p.confirmation-description {
     font-style: normal;
     font-weight: normal;
 }
+
+i.add-vocabulary-spinner-hide {
```
noisecapella: We don't need to use CSS here to hide and show the spinner. Please change the classes to `add-vocabulary-spinner` and `lr-spinner` and remove the `display` style attributes.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -267,7 +267,10 @@ define('learning_resources',
       var vocabSelector = "There are no terms for this resource";
       var termSelector = "";
       var termList = "";
-
+      var spinnerClassName = "fa fa-spinner fa-pulse fa-3x lr-spinner-hide";
+      if (this.state.showSpinner) {
```
noisecapella: Instead of setting a different CSS class please just set the spinner to a variable here where it is null if `showSpinner` is false and set to `<i className=... />` if it's true.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -370,15 +375,20 @@ define('learning_resources',
...
         thiz.props.refreshFromAPI();
+        thiz.hideLoaderAndScrollUp();
       }).fail(function () {
         thiz.setState({
           message: {error: "Unable to save form"}
         });
+        thiz.hideLoaderAndScrollUp();
       });
     },
+    hideLoaderAndScrollUp: function() {
```
noisecapella: Can you rename this to `hideLoader`?

noisecapella:
I'll make a PR to this with my suggested changes

amir-qayyum-khan:
ok @noisecapella

pdpinch:
@noisecapella @amir-qayyum-khan what's the status of this PR?

pdpinch:
@amir-qayyum-khan it looks like this needs a rebase.

When it's ready it fixes #439

amir-qayyum-khan:
Hi @pdpinch,
@noisecapella is working on this PR. I am not up to date on this.

noisecapella:
Sorry for the long delay in replying. I'm a bit uneasy with the UX of the message box here. Regarding the loader we should use the new loader added recently in #750. I'll rewrite the PR to limit it to the loader change and we can deal with the message box in another PR

jamiefolsom:
@noisecapella I see two loaders when I add a term to a vocabulary. It appears that both the listing loader and the panel loader are shown/hidden on term save. See below.

image

noisecapella:
I think that's appropriate behavior since first it shows the loader while the term is saved, then it shows the other loader while the page is refreshed, although the listing spinner should be hidden under the panel. Do you think it would look better to have it happen within the single loader?

jamiefolsom:
Does adding a term trigger a whole page refresh, or just a panel rerender?

noisecapella:
It does the panel rerender then the whole page refresh so it can update the values in the facets. Although now that I think of it adding a term shouldn't ever change the facet count since a new term will not be associated with any resources.

jamiefolsom:
Right, not until the LR panel is opened, and a term applied to the LR.

noisecapella:
Ok, the spinner should not show up above the panel now. Do you think it's a clean UX to have two loaders for different parts of the page? I can go either way

jamiefolsom:
@noisecapella looks much better with the secondary loader in the background.

jamiefolsom:
- [x] Code
- [x] Functionality
- [x] Tests
- [x] Questions

ui/jstests/test-learning-resource.jsx
```diff
@@ -176,17 +176,128 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     function (assert) {
       var done = assert.async();
 
+      var vocab = $.extend({}, selectedVocabulary);
+      var appendTermSelectedVocabulary = function(term) {
+        vocab.terms = vocab.terms.concat(term);
+      };
+
+      var loadedState;
+      var setLoadedState = function(loaded) {
```
jamiefolsom, line 185: I assume this function expects a boolean? Out of curiosity, do we have any conventions about parameter naming to help with clarity? Maybe just a comment would be helpful.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -379,12 +386,17 @@ define('learning_resources',
         thiz.setState({
           message: {error: "Unable to save form"}
         });
+      }).always(function() {
+        thiz.setState({loaded: true});
```
jamiefolsom, line 390: Curious about the semantics of this. If I understand correctly, setting loaded state to true just removes the spinner, correct? What if we want to tie other events to success in the future? Wouldn't having loaded state set to true (removing the spinner), even when the promise results in a rejection, in turn result in our triggering those hypothethical future success events when we shouldn't? @noisecapella

noisecapella, line 390: This makes the spinner go away even on failure but I don't think that would trigger future success events. Can you think of a scenario where it would?

ui/jstests/test-learning-resource.jsx
noisecapella, line 185: It should be a boolean, I don't think we have any conventions on parameters. There is JSDoc but I'm not sure it's necessary for functions with a limited purpose like this

ui/static/ui/js/learning_resources.jsx
jamiefolsom, line 390: No agreed, it wouldn't trigger anything in and of itself. What I meant is, the "loaded" state being set to true implies that the tab content was loaded successfully. If we ever relied on that state for anything else, like allowing XML content to be edited for example, or letting applied terms be modified, we might assume the content was successfully loaded, when it's not necessarily; that state is set to true even on failure, to remove the spinner. Really just a semantic thing, I think. Does that make sense?

ui/jstests/test-learning-resource.jsx
jamiefolsom, line 185: OK, just a nit. Carry on.

jamiefolsom:
OK, apart from those nits, this one looks good to me. :+1:

PR #472, opened 2015-08-05
Added management command for xanalytics updates.

### LINKED ISSUE #470: create a management command for populating xanalytics
This management command will accept a course ID and populate xanalytics data. This will be used to manually refresh the data, either because the course was imported before the xanalytics API was plugged in, or because the xanalytics data has been updated and should be reflected in LORE.

In its first iteration, it will generate random values for demo purposes. In the next, it will use the real xanalytics API.

Proposed syntax for populating xanalytics data for a course with primary key `23`.

``` sh
./manage.py populate_xanalytics 23
```

The work will be based on #166, and will be dependent on its PR (#403) being merged first.

ShawnMilo:
Closes #470

Fills the xanalytics fields with random data until API is available.

### DIFF
--- a/xanalytics/management/commands/populate_xanalytics.py
+++ b/xanalytics/management/commands/populate_xanalytics.py
@@ -1,0 +1,42 @@
+"""Management command to populate xanalytics data by course ID."""
+
+import logging
+from random import randint
+
+from django.core.management.base import BaseCommand, CommandError
+
+from learningresources.models import Course
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Populate xanalytics data for a course."""
+
+    help = "Populate xanalytics data for a course by primary key."
+
+    def add_arguments(self, parser):
+        """Add argparse arguments."""
+        parser.add_argument('course_id', nargs=1, type=int)
+
+    def handle(self, *args, **options):
+        """Run the command."""
+        try:
+            # Args are passed in a list; there's only one, so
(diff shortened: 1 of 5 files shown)


xanalytics/management/commands/populate_xanalytics.py
```diff
@@ -0,0 +1,42 @@
+"""Management command to populate xanalytics data by course ID."""
+
+import logging
+from random import randint
```
noisecapella, line 4: Would it be possible to seed random first, for example  `Random(0).randint(2,3)`? That way we have reproduceable data

ShawnMilo, line 4: It's possible. However, since courses can't be imported twice it wouldn't make a difference in the CI environment, and in tests all that matters is that they now have non-zero values. In short, I don't see any benefit, and it's slightly more code.

noisecapella:
I'm fine with whatever you choose regarding `Random`. Tests are failing but since you're just merging into another feature branch I guess that doesn't matter

ShawnMilo:
I just pushed an update that fixes the tests. I'll ping you once they pass.

ShawnMilo:
@noisecapella: Tests have passed.

ShawnMilo:
That must be a record.

pdpinch:
Uh, wait. This was merged into a feature branch, not master? Don't we need to merge it into master?

PR #475, opened 2015-08-06
Added xanalytics management command.

ShawnMilo:
Populate  xanalytics data from the command line with
random data.

This is as stand-alone extraction of the management command merged into #166 as #472.

### DIFF
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -118,7 +118,7 @@ class LearningResource(BaseModel):
     The units that compose an edX course:
     chapter, sequential, vertical, problem, video, html, etc.
     """
-    course = models.ForeignKey(Course)
+    course = models.ForeignKey(Course, related_name="resources")
     learning_resource_type = models.ForeignKey('LearningResourceType')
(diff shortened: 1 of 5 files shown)


carsongee:
I'm not seeing the management command here

ShawnMilo:
@carsongee: FYI. I tested the app locally; if the full suite passes on Travis this should be "mergeable" into master.

ShawnMilo:
It's here:

`xanalytics/management/commands/populate_xanalytics.py`

This is how the [current docs](https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/) describe it.

carsongee:
Cool, looks good, assuming tests pass, maybe we can get this post code freeze so I don't have to do it on the side. What do you think @pwilkins?

carsongee:
When I had looked the diff only had 2 files in it for whatever reason, and did not have the management command.

ShawnMilo:
Yeah, I failed to `git add` some files at first after applying a patch locally.

PR #476, opened 2015-08-06
Added xanalytics icons to listing page

noisecapella:
Added icons to xanalytics page. Some notes:
- this information is pulled from the database. We will probably want another PR to pull from elasticsearch instead
- All information is shown for each learning resource type. If this isn't appropriate please let me know which learning resource types will show what kinds of data
- No histogram

### DIFF
--- a/ui/templates/includes/learning_resource.html
+++ b/ui/templates/includes/learning_resource.html
@@ -48,7 +48,12 @@
         <div class="tile-meta">
           <span class="meta-item">{{ result.course }}</span>
           <span class="meta-item">{{ result.run }}</span>
-          <span class="meta-item"><a href="{{ result.preview_url }}" target="_blank">Preview</a></span>
+          <span class="sub-meta">
+            <span class="meta-item mi-col-1"><i class="fa fa-eye"></i> {{ result.nr_views }}</span>
+            <span class="meta-item mi-col-2"><i class="fa fa-edit"></i> {{ result.nr_attempts }}</span>
+            <span class="meta-item mi-col-3"><i class="fa fa-graduation-cap"></i> {{ result.avg_grade }}</span>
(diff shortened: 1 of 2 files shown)


ui/templates/includes/learning_resource.html
```diff
@@ -48,7 +48,12 @@
         <div class="tile-meta">
           <span class="meta-item">{{ result.course }}</span>
           <span class="meta-item">{{ result.run }}</span>
-          <span class="meta-item"><a href="{{ result.preview_url }}" target="_blank">Preview</a></span>
+          <span class="sub-meta">
+            <span class="meta-item mi-col-1"><i class="fa fa-eye"></i> {{ result.object.xa_nr_views }}</span>
```
ShawnMilo: These all use `result.object`, which bypasses the work done in #417. 

However, those fields are actually in `result`, so if you replace, for example, `result.object.xa_nr_views` with `result.xa_nr_views`, that should make this better.

For reference, see `search/search_indexes.py`. Any fields in `class LearningResourceIndex` should be available to you on `result`. Any calls to `result.object` will cause a database query.

noisecapella: I was aware of the performance hit but I had assumed this wasn't in the search index, but since it is I'll use that. By the way it's `result.nr_views`, the `xa_` is removed from the fields on that class.

ShawnMilo:
I'll test it locally. The code looks fine.

Yes, the fields are inconsistently named in the search index. I should fix that. It'll require a re-indexing everywhere once it deploys, but that'll be for another day.

ShawnMilo:
xa_screenshot

Looks good to me. If tests pass, please flatten.

ShawnMilo:
Also, sorting works.

noisecapella:
I love that coveralls offers coverage numbers before the tests complete, it's a real timesaver

PR #482, opened 2015-08-06
Removed react entry from require_config.js in favor of reactaddons

noisecapella:
Remove `react` since we already have `reactaddons` and we shouldn't confuse the two


carsongee:
This seems fine, but do we maybe just want to change the name of reactaddons to react?  Just a thought

noisecapella:
Name was changed, any objections to merging?

carsongee:
Good to go :ship:

Issue #483, opened 2015-08-07
close panels with ESC key


pdpinch:
As a user, I would like to be able to close panels (like the learning resource panel) with a keyboard command. The ESC key has been suggested. 

Be sure to return focus to the listing view when the panel is closed.

pdpinch:
Closed via #484

PR #484, opened 2015-08-07
Closing panels with ESC key

amir-qayyum-khan:
Hi,

In this PR i added event handlers for escape button (keyboard) to close panels.

https://github.com/mitodl/lore/issues/483
@pdpinch @carsongee @noisecapella

### DIFF
--- a/ui/static/ui/js/listing.js
+++ b/ui/static/ui/js/listing.js
@@ -79,7 +79,24 @@ define('listing',
     $(document).ready(function() {
 
       $('[data-toggle=popover]').popover();
-
+      //Close panels on escape keypress
+      $(document).keyup(function(event) {
+        if (event.keyCode === 27) { // escape key maps to keycode `27`
+          if ($('.cd-panel').hasClass('is-visible')) {
+            $('.cd-panel').removeClass('is-visible');
+          }
+          if ($('.cd-panel-2').hasClass('is-visible')) {
+            $('.cd-panel-2').removeClass('is-visible');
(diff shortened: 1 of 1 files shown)


noisecapella:
This doesn't work on Firefox

noisecapella:
It seems that there should be only one key handler and it should close whatever panel happens to be open at the time. You should have only one event handler for this. The `keyup` function will attach event handlers without removing any so on repeated panel opens you will have multiple event handlers being called for every key press.

amir-qayyum-khan:
@noisecapella it is working on firefox on mac.

amir-qayyum-khan:
@noisecapella centralised code to avoid multiple event binding.

ui/static/ui/js/listing.js
```diff
@@ -80,6 +80,23 @@ define('listing',
 
       $('[data-toggle=popover]').popover();
 
+       $(document).keyup(function(e) {
```
noisecapella: Please add a comment before this block saying 'Close panels on escape keypress'. You will also need to fix indentation here

noisecapella:
This works for me now on Firefox, not sure why it wasn't working before.

amir-qayyum-khan:
Done @noisecapella

noisecapella:
Can you git amend and repush? Looks like a connection timeout on the tests, probably unrelated to this

noisecapella:
I think this is ready to merge when tests pass

PR #485, opened 2015-08-07
Updated indexing caching from dict to Django's cache.

### LINKED ISSUE #467: Use Django's caching for indexing data caching.
As an upgrade to #460, use Django's built-in caching (with its [expiration support](https://docs.djangoproject.com/en/1.8/topics/cache/#cache-arguments) to replace the dicts in `search/search_indexes.py`. 

This will make the code simpler. This may also mean replacing the `ALLOW_CACHING` setting that was added in that PR in favor of a `CACHE_TIMEOUT` setting used to populate the Django caching `TIMEOUT` setting. This would be zero (no caching) by default and could be changed as appropriate in production.

ShawnMilo:
Closes #467.

Improves upon PR #460 by getting rid of a "global" variable and
a custom settings.py value. Also adds tests which confirm that
fewer queries are done when the caching is used.

### DIFF
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -353,5 +353,15 @@ HAYSTACK_CONNECTIONS = {
     }
 }
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
-ALLOW_CACHING = get_var("ALLOW_CACHING", get_var("ALLOW_CACHING", False))
+
 XANALYTICS_URL = get_var('XANALYTICS_URL', "")
+
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    },
+    "lore_indexing": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        "TIMEOUT": get_var("LORE_INDEXING_CACHE_TIMEOUT", "60"),
+    }
+}
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -15,41 +15,17 @@ the index. They can not be mixed.
 from __future__ import unicode_literals
 
 from collections import defaultdict
-from datetime import datetime, timedelta
 import logging
 
-from django.conf import settings
+from django.core.cache import caches
 from haystack import indexes
 from lxml import etree
 
 from learningresources.models import Course, LearningResource
 
-INDEX_CACHE = {
-    "course": {},
-    "term": {},
-    "born": datetime.now(),
-}
-MAX_INDEX_AGE = timedelta(minutes=1)
-
 log = logging.getLogger(__name__)
 
-
-def check_cache_age():
-    """
-    Keep the cache from getting stale.
-
-    Must not cache during tests or development, or it will
-    cause confusion. Must manually enable caching in production.
-    """
-    # pylint: disable=global-statement
-    global INDEX_CACHE
-    if (datetime.now() - INDEX_CACHE["born"] > MAX_INDEX_AGE) \
-            or settings.ALLOW_CACHING is False:
-        INDEX_CACHE = {
-            "course": {},
-            "term": {},
-            "born": datetime.now(),
-        }
+cache = caches["lore_indexing"]
 
 
 def get_course_metadata(course_id):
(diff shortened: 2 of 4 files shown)


lore/settings.py
```diff
@@ -353,4 +353,10 @@ def get_var(name, default):
     }
 }
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
-ALLOW_CACHING = get_var("ALLOW_CACHING", get_var("ALLOW_CACHING", False))
+
+CACHES = {
+    "default": {
```
carsongee, line 360: So as suggested in stand up, I think it would be more flexible if we use a dedicated cache instead of the default for this in case we find need to have a more general use redis or memcached based cache later.

search/search_indexes.py
```diff
@@ -60,36 +34,57 @@ def get_course_metadata(course_id):
...
         data["repo_slug"] = course.repository.slug
-        INDEX_CACHE["course"][course_id] = data
+        cache.set(key, data)
+    # Return `data` directly, not from the cache. Otherwise, if caching
+    # is disabled (TIMEOUT == 0), this will always return nothing.
     return data
 
 
-def get_vocabs(course_id, lid):
+def get_vocabs(course_id, resource_id):
```
carsongee: I don't see a good reason to get hung up on this, but this is an "API" breaking change, have you double checked all the callers and that they are just using args and not kwargs?

carsongee: lol, quotes should have been on "breaking" there.

carsongee:
I am still hesitant to want to use this cache in production because we aren't doing are tests with it enabled.  What are your thoughts on that?  For example, what tests fail/what breaks if we turn this on to 30 minutes for example?   For the specific case of importing a course, it would be nice if we could just turn on a cache not based on time, but that turns on and turns off during import.  Particularly since for large courses we would need this cache to potentially be set to hours

carsongee:
:+1: other than the cache name.

lore/settings.py
ShawnMilo, line 360: Yep, that was done today. Still testing and about to flatten.

search/search_indexes.py
ShawnMilo: No breaking at all. This function is only used in this file.

ShawnMilo:
I have tests for the cache being on and off. I don't think the cache should be set to too long, mostly because any changes made to the `LearningResource`s (especially taxonomy stuff) would then be lost until a future re-index.

Also, it doesn't take too long to re-fill the cache, since it's filling the cache for the entire course at once. What this saves us is doing multiple database queries per `LearningResource` during import.

carsongee:
This was the only test I had fail with the cache turned on by default:

```
__________________________________________________________________________________ TestSearchView.test_terms_with_spaces ___________________________________________________________________________________

self = <search.tests.test_search_view.TestSearchView testMethod=test_terms_with_spaces>

    def test_terms_with_spaces(self):
        """
            Terms with spaces should show up in facet list correctly.
            """
        for term in self.terms:
            self.resource.terms.add(term)
        resp = self.client.get(reverse("repositories", args=(self.repo.slug,)))
>       self.assertContains(resp, "easy")

search/tests/test_search_view.py:45:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.tox/py27/local/lib/python2.7/site-packages/django/test/testcases.py:358: in assertContains
    msg_prefix + "Couldn't find %s in response" % text_repr)
E   AssertionError: Couldn't find 'easy' in response
```

It would be cool if we could maybe figure out how to make that pass in a sane way with the cache, and then we can just have it on by default, which would be nice for configuration and testing.

ShawnMilo:
@carsongee: Thanks for that last comment. It helped me figure out a way around it.

Since adding a term to a `LearningResource` requires us to use a custom signal, I just triggered a cache update for the single `LearningResource` in that signal. Minimal additional queries, but taxonomies should now _never_ be stale.

lore/settings.py
```diff
@@ -353,5 +353,16 @@ def get_var(name, default):
...
 XANALYTICS_URL = get_var('XANALYTICS_URL', "")
+
+CACHES = {
+    "default": {
+        "BACKEND": get_var(
+            "CACHE_BACKEND", "django.core.cache.backends.locmem.LocMemCache")
+    },
+    "indexing": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        "TIMEOUT": get_var("CACHE_TIMEOUT", "60"),
```
carsongee: Maybe `LORE_INDEX_CACHE_TIMEOUT` to avoid name collisions

lore/settings.py
```diff
@@ -353,5 +353,16 @@ def get_var(name, default):
...
 }
 HAYSTACK_SIGNAL_PROCESSOR = 'haystack.signals.RealtimeSignalProcessor'
-ALLOW_CACHING = get_var("ALLOW_CACHING", get_var("ALLOW_CACHING", False))
+
 XANALYTICS_URL = get_var('XANALYTICS_URL', "")
+
+CACHES = {
+    "default": {
+        "BACKEND": get_var(
+            "CACHE_BACKEND", "django.core.cache.backends.locmem.LocMemCache")
```
carsongee: similar, maybe `LORE_DEFAULT_CACHE_BACKEND`?

carsongee:
Other than the minor setting name changes, after tests pass :+1:

lore/settings.py
ShawnMilo: I assume this is in case we move from LocMem.

ShawnMilo: I misunderstood your comment, but it makes sense to update `indexing` to be more specific also. I'll change both.

lore/settings.py
ShawnMilo: How about just removing the `get_var` altogether until we have a reason for it?

carsongee: sounds good, while we're at it, do we even need to define the default cache at all?

ShawnMilo: Yes, unfortunately. If you define anything and don't define `default`, it [blows up](https://github.com/django/django/blob/master/django/core/cache/__init__.py#L33).

ShawnMilo:
@carsongee: I left the latest changes as separate commits so you can let me know if your :+1: is still valid. If so, please remove "review needed" and I'll flatten after tests pass. Thanks. :o)

carsongee:
Yep, good to go

carsongee:
whoops, forgot to flip the flag

ShawnMilo:
I assume "waiting on author" can be removed, since we're just waiting for tests, right?

carsongee:
Yeah, we didn't have a label for that, so waiting on author seemed closest since it triggers doof to notify you.

ShawnMilo:
Roger. I get e-mail from github more reliably than I'm in HipChat, so even a comment works for me. :o)

PR #498, opened 2015-08-11
Split builds and removed python 3.3 testing

carsongee:
As discussed in our team meeting, this removes the tests for python 3.3.  It also splits the tests onto different workers for each test environment so that the testing time should just be the maximum of the environments (currently 3.4).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ services:
 before_script:
   - sleep 10
 env:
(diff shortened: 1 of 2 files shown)


carsongee:
It looks like the coverage is just from whichever environment finished first (javascript in this case).  I'll need to look into that.

carsongee:
It turns out, travis will split out all the deploys as well, so I'm removing the matrix until that is handled better.

PR #500, opened 2015-08-11
Taxonomy "double dropdown" editing interface on Learning Resources

jamiefolsom:
Hi @carsongee,

Thanks for offering to have a look at this PR. Couple of items to note.
- I have not written any tests yet;
- My commits not squashed;
- Otherwise, seems to work for the UX as I understand it.

Happy for your thoughts on style, terseness, functionality, or anything else.

Jamie

### DIFF
--- a/ui/jstests/test-learning-resource.jsx
+++ b/ui/jstests/test-learning-resource.jsx
@@ -3,7 +3,9 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
   QUnit, $, React, _, LearningResources, TestUtils) {
   'use strict';
 
-  var VocabularyOption = LearningResources.VocabularyOption;
+  var VocabSelect = LearningResources.VocabSelect;
+  var TermList = LearningResources.TermList;
+  var TermSelect = LearningResources.TermSelect;
   var LearningResourcePanel = LearningResources.LearningResourcePanel;
   var waitForAjax = TestUtils.waitForAjax;
 
@@ -35,13 +37,13 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     "label": "hard",
     "weight": 1
   };
-  var prereqTermsResponseRequired = {
+  var preReqTermsResponseReq = {
     "id": 3,
     "slug": "required",
     "label": "required",
     "weight": 1
   };
-  var prereqTermsResponseNotRequired = {
+  var preReqTermsResponseNotReq = {
     "id": 4,
     "slug": "notrequired",
     "label": "notrequired",
@@ -55,11 +57,11 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     "vocabulary_type": "m",
     "required": false,
     "weight": 2147483647,
-    "terms": [prereqTermsResponseRequired, prereqTermsResponseNotRequired],
+    "terms": [preReqTermsResponseReq, preReqTermsResponseNotReq],
     "multi_terms": true
   };
   var vocabularyResponseDifficulty = {
-    "id": 1,
+    "id": 2,
     "slug": "difficulty",
     "name": "difficulty",
     "description": "Difficulty",
@@ -67,8 +69,21 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     "required": false,
     "weight": 2147483647,
     "terms": [termResponseEasy, termResponseHard],
-    "multi_terms": false
+    "multi_terms": true
   };
+  var selectedVocabulary = vocabularyResponseDifficulty;
+  var vocabulariesAndTerms = [
+    {
+      "terms": [preReqTermsResponseReq, preReqTermsResponseNotReq],
+      "selectedTerms": [preReqTermsResponseReq, preReqTermsResponseNotReq],
+      'vocabulary': vocabularyResponsePrereq,
+    },
+    {
+      "terms": [termResponseEasy, termResponseHard],
+      "selectedTerms": [termResponseEasy, termResponseHard],
+      'vocabulary': vocabularyResponseDifficulty,
+    }
+  ];
   var vocabulariesResponseFirst = {
     "count": 1,
     "next": "/api/v1/repositories/repo/vocabularies/?type_name=course&page=2",
@@ -80,12 +95,6 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     "next": null,
     "previous": "/api/v1/repositories/repo/vocabularies/?type_name=course",
     "results": [vocabularyResponsePrereq]
-  };
-  var difficultyTermsResponse = {
-    "count": 2,
-    "next": null,
-    "previous": null,
-    "results": [termResponseEasy, termResponseHard]
   };
 
   QUnit.module('Test learning resource panel', {
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
   });
 
   QUnit.test(
-    'Assert that VocabularyOption renders properly',
+    'Assert that VocabSelect renders properly',
     function (assert) {
       var done = assert.async();
-      var afterMount = function (component) {
+
+      var afterMount = function(component) {
         var $node = $(React.findDOMNode(component));
 
-        // one vocabulary
         var $vocabSelect = $node.find("select");
         assert.equal($vocabSelect.size(), 1);
-
-        // two terms
-        var $termsSelect = $vocabSelect.find("option");
-        assert.equal($termsSelect.size(), 2);
-        assert.equal($termsSelect[0].text, "easy");
-        assert.equal($termsSelect[0].value, "easy");
-        assert.equal($termsSelect[1].text, "hard");
-        assert.equal($termsSelect[1].value, "hard");
 
         done();
       };
 
       React.addons.TestUtils.renderIntoDocument(
-        <VocabularyOption vocabulary={vocabularyResponseDifficulty}
-                          terms={difficultyTermsResponse.results}
-                          selectedTerm="hard"
-                          ref={afterMount}/>
+        <VocabSelect
+          vocabs={vocabulariesAndTerms}
+          selectedVocabulary={selectedVocabulary}
+          ref={afterMount}
+        />
       );
     }
   );
 
   QUnit.test(
-    'Assert that LearningResourcePanel changes state properly',
-    function(assert) {
+    'Assert that TermSelect renders properly',
+    function (assert) {
       var done = assert.async();
 
       var afterMount = function(component) {
+        var $node = $(React.findDOMNode(component));
+
+        var $termSelect = $node.find("select");
+        assert.equal($termSelect.size(), 1);
+
+        done();
+      };
+
+      React.addons.TestUtils.renderIntoDocument(
+        <TermSelect
+          vocabs={vocabulariesAndTerms}
+          selectedVocabulary={selectedVocabulary}
+          ref={afterMount}
+        />
+      );
+    }
+  );
+
+  QUnit.test(
+    'Assert that TermList renders properly',
+    function (assert) {
+      var done = assert.async();
+      var afterMount = function(component) {
+        var $node = $(React.findDOMNode(component));
+        var $termList = $node.find("ul");
+        assert.equal($termList.size(), 1);
+        done();
+      };
+
+      React.addons.TestUtils.renderIntoDocument(
+        <TermList
+          vocabs={vocabulariesAndTerms}
+          ref={afterMount}
+        />
+      );
+    }
+  );
+  QUnit.test(
+    'Assert that LearningResourcePanel changes state properly',
+    function(assert) {
+      var done = assert.async();
+      var afterMount = function(component) {
         // wait for calls to populate form
         waitForAjax(3, function () {
-          // one vocabulary
+          // two menus: vocabulary and terms.
           var $node = $(React.findDOMNode(component));
-          var $vocabSelect = $node.find("select");
-          assert.equal($vocabSelect.size(), 2);
 
-          // two terms, first vocab
-          var $terms1Select = $($vocabSelect[0]).find("option");
-          assert.equal($terms1Select.size(), 2);
-          assert.equal($terms1Select[0].selected, false);
-          assert.equal($terms1Select[0].selected, false);
+          var $allSelects = $node.find("#vocabularies select");
+          assert.equal($allSelects.size(), 2);
+
+          var $vocabSelect = $node.find($allSelects).first();
+          assert.equal($vocabSelect.size(), 1);
+
+          // first vocab, two options
+          var $vocabOptions = $vocabSelect.find("option");
+          assert.equal($vocabOptions.size(), 2);
+
+          assert.equal($vocabOptions[0].selected, true);
+          assert.equal($vocabOptions[1].selected, false);
 
           // TestUtils.Simulate.change only simulates a change event,
           // we need to update the value first ourselves
-          $($vocabSelect[0]).val("hard").trigger('change');
-          React.addons.TestUtils.Simulate.change($vocabSelect[0]);
+          $vocabSelect.val("prerequisite").trigger('change');
+          React.addons.TestUtils.Simulate.change($vocabSelect);
           component.forceUpdate(function() {
-            assert.equal($terms1Select[0].selected, false);
-            assert.equal($terms1Select[1].selected, true);
-            // make sure second vocab (which has a default value) is set properly
-            var $terms2Select = $($vocabSelect[1]).find("option");
-            assert.equal($terms2Select.size(), 2);
-            assert.equal($terms2Select[0].selected, true);
-            assert.equal($terms2Select[1].selected, false);
-            //the second vocabulary can be a multi select
-            $($vocabSelect[1])
-              .val(["notrequired", "required"])
+            assert.equal($vocabOptions[0].selected, false);
+            assert.equal($vocabOptions[1].selected, true);
+            // on selection of the second vocabulary, make sure the term selector updates.
+            var termsSelect = $allSelects[1];
+            var $termsOptions = $(termsSelect).find("option");
+            assert.equal($termsOptions.size(), 2);
+            assert.equal($termsOptions[0].selected, false);
+            assert.equal($termsOptions[1].selected, false);
+            // the second vocabulary can be a multi select
+            $(termsSelect)
+              .val(["hard", "easy"])
               .trigger('change');
-            React.addons.TestUtils.Simulate.change($vocabSelect[1]);
+            React.addons.TestUtils.Simulate.change(termsSelect);
             component.forceUpdate(function() {
-              assert.equal($terms2Select[0].selected, true);
-              assert.equal($terms2Select[1].selected, true);
-              //be sure that the state reflects the selection
+              assert.equal($termsOptions[0].selected, true);
+              assert.equal($termsOptions[1].selected, true);
+              // be sure that the state reflects the selection
               var terms = _.map(component.state.vocabulariesAndTerms,
                 function (tuple) {
                   return tuple.selectedTerms;
@@ -196,13 +246,11 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
               terms = _.flatten(terms);
               terms.sort();
               var expectedTerms = [
-                $terms1Select[1].value,
-                $terms2Select[0].value,
-                $terms2Select[1].value
+                $termsOptions[0].value,
+                $termsOptions[1].value
               ];
               expectedTerms.sort();
               assert.deepEqual(terms, expectedTerms);
-
               done();
             });
           });
@@ -215,7 +263,6 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
         ref={afterMount} />);
     }
   );
-
   QUnit.test(
     'Assert that LearningResourcePanel saves properly',
     function(assert) {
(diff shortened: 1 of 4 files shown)


pdpinch:
Not sure what the etiquette is here, but since you're looking for feedback, I added the "Needs Review" label.

But I also added the "work in progress" label too.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,160 @@ define('learning_resources', [
...
+              terms: tuple.terms,
+              selectedTerms: selectedTerms
+            };
+            return newTuple;
+          } else {
+            return tuple;
+          }
+        });
+
+      thiz.setState({
```
carsongee: I don't think you should need to use `thiz` for these unless you need to access the parent `this` from within a closure.

jamiefolsom: Good point. Fix via new commit just pushed.

amir-qayyum-khan:
@jamiefolsom I added 4 Vocabularies

screen shot 2015-08-12 at 5 27 36 pm
 
##### But it is showing me only 2 vocabularies in learning resource panel also i am not able to search other Vocabularies

screen shot 2015-08-12 at 5 27 44 pm

screen shot 2015-08-12 at 5 34 06 pm

pdpinch:
FYI, vocabulary facets don't show up in the listing view unless they are part of the result set. 

In other words, you need to add a vocabulary term to at least one learning resource before the vocabulary will appear in the left nav.

amir-qayyum-khan:
@pdpinch  here i am not referring to left nav, I am talking about learning resource panel.  it is not showing all vocabularies that i added in taxonomy panel.

amir-qayyum-khan:
@jamiefolsom 

In **Learning resource panel**, when i selects **second vocabulary** it shows me list of **terms** of first and second  **vocabularies**. 
- Should it only show list of **terms** related to that vocabulary?
- When I selects 3rd or nth vocabulary then it show me **terms** of  first and that selected **vocabulary**.  Is this right behaviour?

screen shot 2015-08-12 at 5 50 07 pm

screen shot 2015-08-12 at 5 50 25 pm

amir-qayyum-khan:
@jamiefolsom 
When I selected **colour** as **vocabulary** and  choosed  **terms** of **vocabulary name** in second field and press **save**. It has not given me any error.

screen shot 2015-08-12 at 5 59 05 pm

screen shot 2015-08-12 at 5 59 17 pm

amir-qayyum-khan:
@jamiefolsom 
When I selected **colour** as **vocabulary** and  choosed  **terms** of **vocabulary "name"** in second field then saved and closed the panel. 
When i reopened it terms are allocated to  **vocabulary "name"** instead of colour. 

##### This seem good but it should inform user if he is mapping wrong terms with vocabulary or it should only show relevant terms against a vocabulary.

screen shot 2015-08-12 at 5 59 05 pm

screen shot 2015-08-12 at 6 03 37 pm

jamiefolsom:
Hi @amir-qayyum-khan, thanks for taking a look! You should only see terms in the term menu for the specific vocabulary you selected from the vocabulary menu. Sounds like a bug; I will look into it.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,156 @@ define('learning_resources', [
...
+          <strong>{this.props.label}: </strong> {this.props.terms}
+        </li>
+      );
+    }
+  });
+
+  var VocabSelect2 = React.createClass({
+
+    render: function () {
+
```
amir-qayyum-khan: Please remove blank line after render?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,156 @@ define('learning_resources', [
...
+      return (
+        <li className="applied-term list-group-item">
+          <strong>{this.props.label}: </strong> {this.props.terms}
+        </li>
+      );
+    }
+  });
+
+  var VocabSelect2 = React.createClass({
+
```
amir-qayyum-khan: Please remove blank line before render?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,156 @@ define('learning_resources', [
...
-      var options = _.map(this.props.terms, function (term) {
+      return (
+        <li className="applied-term list-group-item">
+          <strong>{this.props.label}: </strong> {this.props.terms}
+        </li>
+      );
+    }
+  });
+
+  var VocabSelect2 = React.createClass({
```
amir-qayyum-khan: Why there is 2 with selector name? Should it be more clear like VocabSelect, VocabSelectForLR or any other?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,156 @@ define('learning_resources', [
...
+    handleChange: function(e) {
+      var selectedValue = _.pluck(
+        _.filter(e.target.options, function(option) {
+          return option.selected && option.value !== null;
+        }), 'value');
+      this.props.updateVocabulary(_.pluck(this.props.vocabs, 'vocabulary'), selectedValue[0]);
+    }
+  });
+
+  var TermSelect2 = React.createClass({
```
amir-qayyum-khan: Why 2 is append with TermSelect because i havent seen any other TermSelect in the same file. Should it be more clear like TermSelect or TermSelectForLR or LRTermSelect or any other

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,156 @@ define('learning_resources', [
...
+
+    render: function () {
+
```
amir-qayyum-khan: please remove this line

ui/static/ui/js/learning_resources.jsx
jamiefolsom: The "2"s refer to the fact that the components are built with the Select2 component, in utils. But you're right -- that's unnecessary. I'll change those names.

ui/static/ui/js/learning_resources.jsx
jamiefolsom: Will do.

ui/static/ui/js/learning_resources.jsx
jamiefolsom: Yep -- there were comments in there, now gone -- will tighten up the whitespace. Thx.

ui/static/ui/js/learning_resources.jsx
jamiefolsom: Line 52?

jamiefolsom:
Hi, @amir-qayyum-khan thanks for catching that issue (terms from other vocabularies getting duplicated). See [latest commit](https://github.com/mitodl/lore/commit/8ea6cc6e74c6b50e08355add33d82ccc197ae768) for the solution; I was not setting a `key` prop on the dynamically updating child component, TermSelect; without that, React didn't know how to update it properly. cc @pdpinch and @carsongee.

carsongee:
Cool, thanks for tagging me in with the answer.  I didn't know about the key prop

pdpinch:
Is that `key` prop a generic react requirement, or something specific to the Select2 implementation?

ShawnMilo:
`key` is a generic React thing.

jamiefolsom:
Yes, it's generic, but as I understand it, it's not required in all cases, only those in which the component has dynamic content, updated by a parent.

The fact that it's required in that case seems sort of odd, since (at least from my pt. of view), the expected behavior would be for an updated prop to result in a updated UI.

But I think it's because React is trying to be efficient in terms of how much of the DOM to change, so it reuses existing DOM nodes rather than replacing/deleting them, which in turn means it needs unique identifiers. Is that right, @ShawnMilo?

ShawnMilo:
@jamiefolsom I don't know all the details. I just remember seeing it mentioned in one of the videos and that it had to be unique. I assumed from context it's exactly what you said (used to determine what parts of the DOM need updating), but I don't know of situations where it's optional.

I just scanned the video again and the brief explanation given is that you need it any time you have an array, so React can identify the same elements in an updated version of the array.

jamiefolsom:
Makes sense. Thanks, @ShawnMilo.

ui/static/ui/js/utils.jsx
```diff
@@ -109,8 +109,13 @@ define("utils", ["jquery", "lodash", "reactaddons", "select2"],
       //this function can be used only if the component has been mounted
       var thiz = this;
       var isMultiTerms = this.props.multiple;
+
       //not allowing the clear for the multi terms dropdown
-      var allowClear = !isMultiTerms;
+      var allowClear;
+      if (this.props.allowClear === undefined) {
```
noisecapella, line 115: It seems like this needs an `else` block to assign allowClear

ui/static/ui/css/mit-lore.css
```diff
@@ -498,6 +498,32 @@ ul.with-utility-features {
     word-wrap: break-word;
 }
 
+div#term-list-container .panel-heading {
```
noisecapella, line 500: Can you change the indenting to match the rest of the CSS?

ui/static/ui/js/utils.jsx
jamiefolsom, line 115: Good catch.

ui/static/ui/css/mit-lore.css
jamiefolsom, line 500: Sure thing, will do.

ui/static/ui/js/utils.jsx
jamiefolsom, line 115: So correct me if I am wrong, but as written, this conditional essentially sets a default, if a value is _not_ passed into the component when it's created, so an else here would be redundant.

noisecapella, line 115: `this.props.allowClear` is only being tested against `undefined`. What if the caller passes `allowClear={true}`?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,179 @@ define('learning_resources', [
...
+        if (selectedTerms.length) {
+          return (
+            <TermListItem
+              key={vocab.vocabulary.id}
+              label={vocabularyName}
+              terms={selectedTerms.join(", ")}
+            />
+          );
+        }
+      }, this);
```
noisecapella: It doesn't seem like this argument is necessary here since there is no `this` within the closure

jamiefolsom: Which one?

jamiefolsom: ah, `this`?

ui/static/ui/js/utils.jsx
jamiefolsom, line 115: Right. Didn't notice the later code relying on allowClear, rather than this.props.allowClear. Thanks!

ui/static/ui/css/mit-lore.css
jamiefolsom, line 500: Changed tabs to four spaces -- let me know if that fixed what you were seeing. Thanks.

jamiefolsom:
Hey @pdpinch, I've addressed @noisecapella's suggestions. Others are of course welcome. I'm removing the WIP tag, since this is under review. It's rebased on master; only other outstanding step that I know of is to squash commits.

jamiefolsom:
@bdero if you have a sec, it appears that a Travis build is still failing on this PR. I haven't rerun, or pushed a new commit. See https://travis-ci.org/mitodl/lore/builds/76001652#L866. It appears unrelated to my PR, but any thoughts you may have to offer would be welcome. Thanks!

pdpinch:
You failed a test, which appears to be unrelated & due to a publicsuffixes.org being down:

> it appears that https://publicsuffix.org/list/effective_tld_names.dat has either changed or was offline, it appears up now, retrying

I'm rerunning Travis now.

jamiefolsom:
@pdpinch Thanks!

noisecapella:
I still need to look over this, I haven't made it completely through yet

ui/static/ui/js/utils.jsx
```diff
@@ -109,8 +109,13 @@ define("utils", ["jquery", "lodash", "react", "select2"],
       //this function can be used only if the component has been mounted
       var thiz = this;
       var isMultiTerms = this.props.multiple;
+
       //not allowing the clear for the multi terms dropdown
```
noisecapella: Can you update the comment according to the change you made?

jamiefolsom: Done.

noisecapella:
When I run the tests I get this warning:

```
Warning: flattenChildren(...): Encountered two children with the same key, `.$1`. Child keys must be unique; when two children share a key, only the first child will be used.
```

I narrowed it down to 'Assert that TermList renders properly'. Can you investigate? Duplicate keys are bad since React is effectively throwing one element away.

noisecapella:
I've gotten the key error before in testing, it's usually just something to do with the test data being mocked.

jamiefolsom:
Looks like the mock data in the test has two vocabularies with the same value for id. See: https://github.com/mitodl/lore/blob/feature/jf/taxonomy_double_dropdown/ui/jstests/test-learning-resource.jsx#L52-L73

I am going to change that, and see if it fixes this.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,179 @@ define('learning_resources', [
...
+        <li className="applied-term list-group-item">
+          <strong>{this.props.label}: </strong> {this.props.terms}
+        </li>
+      );
+    }
+  });
+
+  var VocabSelect = React.createClass({
+    render: function () {
+      var options = [];
```
noisecapella: Can you combine this with the next line?

ui/static/ui/css/mit-lore.css
noisecapella, line 500: Yes that's fine

ui/static/ui/js/learning_resources.jsx
jamiefolsom: I broke them out because I was getting line length complaints from the linter.

ui/jstests/test-learning-resource.jsx
```diff
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
-          assert.equal($terms1Select[0].selected, false);
-          assert.equal($terms1Select[0].selected, false);
+          var $allSelects = $node.find("#vocabularies select");
+          assert.equal($allSelects.size(), 2);
+
+          var $vocabSelect = $node.find($allSelects).first();
+          assert.equal($vocabSelect.size(), 1);
+
+          // first vocab, two options
+          var $vocabOptions = $($vocabSelect).find("option");
```
noisecapella: If `$vocabSelect` is a jquery object you shouldn't need to use `$()` here

ui/jstests/test-learning-resource.jsx
```diff
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+          assert.equal($vocabOptions.size(), 2);
+
+          assert.equal($vocabOptions[0].selected, true);
+          assert.equal($vocabOptions[1].selected, false);
 
           // TestUtils.Simulate.change only simulates a change event,
           // we need to update the value first ourselves
-          $($vocabSelect[0]).val("hard").trigger('change');
-          React.addons.TestUtils.Simulate.change($vocabSelect[0]);
+          $($vocabSelect).val("prerequisite").trigger('change');
```
noisecapella: Same comment about `$()` here

ui/static/ui/js/learning_resources.jsx
noisecapella: Would it be possible to move the function argument to the next line instead? Or just use `var options;`

ui/static/ui/js/learning_resources.jsx
```diff
@@ -6,73 +6,179 @@ define('learning_resources', [
...
-        <label className="col-sm-6 control-label">
-        {this.props.vocabulary.name}
-          </label>
-        <div className="col-sm-6">
+      var currentVocabulary = {};
+      var selectedVocabulary = this.props.selectedVocabulary;
+
+      _.forEach(this.props.vocabs, function(vocab) {
+        if (vocab.vocabulary.slug === selectedVocabulary.slug) {
+          currentVocabulary = vocab;
```
noisecapella, line 99: Can you return false after this line? This will terminate the foreach early

ui/static/ui/js/learning_resources.jsx
```diff
@@ -188,8 +299,10 @@ define('learning_resources', [
             });
 
           thiz.setState({
-            vocabulariesAndTerms: vocabulariesAndTerms
+            vocabulariesAndTerms: vocabulariesAndTerms,
+            selectedVocabulary: vocabulariesAndTerms[0].vocabulary
```
noisecapella: What happens if there are no vocabularies or terms for a resource?

ui/jstests/test-learning-resource.jsx
```diff
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
-            assert.equal($terms2Select[0].selected, true);
-            assert.equal($terms2Select[1].selected, false);
-            //the second vocabulary can be a multi select
-            $($vocabSelect[1])
-              .val(["notrequired", "required"])
+            assert.equal($vocabOptions[0].selected, false);
+            assert.equal($vocabOptions[1].selected, true);
+            // on selection of the second vocabulary, make sure the term selector updates.
+            var $termsSelect = $allSelects[1];
+            var $termsOptions = $($termsSelect).find("option");
```
noisecapella: Remove `$()` here

ui/jstests/test-learning-resource.jsx
```diff
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+            assert.equal($termsOptions.size(), 2);
+            assert.equal($termsOptions[0].selected, false);
+            assert.equal($termsOptions[1].selected, false);
+            // the second vocabulary can be a multi select
+            $($termsSelect)
```
noisecapella: And `$()` here

ui/jstests/test-learning-resource.jsx
jamiefolsom: Did so, ran tests, and got: `TypeError: 'undefined' is not a function (evaluating '$termsSelect.find('option')')`

ui/jstests/test-learning-resource.jsx
```diff
@@ -119,76 +128,117 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+
+          assert.equal($vocabOptions[0].selected, true);
+          assert.equal($vocabOptions[1].selected, false);
 
           // TestUtils.Simulate.change only simulates a change event,
           // we need to update the value first ourselves
-          $($vocabSelect[0]).val("hard").trigger('change');
-          React.addons.TestUtils.Simulate.change($vocabSelect[0]);
+          $vocabSelect.val("prerequisite").trigger('change');
+          React.addons.TestUtils.Simulate.change($vocabSelect);
```
noisecapella, line 223: `Simulate.change` will not work on a jquery object (see also `$termsSelect` below)

ui/jstests/test-learning-resource.jsx
noisecapella: Ok, it looks like you should change `$termsSelect` to just `termsSelect` instead and leave the `$()` on

jamiefolsom: There's no var called termSelect:

```
ui/jstests/test-learning-resource.jsx: line 229, col 35, 'termsSelect' is not defined.
ui/jstests/test-learning-resource.jsx: line 234, col 15, 'termsSelect' is not defined.
```

noisecapella: I mean rename `$termsSelect` to `termSelect` since it isn't a jquery object.

noisecapella:
Code coverage looks good, functionality works excepting the case with zero vocabularies for a learning resource type. Clean implementation :)

ui/jstests/test-learning-resource.jsx
jamiefolsom: Ah, gotcha.

ui/static/ui/js/learning_resources.jsx
jamiefolsom, line 99: Done.

jamiefolsom:
Great, I will handle the zero vocabulary case, then squash commits. Thanks, @noisecapella!

ui/static/ui/js/learning_resources.jsx
jamiefolsom: This solves the proximate issue, avoiding accessing an attribute of an undefined object, but you raise a broader question, which we haven't dealt with, which is that the whole LR panel should not show this chooser UI if there are no vocabularies assigned to the resource type.

```
          if (vocabulariesAndTerms.length) {
            thiz.setState({
              selectedVocabulary: vocabulariesAndTerms[0].vocabulary
            });
          };
```

ui/jstests/test-learning-resource.jsx
noisecapella, line 223: It looks like you missed this here. The change event may be provided to React via `trigger()` in which case you can just remove the `Simulate.change` line.

noisecapella, line 223: I'm fine merging anyway, we can do this in a separate PR

jamiefolsom, line 223: OK, merging. Thanks!

noisecapella:
Just commenting on `key` in React. It's used to simplify the tree diff operation by matching up old elements with new ones in their fake DOM. There was a presentation as OSCON that claimed a naive tree diff is O(n^3) but I can't remember why at the moment.

PR #508, opened 2015-08-13
Added handling for Redis ResponseError.

ShawnMilo:
In addition to failure to connect, we could
connect then fail to provide a required password.

This is to handle the error that happened once the health check branch hit production:

```
ResponseError: NOAUTH Authentication required
```

That error, according to the [docs](http://www.tutorialspoint.com/redis/redis_security.htm), happens when the Redis server requires a password and it's not sent.


carsongee:
:first_quarter_moon:

PR #520, opened 2015-08-14
Added configuration option and heroku command to pre-compress assets

carsongee:
FYI @mitodl/odl-engineering this adds a little doc and config option for disabling static assets from being built locally on every request, which can be handy if you are working on something that isn't JS/React.


giocalitri:
Checked functionality locally and using heroku lore-branch-test: everything looks good. Merging

PR #522, opened 2015-08-17
Edit vocabulary in taxonomy panel

### LINKED ISSUE #504: Edit vocabulary in taxonomy panel
##### Requirement

Users may want to edit the vocabulary to , for example, change the types of learning resources that vocabulary applies to.
- [ ] Editing a vocabulary should open the Vocabulary tab and populate the form with current values.

Epic https://github.com/mitodl/lore/issues/380
##### Detail

On clicking edit vocabulary button, I will switch tab and open vocabulary in edit mode. Save button will be change into update. And will add cancel button along with update. On pressing cancel or update I will clear form and switch back to listing tab. In case of update I will update vocabulary. (Will see if there is any need to change API)

cc @pdpinch

### LINKED ISSUE #632: There's only one taxonomy per repository (and multiple vocabularies)
- [x] Change the label for the Manage Taxonomies button to `Manage Taxonomy` 
- [x] Change the title for the Manage Taxonomies panel to `Manage Taxonomy'
- [x] Change the label of the Taxonomies tab to `Vocabularies` (part of PR #522)

amir-qayyum-khan:
Hi,

In this PR i wrote code to edit a vocabulary in taxonomy panel.
Anyone can review it that will be great.

fixes https://github.com/mitodl/lore/issues/504 
fixes https://github.com/mitodl/lore/issues/632
@pdpinch @noisecapella @giocalitri 

screen shot 2015-08-17 at 9 30 38 pm

screen shot 2015-08-17 at 9 30 45 pm

screen shot 2015-08-17 at 9 31 06 pm

screen shot 2015-08-17 at 9 37 13 pm

screen shot 2015-08-17 at 9 32 52 pm

### DIFF
--- a/ui/static/ui/js/manage_taxonomies.jsx
+++ b/ui/static/ui/js/manage_taxonomies.jsx
@@ -1,5 +1,6 @@
-define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
-  function (React, _, $, Utils) {
+define('manage_taxonomies', ['react', 'lodash', 'jquery', 'uri',
+    'utils', 'bootstrap'],
+  function (React, _, $, URI, Utils) {
   'use strict';
 
   var StatusBox = Utils.StatusBox;
@@ -10,9 +11,7 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
     getInitialState: function() {
       return {
         formatActionState: 'show',
-        label: this.props.term.label,
-        errorMessage: '',
-        showError: false
+        errorMessage: ''
       };
     },
     render: function () {
@@ -37,12 +36,12 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
         deleteButtonClass = 'revert-button disabled';
       } else if (this.state.formatActionState === 'show') {
         label = <label className="term-title"
-          htmlFor="minimal-checkbox-1-11">{this.state.label}</label>;
+          htmlFor="minimal-checkbox-1-11">{this.props.term.label}</label>;
         formatActionClassName = "fa fa-pencil no-select";
       }
 
       var listClassName = "form-group";
-      if (this.state.showError) {
+      if (this.state.errorMessage) {
         listClassName = "form-group has-error";
       }
 
@@ -61,8 +60,9 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
      */
     editTerm: function() {
       this.setState({
-          formatActionState: 'edit'
-        }, function() {
+        formatActionState: 'edit',
+        label: this.props.term.label
+      }, function () {
         var $editText = $(React.findDOMNode(this.refs.editText));
         $editText.focus();
       });
@@ -95,11 +95,11 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
      */
     updateTermSubmit: function() {
       if (_.isEmpty(this.state.label)) {
-        this.setState({showError: true});
+        this.setState({errorMessage: "Term cannot be empty."});
         return;
       }
       var API_ROOT_VOCAB_URL = '/api/v1/repositories/' + this.props.repoSlug +
-        '/vocabularies/' + this.props.vocabularySlug + '/terms/' +
+        '/vocabularies/' + this.props.vocabulary.slug + '/terms/' +
         this.props.term.slug + "/";
       var thiz = this;
       var termData = {
@@ -113,12 +113,11 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
         contentType: "application/json"
       }).fail(function() {
         thiz.setState({
-          showError: true,
           errorMessage: 'Unable to update term'
         });
       }).done(function(term) {
         thiz.resetUtilityFeatures();
-        thiz.props.updateTerm(thiz.props.vocabularySlug, term);
+        thiz.props.updateTerm(thiz.props.vocabulary.id, term);
         thiz.props.refreshFromAPI();
       });
     }
@@ -131,10 +130,10 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
       var items = _.map(this.props.terms, function (term) {
         return <TermComponent
           updateTerm={thiz.props.updateTerm}
-          vocabularySlug={thiz.props.vocabulary.slug}
+          vocabulary={thiz.props.vocabulary}
           repoSlug={thiz.props.repoSlug}
           term={term}
-          key={term.slug}
+          key={term.id}
           refreshFromAPI={thiz.props.refreshFromAPI}
           />;
       });
@@ -143,21 +142,22 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
           <div className="panel-heading">
             <h4 className="panel-title">
               <span className="utility-features">
-                <a href="#">
+                <a href="#" onClick={this.onEdit}>
                   <i className="fa fa-pencil"></i>
-                </a> <a data-toggle="modal" data-target="#confirm-delete"
+                </a>
+                <a data-toggle="modal" data-target="#confirm-delete"
                   onClick={this.onDeleteHandler} className="delete-vocabulary">
                   <i className="fa fa-remove"></i>
                 </a>
               </span> <a className="accordion-toggle vocab-title"
                          data-toggle="collapse"
                          data-parent="#accordion"
-                         href={'#collapse-' + this.props.vocabulary.slug}>
+                         href={'#collapse-' + this.props.vocabulary.id}>
                 {this.props.vocabulary.name}
               </a>
             </h4>
           </div>
-          <div id={'collapse-' + this.props.vocabulary.slug}
+          <div id={'collapse-' + this.props.vocabulary.id}
                className="panel-collapse collapse in">
             <div className="panel-body">
               <ul className="icheck-list with-utility-features">
@@ -208,14 +208,19 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
           var message = "Unable to delete vocabulary '" + vocab.name + "'";
           thiz.props.reportMessage({error: message});
         }).then(function () {
-          thiz.props.deleteVocabulary(vocab.slug);
+          thiz.props.deleteVocabulary(vocab.id);
 
           thiz.props.refreshFromAPI();
         });
       }
     },
+    onEdit: function(e) {
+      e.preventDefault();
+      this.props.editVocabulary(this.props.vocabulary.id);
+    },
     onKeyUp: function(e) {
       if (e.key === "Enter") {
+        e.preventDefault();
         this.handleAddTermClick();
       }
     },
@@ -237,8 +242,8 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
             error: "Error occurred while adding new term."
           });
         })
-      .done(function(newTerm) {
-          thiz.props.addTerm(thiz.props.vocabulary.slug, newTerm);
+      .then(function(newTerm) {
+          thiz.props.addTerm(thiz.props.vocabulary.id, newTerm);
           thiz.setState({
             newTermLabel: null
           });
@@ -264,8 +269,9 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
           updateTerm={thiz.props.updateTerm}
           vocabulary={obj.vocabulary}
           deleteVocabulary={thiz.props.deleteVocabulary}
+          editVocabulary={thiz.props.editVocabulary}
           terms={obj.terms}
-          key={obj.vocabulary.slug}
+          key={obj.vocabulary.id}
           repoSlug={repoSlug}
           reportMessage={thiz.reportMessage}
           addTerm={thiz.props.addTerm}
@@ -276,13 +282,13 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
       return <div className="panel-group lore-panel-group">
         <div className="panel panel-default">
         </div>
-        <StatusBox message={this.state.message} />
+        <StatusBox message={this.props.message} />
         {items}
       </div>;
 
     },
     reportMessage: function(message) {
-      this.setState({message: message});
+      this.props.reportMessage(message);
     },
     getInitialState: function() {
       return {
@@ -294,17 +300,10 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
   var AddVocabulary = React.createClass({
     mixins: [React.addons.LinkedStateMixin],
     getInitialState: function() {
-      return {
-        name: '',
-        description: '',
-        vocabularyType: 'm',
-        learningResourceTypes: [],
-        multiTerms: false,
-      };
+      return this.makeState(this.props.vocabularyInEdit);
     },
     updateLearningResourceType: function(e) {
       var checkedType = e.target.value;
-
       var newTypes;
       if (e.target.checked) {
         if (!_.includes(this.state.learningResourceTypes, checkedType)) {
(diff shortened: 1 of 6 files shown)


giocalitri:
Can you please rebase an fix conflicts?

rest/permissions.py
```diff
@@ -105,7 +106,7 @@ def has_permission(self, request, view):
         except PermissionDenied:
             return False
 
-        if request.method in SAFE_METHODS:
+        if request.method in self.SAFE_METHODS_CUSTOM:
```
giocalitri: Why do you consider `PUT` as a safe method? Should not be only who has `edit taxonomy` permission to be able to perform a `PUT` request?
cc @pdpinch

pdpinch: Giovanni's right. Only those with permission to edit a taxonomy should be able to PUT (i.e. repository admins and curators).

amir-qayyum-khan: @giocalitri do i need to create `edit taxonomy permission` or it is already present? Can you guide me on this?
Currently i am using this view https://github.com/mitodl/lore/blob/master/rest/views.py#L153 for editing vocabulary. If I create a new `edit taxonomy permission` should i also create new API (rest view) that will only serve for delete and update purpose.

As you can see this API https://github.com/mitodl/lore/blob/master/rest/views.py#L153 extends `RetrieveUpdateDestroyAPIView` which allows delete, update. So should I only need to update permission of this API (see current permission https://github.com/mitodl/lore/blob/master/rest/views.py#L160)

giocalitri: I am not sure what you mean: the class `ManageTaxonomyPermission` that you modified here implements exactly the check you need (for the rest APIs).

The permission itself already exists and in this case it is checked later in the `has_permission` method of this class: see line [119](https://github.com/mitodl/lore/blob/feature/aq/edit_vocabulary/rest/permissions.py#L119) of your file

giocalitri: What you need (if you did not already implemented it) is a way for the interface to allow edit according to the logged user:
- if the user has permissions to edit taxonomy, the interface should allow the edit (and the REST APIs to edit taxonomy should work for him, but they already do it)
- if the user does not have permissions to edit taxonomy the interface should not allow even to show the possibility to edit

giocalitri: Actually, I take back the last comment, since the panel should not be available if the user does not have permissions to edit the taxonomy (and if it does not work in this way, we have a bug)

noisecapella:
The REST API should already allow editing a vocabulary and check permissions correctly. Does it not?

giocalitri:
@noisecapella It does as far as I know.

taxonomy/models.py
```diff
@@ -50,8 +50,7 @@ class Meta:
     @transaction.atomic
     def save(self, *args, **kwargs):
         """Handle slugs."""
-        if self.id is None or self.name != get_object_or_404(
-                Vocabulary, id=self.id).name:
+        if self.id is None:
```
noisecapella: Why was this change necessary?

ShawnMilo: From the code, I'd say that if the vocabulary name is changed, it is desired that the slug be changed.

amir-qayyum-khan: it i allow this code `self.name != get_object_or_404(Vocabulary, id=self.id).name` each time user updates a vocabulary it updates it slug and create new entry in db.

noisecapella: It doesn't necessarily create a new entry in the database, see [the Django docs](https://docs.djangoproject.com/en/1.8/ref/models/instances/#how-django-knows-to-update-vs-insert) for information on whether this is done. I'm more concerned why we need to change this behavior at all. We change the slug here on purpose to match the change in the name, and we should keep this behavior. You should update references to the term with the returned `slug` value from the API call.

amir-qayyum-khan:
Done with fix @noisecapella and @giocalitri

jamiefolsom:
In general, the tab labels are a little confusing, in that adding/editing terms is done directly in the tab currently labeled "Taxonomies", and adding/editing vocabularies is currently done in the tab labelled "Add Vocabulary". 

Arguably, the tab labels should be clarified, to read: "Manage Terms", and "Manage Vocbularies", or similar.

screenshot 2015-09-01 14 44 07

screenshot 2015-09-01 14 35 46

jamiefolsom:
- [x] One can exit the panel, after making a change to a vocabulary, and not get a warning, although upon re-opening it, one's edits are still present. Neither behavior is intuitive, IMHO.
- [x] Setting the name of the vocabulary to an empty string and submitting makes a bad PUT XHR request to the API. Arguably, this should be handled client side, and not make a request.
- [x] Setting the name of the vocabulary to an empty string and submitting displays an error, which is good, but the message should be more specific, like: "the name of the vocabulary can't be blank."

jamiefolsom:
- [x] There are no js errors in the console.
- [x] All JS tests pass

amir-qayyum-khan:
Done @jamiefolsom

amir-qayyum-khan:
Looking into coverage drop

noisecapella:
@pdpinch @Ferdi Can either of you take a look at this and verify look and feel?

pdpinch:
@amir-qayyum-khan a couple more UX change to this:
- [x] The tab label "Manage Vocabulary" should change to reflect the current state. The default should be "Add Vocabulary." When the user is editing a vocabulary it should be "Edit Vocabulary"
- [x] The first tab label "Taxonomies" should be "Vocabularies"
- [x] I opened an issue for changing the "Manage Taxonomies" button and panel title to "Manage Taxonomy" #632

amir-qayyum-khan:
Done @pdpinch

noisecapella:
If you edit a vocabulary then click cancel it will cause a page refresh. Maybe you're missing an `event.preventDefault` somewhere?

amir-qayyum-khan:
Fixed @noisecapella

noisecapella:
When adding a vocabulary, if the vocabulary name is empty and you click 'Save', the page will refresh.

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -535,6 +680,23 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
         </div>
       );
     },
+    closeTaxonomyPanel: function() {
+      this.props.closeTaxonomyPanel();
+    },
+    componentWillReceiveProps: function(nextProps) {
```
noisecapella: What is the purpose of putting `removePanelVisibility` in the state? Why not just use `this.props.removePanelVisibility` in the `render` method?

amir-qayyum-khan:
Fixed @noisecapella

pdpinch:
The coverage drop should go away when #663 is merged.

pdpinch:
@amir-qayyum-khan we need a rebase

@noisecapella did you say you wanted to look at this again?

This looks good to me, unless George has more to say.

noisecapella:
Yes, I'm going to submit a PR against this with my preferred changes

noisecapella:
@pdpinch @Ferdi If a user edits a vocabulary and deselects some learning resource types, should that remove all vocabulary<->learning resource associations for that type at that time?

pdpinch:
I think we should, because there's no way for the user to remove them later.

> @pdpinch https://github.com/pdpinch @Ferdi https://github.com/Ferdi If a user edits a vocabulary and deselects some learning resource types, should that remove all vocabulary<->learning resource associations for that type at that time?

Ferdi:
- When user deselect a type,  a) that vocab wont be available for LR of that type b) any tag assigned to existing LR should be removed 
- When user select an additional type,  the vocab shows up on LR of that type

ui/templates/includes/taxonomy_panel.html
```diff
@@ -1,24 +1,11 @@
 <div class="cd-panel-2 from-right">
   <header class="cd-panel-header">
-    <h1>Manage Taxonomies</h1>
+    <h1>Manage Taxonomy</h1>
```
jamiefolsom, line 3: Probably not in scope for this PR, but could we use "Vocabulary" instead of "Taxonomy"? Since we're editing at most one at a time, that would seem logical.

ui/templates/repo_base.html
```diff
@@ -8,7 +8,7 @@
  	{% endif %}
   	{% if 'manage_taxonomy' in perms_on_cur_repo %}
       <li>
-        <a href="#0" class="btn-taxonomies"><i class="fa fa-cogs"></i> Manage Taxonomies</a>
+        <a href="#0" class="btn-taxonomies"><i class="fa fa-cogs"></i> Manage Taxonomy</a>
```
jamiefolsom, line 11: Same question here.

noisecapella, line 11: This was requested by @pdpinch in #632

noisecapella:
I also realized I never added tests for the confirmation dialog, I'll do that

ui/templates/repo_base.html
jamiefolsom, line 11: :+1:

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -323,15 +322,34 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
...
-      e.target.isChecked = true;
     },
     updateMultiTerms: function(e) {
       this.setState({multiTerms: e.target.checked});
     },
     submitForm: function(e) {
-      var API_ROOT_VOCAB_URL = '/api/v1/repositories/' + this.props.repoSlug +
-        '/vocabularies/';
       e.preventDefault();
+      if (_.isEmpty(this.state.name)) {
```
jamiefolsom, line 331: If the name is an integer, this would return true. Would that ever happen?

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -323,15 +322,34 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
...
+        this.props.reportMessage({error: 'Please enter vocabulary name.'});
+        return;
+      }
+
+      if (_.isEmpty(this.state.description)) {
```
jamiefolsom, line 336: Same here.

jamiefolsom:
When I run tox, I get some js errors: 

```
11:49 $ docker-compose run web tox -e js
js installed: wheel==0.24.0
js runtests: PYTHONHASHSEED='1489960556'
js runtests: commands[0] | npm install --no-bin-links --prefix /src
npm WARN package.json lore@0.0.0 No repository field.
js runtests: commands[1] | /src/node_modules/jshint/bin/jshint ui/static/ui/js ui/jstests/
js runtests: commands[2] | /src/node_modules/jsxhint/cli.js ui/static/ui/js/**/*.jsx ui/jstests/**/*.jsx --jsx-only
ui/jstests/test_manage_taxonomies.jsx: line 209, col 11, It's not necessary to initialize 'errorMessage' to 'undefined'.
ui/jstests/test_manage_taxonomies.jsx: line 750, col 11, It's not necessary to initialize 'errorMessage' to 'undefined'.

2 errors
ERROR: InvocationError: '/src/node_modules/jsxhint/cli.js ui/static/ui/js/**/*.jsx ui/jstests/**/*.jsx --jsx-only'
_____________________________________________________________________________________________________________________________ summary _____________________________________________________________________________________________________________________________
ERROR:   js: commands failed
```

jamiefolsom:
- [x] tests pass
- [x] feature works
- [x] code clear
- [x] travis checks pass
- [x] questions answered

ui/static/ui/js/manage_taxonomies.jsx
noisecapella, line 331: This is from @amir-qayyum-khan's code, I'll change it to be a `!== ''` instead

jamiefolsom, line 331: Actually, it appears that numbers entered by a user are handled as strings, so this is a non-issue.

jamiefolsom:
I am intermittently getting "There was a problem updating the Vocabulary" when I click the "Update" button.
screenshot 2015-09-24 12 25 58

jamiefolsom:
So, as you can see above -- I changed the name of the vocabulary to "1", which appears to have worked, then when I try to edit that name, I get the error mentioned above. It appears that these vocabularies are looked up by the user-entered name, rather than by db id; is that correct? 

Wouldn't that lead to all sorts of issues, if people start editing the names of their vocabularies? Here is some log output from requests to add terms, after I changed the name from "colors" to "1". Note the 400 error on the POST to "colors" -- that's when I try to add "Blue", which is an existing term on the vocabulary which I renamed from "colors" to "1". "Yellow", "1234", "asdf" all worked fine.

```
web_1     | [24/Sep/2015 16:27:41] "PATCH /api/v1/repositories/test/vocabularies/1/ HTTP/1.1" 404 23
web_1     | [24/Sep/2015 16:28:00] "PATCH /api/v1/repositories/test/vocabularies/1/ HTTP/1.1" 404 23
web_1     | [24/Sep/2015 16:28:48] "DELETE /api/v1/repositories/test/vocabularies/1/ HTTP/1.1" 404 23
web_1     | [24/Sep/2015 16:29:23] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 400 77
web_1     | [24/Sep/2015 16:29:26] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 400 77
web_1     | [24/Sep/2015 16:29:41] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 201 53
web_1     | [24/Sep/2015 16:29:41] "GET /api/v1/repositories/test/search/ HTTP/1.1" 200 8497
web_1     | [24/Sep/2015 16:29:45] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 201 55
web_1     | [24/Sep/2015 16:29:46] "GET /api/v1/repositories/test/search/ HTTP/1.1" 200 8497
web_1     | [24/Sep/2015 16:29:48] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 201 43
web_1     | [24/Sep/2015 16:29:48] "GET /api/v1/repositories/test/search/ HTTP/1.1" 200 8497
web_1     | [24/Sep/2015 16:29:53] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 201 49
web_1     | [24/Sep/2015 16:29:53] "GET /api/v1/repositories/test/search/ HTTP/1.1" 200 8497
web_1     | [24/Sep/2015 16:29:54] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 201 49
web_1     | [24/Sep/2015 16:29:54] "GET /api/v1/repositories/test/search/ HTTP/1.1" 200 8497
web_1     | [24/Sep/2015 16:29:57] "POST /api/v1/repositories/test/vocabularies/colors/terms/ HTTP/1.1" 400 77
```

jamiefolsom:
Shouldn't the [ x ] button cancel the edit on this UI? It doesn't do anything, that I can tell, and in fact there's no obvious way to cancel the editing of a term. Maybe out of scope, since that's term editing, not vocab editing.

screenshot 2015-09-24 12 37 50

ui/static/ui/js/manage_taxonomies.jsx
jamiefolsom, line 331: Now it seems that using an integer for a vocab title is an issue, but for a different reason. Is it the case that when the API gets a request for a vocabulary with an integer for a slug, it attempts to look it up by id?

noisecapella:
The urls will change when you update the vocabularies and terms, that's the problem with using slugs. The alternative is to have slugs that don't change with the name which may also lead to confusion, might be more correct in certain cases, but in any case our code now needs to handle keeping itself consistent. I'll look into this issue

noisecapella:
Regarding your comment on the edit terms UI, that was merged in PR #574 so it's out of scope here but feel free to file an issue

noisecapella:
Oh, the button doesn't work for you at all? That's something I should look into

jamiefolsom:
@noisecapella Could the display title be editable, but the slug derived from the first title entered be uneditable? That's more or less how it's done for rails/active_record, using the friendly_id gem: https://github.com/norman/friendly_id

ui/static/ui/js/manage_taxonomies.jsx
noisecapella, line 331: It should never do that, any number in a slug field or in a url should be treated as just another slug. If you changed the name to a number it would look as if the vocabulary was looked up by id.

jamiefolsom:
@noisecapella appears that the "edit term [x] button not doing anything" issue is only when the parent vocabulary title has been edited, because it's in that case that neither the vocabulary nor (as a consequence) the term can be located.

noisecapella:
I think the slug consistency issues should be fixed now. Are there any remaining functionality issues? I'm working on getting the tests back in line.

jamiefolsom:
None that I can see. Looks good. :+1:

noisecapella:
Ok I'll merge once tests pass

PR #524, opened 2015-08-17
Added base sorting field in case used sorting is working on same values

### LINKED ISSUE #511: Sort doesn't appear to be stable
As a curator, I need to be able to save metadata to resources and have the search results stay stable. Currently, it appears that the index is updated every time metadata is saved, which upsets the order of the search results. 

It's not clear if this is due to a bug in the sort widget, or if it is due to sorting on fields that are unpopulated (i.e. number of views).

giocalitri:
Closes #511

### DIFF
--- a/search/tests/test_indexing.py
+++ b/search/tests/test_indexing.py
@@ -93,39 +93,63 @@ class TestIndexing(SearchTestCase):
             xa_nr_attempts=101,
             xa_avg_grade=7.3
         ))
-        self.assertEqual(self.count_results(''), 3)
+        res4 = self.create_resource(**dict(
+            resource_type="example",
+            title="silly example 4",
+            content_xml="<blah>blah 4</blah>",
+            mpath="/blah3",
+            xa_nr_views=1003,
+            xa_nr_attempts=101,
+            xa_avg_grade=9.9
+        ))
+        self.assertEqual(self.count_results(''), 4)
         # sorting by number of views
         results = self.search(
             '',
             sorting=LoreSortingFields.SORT_BY_NR_VIEWS[0]
         )
-        # expected position (res2, res3, res1)
+        # expected position res2, res4
         top_res = results[0]
+        sec_res = results[1]
         self.assertEqual(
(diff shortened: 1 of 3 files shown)


search/tests/test_indexing.py
```diff
@@ -101,10 +110,15 @@ def test_sorting(self):
         )
         # expected position (res2, res3, res1)
         top_res = results[0]
+        sec_res = results[1]
         self.assertEqual(
             top_res.nr_views,
             res2.xa_nr_views
         )
+        self.assertEqual(
```
ShawnMilo, line 118: I think these tests don't test that the results appear in the proper order, because it's comparing the column that is identical between the two. The foreign key would be better (at least asserting that it's lower in `top_res` than in `sec_res`.

giocalitri, line 118: I don't think I understood

giocalitri, line 118: I think the test of the proper order is implicit:
1- res 4 is created after the others and has all the top values of the sorting
2- since res4 has the top value of each sorting, but a higher ID, it appears always as second in the sorting

ShawnMilo, line 118: If you change this:

``` python
        self.assertEqual(
            top_res.nr_views,
            res2.xa_nr_views
        )
        self.assertEqual(
            sec_res.nr_views,
            res4.xa_nr_views
        )
```

to this:

``` python
        self.assertEqual(
            top_res.nr_views,
            res4.xa_nr_views
        )
        self.assertEqual(
            sec_res.nr_views,
            res2.xa_nr_views
        )
```

You will get the same results (the tests will still pass), but it won't be correct.

ShawnMilo, line 118: The test of the sorting _would_ be implicit if you were comparing the IDs, but by only comparing `nr_views` they will always be the same, so you can't test the ordering.

giocalitri, line 118: I see what you mean: I guess I need to check that the ID of the resource is the same, then

PR #525, opened 2015-08-17
Added lazy loading of static asset information

noisecapella:


carsongee:
Just happened to come by, looks like the package is missing from bower

carsongee:
and now I know why from #526

pdpinch:
This is working well for me.

ui/static/ui/js/utils.jsx
```diff
@@ -1,12 +1,75 @@
...
+      });
+    },
+    handleInfiniteLoad: function () {
+      this.setState({
+        isInfiniteLoading: true
+      });
+      this.loadPage();
+    },
+    render: function () {
+      var spinner = <div className="infinite-item-list">Loading...</div>;
```
giocalitri, line 50: This can be leaved as it is and maybe it's out of scope, but isn't a real spinner better?

noisecapella, line 50: This was pulled from their example code, I didn't do much thinking on how this would look. I have no argument switching to a spinner it may be better to do it in a separate PR

ui/static/ui/js/static_assets.jsx
```diff
@@ -1,8 +1,15 @@
...
+  ], function (React, $, _, Utils) {
   'use strict';
 
   var StatusBox = Utils.StatusBox;
+  var InfiniteList = Utils.InfiniteList;
+
+  // Guess at a good value for the height of each element. Infinite requires
+  // an elementHeight for each item. If we wanted to do a per item elementHeight
+  // we could make this a list instead.
+  var elementHeight = 50;
```
giocalitri, line 12: how did you determine this number is appropriate? trying to understand the logic

ui/static/ui/js/static_assets.jsx
```diff
@@ -10,49 +17,49 @@ define('static_assets', [
...
-            <a href={asset.asset}>{asset.name}</a>
-          </li>;
-      });
+      var url = "/api/v1/repositories/" + this.props.repoSlug +
+          "/learning_resources/" +
+          this.props.learningResourceId + "/static_assets/";
+
+      // Infinite requires you to specify a containerHeight. This value
+      // is a guess at how big the area should be relative to the height.
+      var containerHeight = $(window).height() - 200;
```
giocalitri, line 33: same as before for the number (the 200)

ui/static/ui/js/static_assets.jsx
noisecapella, line 12: It's just a guess, I picked a value that seemed to look good in the UI

ui/static/ui/js/static_assets.jsx
noisecapella, line 33: I was trying to pick a height which is the window height minus 200 pixels for the panel borders. It's also just a guess

ui/static/ui/js/static_assets.jsx
giocalitri, line 12: I'm fine with this, but I'm wondering if we need to test on all the browsers if it has the same behavior (same for the other number below)

giocalitri, line 12: cc @pdpinch

giocalitri:
The code looks fine to me other than two small comments: I'm going to test functionality

giocalitri:
If you make any change to the code, please rebase master.

ui/static/ui/js/static_assets.jsx
pdpinch, line 12: I'll do some quick cross-browser testing on my Mac. 

I also saw it do some odd, but ultimately OK things, when my viewport was short (I had the developer panel open to look at the network traffic). 

> I'm wondering if we need to test on all the browsers if it has the same behavior (same for the other number below)

pdpinch, line 12: Everything I've tested behaves the same. 

I will probably run this UI by the designer. I think there's probably some polish we can add, but it totally meets the needs of the story.

giocalitri:
Functionality is fine: feel free to merge

giocalitri:
(sorry it took so long, I wanted to re-run the tests locally)

PR #526, opened 2015-08-17
Installed react-infinite via bower

noisecapella:
Required for PR #525


giocalitri:
Since this is just a new package via Bower, I think it can be merged

PR #527, opened 2015-08-17
Added lodash to bower.json

noisecapella:
Looks like I forgot to save lodash in `bower.json` after committing the bower directory


giocalitri:
This can be merged

PR #532, opened 2015-08-17
Fixed jsxhint to focus on our files

noisecapella:
This makes the globbing more specific to avoid linting library files


jamiefolsom:
Makes sense to me. Will check it out.

jamiefolsom:
Works fine. Were you seeing linting issues on libraries?

noisecapella:
Every library has its own code style, even things like semicolons may be different, so we can't just lint library code with our own style.

PR #538, opened 2015-08-18
Added deploy button and app.json

carsongee:
This is a minimal set of addons, not the production grade addons.


pdpinch:
This is good to go now, right?

carsongee:
Yep

Issue #512, opened 2015-08-13
New facet: Curation status


pdpinch:
Physics team would want a "default" facet  (like "Item type") that will have the following possible state/values :
- Ready-To-Use: Content is ready to drop in a course without editing, and has a complete set of metadata associated with it (topic, learning objectives, keywords).  
- Tagged: Minimally curated content. Incomplete metadata, or needs editing to be used in a course.  
- Course Information: Content that is tagged as course (and often course run) specific information, with no library purpose other than rerunning a course.
- Discarded: Material that is completely irrelevant and not intended for further use. By default, anything with this status should not be exported in a rerun export. E.g., temporary working chapters, discussion blocks. Can also include irrecoverably bad content.
- Hidden: Objects required for internal structure, but not intended to be searchable. E.g., verticals with one entry.
- (No tag): Default status. Essentially untouched material, uncurated , directly imported from a course.

Some of these values must be manually assigned by a user in the learning resource edit form (e.g.  "Ready-To-Use") but others can change state automatically.

pdpinch:
@Ferdi what should we do for UI for this? I think you had suggested using a special (text) search term.

Ferdi:
To add some more context , Physics team would want a "default" facet  (like "Item type") that will have the following possible state/values :

* Ready-To-Use: Content is ready to drop in a course without editing, and has a complete set of metadata associated with it (topic, learning objectives, keywords).

* Tagged: Minimally curated content. Incomplete metadata, or needs editing to be used in a course.

* Course Information: Content that is tagged as course (and often course run) specific information, with no library purpose other than rerunning a course.

* Discarded: Material that is completely irrelevant and not intended for further use. By default, anything with this status should not be exported in a rerun export. E.g., temporary working chapters, discussion blocks. Can also include irrecoverably bad content.

* Hidden: Objects required for internal structure, but not intended to be searchable. E.g., verticals with one entry.

* (No tag): Default status. Essentially untouched material, uncurated , directly imported from a course.

pdpinch:
Break this into separate stories:
- [ ] #534 automatically add this vocabulary to every repo, with all terms created as normal. Regular vocabulary, added to all courses. Still editable. Need to be able to update our existing repos. But we won't update them going forward. 
  _3_
- [ ] #535 create the "-- not set --" search filter so it's possible to facet on items without a value. Scoped within a particular vocabulary. Would be nice to have a "no term" filter for each vocabularies. 
  _2_
[...]

PR #540, opened 2015-08-18
WIP Feature/skm/512 search empty facet

ShawnMilo:
Done: Add a "not set" default to every vocabulary when it is created. This can be used for seeing counts of and filtering down to `LearningResource`s that haven't yet been assigned metadata.

WIP. Next steps:
- ability to filter on the "not set" value
- automatically remove "not set" value when any other term is added (good catch, @pdpinch)

This is a part of Epic #512.

### DIFF
--- a/learningresources/migrations/0015_backfill_curator_vocabularies.py
+++ b/learningresources/migrations/0015_backfill_curator_vocabularies.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+from django.utils.text import slugify
+
+from taxonomy.signals import create_default_vocabulary
+
+# pylint: skip-file
+
+VOCAB_NAME = "curation status"
+EMPTY_VALUE = "--not set--"
+
+
+def backfill_curator_vocabs(apps, schema_editor):
+    """Backfill auto-generated vocabulary to existing repos."""
+    Repository = apps.get_model("learningresources", "Repository")
+    LearningResourceType = apps.get_model(
+        "learningresources", "LearningResourceType")
+    Vocabulary = apps.get_model("taxonomy", "Vocabulary")
+    learning_resource_types = LearningResourceType.objects.all()
+    for repo in Repository.objects.all().iterator():
+        if Vocabulary.objects.filter(
+            name=VOCAB_NAME, repository__id=repo.id).exists():
+            continue
+        vocab = Vocabulary.objects.create(
+            repository_id=repo.id,
+            name=VOCAB_NAME, description=VOCAB_NAME,
+            required=True, vocabulary_type='m',
+            weight=0, multi_terms=False,
+            slug=slugify("{0}_{1}_backfill".format(VOCAB_NAME, repo.id)),
+        )
+        for resource_type in learning_resource_types:
+            vocab.learning_resource_types.add(resource_type)
+        create_terms(apps, vocab.id)
+        backfill_resources(apps, repo.id)
+
+
+def create_terms(apps, vocab_id):
+    """Create the curation status terms for a newly-created Vocabulary."""
+    Term = apps.get_model("taxonomy", "Term")
+    labels = (
+        'ready to use', 'tagged', 'course information',
+        'discarded', 'hidden', EMPTY_VALUE,
+    )
+    for label in labels:
+        Term.objects.create(
+            vocabulary_id=vocab_id,
+            label=label,
+            slug=slugify("{0}_{1}_backfill".format(label, vocab_id)),
+            weight=0,
+        )
+
+
+def backfill_resources(apps, repo_id):
+    """Add terms to existing LearningResource instances."""
+    LearningResource = apps.get_model("learningresources", "LearningResource")
+    Term = apps.get_model("taxonomy", "Term")
+    empty_term = Term.objects.get(
+        vocabulary__repository__id=repo_id,
+        vocabulary__name=VOCAB_NAME,
+        label=EMPTY_VALUE,
+    )
+    resources = LearningResource.objects.filter(
+        course__repository__id=repo_id
+    )
+    for resource in resources.iterator():
+        resource.terms.add(empty_term)
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('learningresources', '0014_learning_resource_related_name'),
+        ('taxonomy', '0007_vocabulary_multi_terms'),
+    ]
+
+    operations = [
+        migrations.RunPython(backfill_curator_vocabs)
+    ]
--- a/learningresources/models.py
+++ b/learningresources/models.py
@@ -10,6 +10,7 @@ import six.moves.urllib.parse as urllib_parse  # pylint: disable=import-error
 
 from django.db import models
 from django.db import transaction
+from django.dispatch import Signal
 from django.contrib.auth.models import User
 from django.utils.text import slugify
 from django.utils.encoding import python_2_unicode_compatible
@@ -26,6 +27,12 @@ log = logging.getLogger(__name__)
 FILE_PATH_MAX_LENGTH = 900
 STATIC_ASSET_PREFIX = 'assets'
 STATIC_ASSET_BASEPATH = STATIC_ASSET_PREFIX + '/{org}/{course_number}/{run}/'
+
+# Used to signify that a repository has been created
+# and its initial permissions have been set.
+# This allows the "curation status" vocabulary to be
+# created automatically.
+repo_created = Signal(providing_args=["repository"])
 
 
 class FilePathLengthException(Exception):
@@ -136,45 +143,6 @@ class LearningResource(BaseModel):
     xa_histogram_grade = models.FloatField(default=0)
     url_name = models.TextField(null=True)
 
-    def get_preview_url(self, org=None, course_number=None, run=None):
-        """
-        Create a preview URL. Accepts optional kwargs to prevent
-        database lookups, especially for during search engine indexing.
-        Args:
-            org (unicode): self.course.org
-            run (unicode): self.course.run
-            course_number (unicode): self.course.course_number
-        """
-        if org is None:
-            org = self.course.org
-        if course_number is None:
-            course_number = self.course.course_number
-        if run is None:
-            run = self.course.run
-        key = "{org}/{course}/{run}".format(
-            org=org,
-            course=course_number,
-            run=run,
-        )
-
-        if self.url_name is not None:
-            url_format = 'courses/{key}/jump_to_id/{preview_id}'
-            return LORE_PREVIEW_BASE_URL + urllib_parse.quote(
-                url_format.format(
-                    base_url=LORE_PREVIEW_BASE_URL,
-                    key=key,
-                    preview_id=self.url_name,
-                )
-            )
-        else:
-            url_format = 'courses/{key}/courseware'
-            return LORE_PREVIEW_BASE_URL + urllib_parse.quote(
-                url_format.format(
-                    base_url=LORE_PREVIEW_BASE_URL,
-                    key=key,
-                )
-            )
-
 
 @python_2_unicode_compatible
 class LearningResourceType(BaseModel):
@@ -200,7 +168,13 @@ class Repository(BaseModel):
 
     @transaction.atomic
     def save(self, *args, **kwargs):
-        """Handle slugs and groups"""
+        """
+        Handle slugs and groups.
+
+        Note that creation of a new Repository also triggers a signal in
+        the taxonomy application to create the curation status vocabulary.
+        That code can't be put here because it would create a circular import.
+        """
         is_update = False
         if self.id is None or self.name != get_object_or_404(
                 Repository, id=self.id).name:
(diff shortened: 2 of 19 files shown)


learningresources/models.py
```diff
@@ -175,6 +175,12 @@ def get_preview_url(self, org=None, course_number=None, run=None):
                 )
             )
 
+    def save(self, *args, **kwargs):
+        """Set default value for taxonomies."""
+
```
ShawnMilo: This is incomplete. WIP, after all.

giocalitri:
Is this PR ready to be reviewed?

ShawnMilo:
It's not complete, but I put it up with the "Needs Review" tag so any questioning of the approach can be caught early and so the final code review will not be one huge dump of new code.

This feature contains at least four or five sub-features, so I think reviewing it incrementally would be easier for the reviewer(s).

learningresources/models.py
```diff
@@ -228,3 +202,46 @@ class Meta:  # pylint: disable=missing-docstring
...
+        run = resource.course.run
+    key = "{org}/{course}/{run}".format(
+        org=org,
+        course=course_number,
+        run=run,
+    )
+
+    if resource.url_name is not None:
+        url_format = 'courses/{key}/jump_to_id/{preview_id}'
+        return LORE_PREVIEW_BASE_URL + urllib_parse.quote(
```
giocalitri: not a fan of string concatenation, but feel free to leave it as it is

learningresources/models.py
```diff
@@ -228,3 +202,46 @@ class Meta:  # pylint: disable=missing-docstring
...
+            url_format.format(
+                base_url=LORE_PREVIEW_BASE_URL,
+                key=key,
+                preview_id=resource.url_name,
+            )
+        )
+    else:
+        url_format = 'courses/{key}/courseware'
+        return LORE_PREVIEW_BASE_URL + urllib_parse.quote(
```
giocalitri: like before

learningresources/models.py
```diff
@@ -228,3 +202,46 @@ class Meta:  # pylint: disable=missing-docstring
             RepoPermission.add_edit_metadata,
             RepoPermission.manage_repo_users,
         )
+
+
+def get_preview_url(resource, org=None, course_number=None, run=None):
```
giocalitri, line 207: if this has become an API, why is it here in `models.py` and not in `api.py`? Circular imports?

ShawnMilo, line 207: I had to take it out of the model because it caused problems during migrations -- the `LearningResource` in a migration, from the migrations ORM, doesn't have the `get_preview_url` method.

However, I didn't see any reason to move it to another file.

giocalitri, line 207: It would be mostly to keep the `models.py` cleaner, but feel free to leave it here

learningresources/models.py
ShawnMilo: I'll fix the string concatenation, but for the record it was not introduced in this commit; just cut & pasted from the model.

giocalitri: like, I said, feel free to leave it as it is: I'm just going through the "green lines"

taxonomy/api.py
```diff
@@ -52,3 +52,32 @@ def get_term(repo_slug, user_id, vocab_slug, term_slug):
         raise NotFound()
     except Term.DoesNotExist:
         raise NotFound()
+
+
+# pylint: disable=too-many-arguments
+def create_vocabulary(
+        user_id, repo_slug, name, description,
+        required=False, vocabulary_type=Vocabulary.MANAGED, weight=0,
+        multi_terms=False):
```
giocalitri, line 61: I know this passed pylint checks, but it looks pretty weird

taxonomy/models.py
```diff
@@ -50,19 +50,30 @@ class Meta:
...
+        if create_notset:
+            Term.objects.create(
+                vocabulary=self, label=Term.EMPTY_VALUE, weight=0)
+        return self
 
 
 class Term(BaseModel):
     """Model for term table"""
+
+    EMPTY_VALUE = '--not set--'
```
giocalitri, line 75: this looks pretty ugly in the interface, but not sure what to use
@pwilkins do you have any suggestion?

giocalitri:
While I'm reviewing the code, I run the migration on my current environment.
There is clearly a bug because this is how the "Item type" looks before the migration:

item type working

and this is how it looks after:
item type

(I reproduced this twice)

ShawnMilo:
From your screenshot, it seems that it's due to the migration having a "fake" `LearningResource` instead of a real instance, and it's getting some other value for the facet. I'll investigate and fix.

taxonomy/signals/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        vocabulary__name=CURATION_VOCAB_NAME,
+        label=Term.EMPTY_VALUE,
+    )
+    if terms.count() != 1:
+        log.error(
+            (
+                "Unable to add default curation status value to "
+                "learning resource pk %s. Expected 1 matching "
+                "term, found %s."
+            ), instance.id, terms.count()
```
giocalitri, line 73: why did you use this type of string formatting?

ShawnMilo, line 73: The string was over 80 characters, so it had to be broken up. How would you have done it?

giocalitri:
Something I noticed in this PR in two places is that you created two packages with only a `__init__.py` file inside with all the code there.
I dislike this practice because I think the `__init__.py` should be use ONLY for package setup purposes and I prefer one of these two approaches:
1- you simply create a module for that package
2- you move the code outside the `__init__.py` and
  a- import in the `__init__.py` only what you need
  b- you leave the `__init__.py` empty and you import using the right module

learningresources/migrations/0015_backfill_curator_vocabularies.py
```diff
@@ -0,0 +1,79 @@
...
+
+from django.db import models, migrations
+from django.utils.text import slugify
+
+from taxonomy.signals import create_default_vocabulary
+
+# pylint: skip-file
+
+VOCAB_NAME = "curation status"
+EMPTY_VALUE = "--not set--"
```
giocalitri, line 12: why are you defining these strings also here?

taxonomy/signals/__init__.py
giocalitri, line 73: I was talking about the `%s`, not the implicit string concatenation

learningresources/migrations/0015_backfill_curator_vocabularies.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def backfill_curator_vocabs(apps, schema_editor):
+    """Backfill auto-generated vocabulary to existing repos."""
+    Repository = apps.get_model("learningresources", "Repository")
+    LearningResourceType = apps.get_model(
+        "learningresources", "LearningResourceType")
+    Vocabulary = apps.get_model("taxonomy", "Vocabulary")
+    learning_resource_types = LearningResourceType.objects.all()
+    for repo in Repository.objects.all().iterator():
```
giocalitri, line 22: I did not know about the `.iterator()`: this is pretty useful!

taxonomy/signals/__init__.py
ShawnMilo, line 73: That's because it's a logging statement, not string interpolation. Note that there's a `,` before the arguments, not `%`.

giocalitri, line 73: sorry, missed it

learningresources/migrations/0015_backfill_curator_vocabularies.py
```diff
@@ -0,0 +1,79 @@
...
+        backfill_resources(apps, repo.id)
+
+
+def create_terms(apps, vocab_id):
+    """Create the curation status terms for a newly-created Vocabulary."""
+    Term = apps.get_model("taxonomy", "Term")
+    labels = (
+        'ready to use', 'tagged', 'course information',
+        'discarded', 'hidden', EMPTY_VALUE,
+    )
```
giocalitri, line 45: I don't see this anywhere else: does it mean that this vocabulary (curation status) will be crated only for old repo?

learningresources/models.py
```diff
@@ -27,6 +28,12 @@
 STATIC_ASSET_PREFIX = 'assets'
 STATIC_ASSET_BASEPATH = STATIC_ASSET_PREFIX + '/{org}/{course_number}/{run}/'
 
+# Used to signify that a repository has been created
+# and its initial permissions have been set.
+# This allows the "curation status" vocabulary to be
+# created automatically.
+repo_created = Signal(providing_args=["repository"])
```
giocalitri, line 35: why this has been defined here and not in the signals? (asking because I actually don't know about best practices in this case)

ShawnMilo, line 35: I created it here because it is called from this file (on the model `save()`). If there's a "correct" way to do it I don't know it. It just made sense to put it here. If I put it in signals then I'd have to import it here (as much code as creating it), so it feels okay to put it here.

learningresources/migrations/0015_backfill_curator_vocabularies.py
ShawnMilo, line 45: Yes, you're right. The `create_default_vocabulary` function in taxonomy signals should also create the terms. Good catch, thank you.

learningresources/models.py
giocalitri, line 35: is it implicitly called? asking because I don't see the call in `save()`

ShawnMilo:
@giocalitri I pushed another commit to fix the issue with the "Item Type."

learningresources/models.py
ShawnMilo, line 35: It is explicitly called. I'm just a bobo and forgot I had to move the call to taxonomy signals.

Because:

Originally I added a `post_save` signal for this (before creating the custom `repo_created` signal). However, the taxonomy app _already_ listened for a `post_save` to fix #549. It would be fine to have both as `post_save`, except that they fire in the wrong order. There is no way to set the order. It's actually great that it failed locally or it could have failed randomly in production.

So, I have to call `repo_created` from the permissions signal handler because I have to know the permissions are set before going further (this is actually the original problem that launched #549).

I still lean towards having the handler defined here, since it is regarding the `Repository` model.

learningresources/migrations/0015_backfill_curator_vocabularies.py
```diff
@@ -0,0 +1,79 @@
...
+            slug=slugify("{0}_{1}_backfill".format(label, vocab_id)),
+            weight=0,
+        )
+
+
+def backfill_resources(apps, repo_id):
+    """Add terms to existing LearningResource instances."""
+    LearningResource = apps.get_model("learningresources", "LearningResource")
+    Term = apps.get_model("taxonomy", "Term")
+    term = Term.objects.get(
```
giocalitri: nit: this would be more clear if named differently: something like `empty_term` or like this

learningresources/models.py
giocalitri, line 35: I'm fine with leaving it here.

giocalitri, line 35: thanks for the explanation BTW

giocalitri:
The Item Type bug is solved

giocalitri:
I don't think there is anything major other than the bug already solved. Probably my `__init__` "philosophical" point of view.

In general I really liked the amount of comments that make really clear what is going on in the code.

ShawnMilo:
I read and understand the preferences on `__init__.py`. I prefer to keep it as-is, because creating another file just to import from in `__init__.py` doesn't gain anything, and lots of projects have code in `__init__.py`. I'd rather not need the file at all, but here we are. I could add another file there, but then I'd have to come up with another level of naming that's unnecessary.

ShawnMilo:
I don't know if it carries any water, but Django has thousands of lines of code in its various `__init__.py` files.

ShawnMilo:
On Mon, Aug 24, 2015 at 3:37 PM, Giovanni Di Milia <notifications@github.com

> wrote:
> 
> I don't think there is anything major other than the bug already solved.
> Probably my **init** "philosophical" point of view.
> 
> In general I really liked the amount of comments that make really clear
> what is going on in the code.
> 
> Thanks! I tried to put good comments, and will continue to do so.

I've found it helpful to review the code myself the day after I wrote it,
and comment anything that isn't obvious.

ShawnMilo:
All comments have been addressed, unless I missed something. I just read through the comments again and I don't think so.

giocalitri:
I guess the only thing that is missing is the definition of constants (mostly strings) that happen in 2 different places. I do not think it's a good way to go.

giocalitri:
I will open issues for:
- "--not set--" string looking weird in the UI
- the "--not set--" term does not appear right away after the new vocabulary is created

ShawnMilo:
What constants are you referring to?

giocalitri:
`VOCAB_NAME` , `EMPTY_VALUE` and `labels` defined both in the migration and in other parts of the code

pdpinch:
I'll see if I can get some attention from the designer. 

Feel free to suggest another phrase/presentation. 

Shawn, I presume the dashes are to make it unlikely to collide with a term the user wants to create?

> I will open issues for:
> "--not set--"
> the "--not set--" term does not appear right away after the new vocabulary is created

ShawnMilo:
Yeah, I thought about that, but since a migration is once-and-done, there's no value in making it get dynamic, up-to-date values. In fact, it would be bad; the migration should be frozen at the state they were in at the time of creation.

Changing the migration could cause a problem because migration 50 could, for example, have a function to look for `--not set--` and replace it with `<empty>`, but it would fail because the migration 15 would have set it to `<empty>` instead. You could argue that's a good thing, but what if migration 26 did something with `--not set--` values?

ShawnMilo:
@pdpinch: Yes. I didn't think anyone would use `--not set--` for anything. Also, historically a lot of drop-down boxes have used that style `-- SELECT COUNTRY --`, for example.

giocalitri:
@ShawnMilo I understand your point, but I think that, considering that we will allow the users to modify terms, the presence of the term "-- not set --" cannot be guaranteed in the future.
If we need to guarantee it, I guess the implementation you made is not completely right.

Anyway, this is not a blocker for this particular PR, so I think that if you don't want to touch the code any more, you can squash and merge

ShawnMilo:
Okay. Well, if a user manually changes `--not set--` then that's on them. I understand there are no guarantees. I'll update the branch from master, flatten, and push for Travis.

ShawnMilo:
@giocalitri: Is this good to merge?

giocalitri:
If nothing else has changed since my previous comment "you can squash and merge", go for it!

Issue #513, opened 2015-08-13
Better navigation of paging in search results


pdpinch:
As a user, when my search results span multiple pages, I can use the footer navigation to move forward or back one page at a time. I would like to have a way to skip ahead to an arbitrary page in the results, and also a way to know how many pages there are in my results.

pdpinch:
This used to be part of the template, but we removed it. I believe the old code was in the [listing template](https://github.com/mitodl/lore/blob/dd0fdff21f4b36b2ac903a8717528cb7fea0ec38/ui/templates/listing.html#L93-L107)

amir-qayyum-khan:
@pdpinch  Is there any mock up for this task? Any design you are interested in. Or i need to give an idea of representation?

pdpinch:
No mock ups. See if you can make it fit in with the existing < and > navigation. Start with the code linked above?

amir-qayyum-khan:
ok, sure. let me try

pdpinch:
closed by #544

PR #544, opened 2015-08-19
Made better navigation of paging in search results

amir-qayyum-khan:
In this PR I added
- Current page and max page information to pagination widget
- On Clicking widget a text field appears. User can enter any page number greater the 0 and less then current page.
- On press enter button user will be navigate to the newly entered page. If page number is valid
- On focus out user will be navigate to the newly entered page. If page number is valid 
- If user do not change page nothing happen, he can exit navigation text field by focus out or pressing enter key

Issue https://github.com/mitodl/lore/issues/513
@pdpinch @giocalitri 

screen shot 2015-08-19 at 5 01 51 pm

screen shot 2015-08-19 at 5 01 55 pm

screen shot 2015-08-19 at 5 02 04 pm

screen shot 2015-08-19 at 5 02 16 pm

### DIFF
--- a/ui/static/ui/css/mit-lore.css
+++ b/ui/static/ui/css/mit-lore.css
@@ -535,3 +535,22 @@ ul.repos {
     padding: 0px;
     margin: 0px;
 }
+
+.repo-page-status {
+    white-space: nowrap;
+    cursor: pointer;
+    display: inline;
+    font-size: 25px;
+    position: relative;
+    top: 27%;
+    transform: translateY(-27%);
+}
+
+input.repo-page-status {
+    width: 7%;
+    display: inline;
+    font-size: medium;
+    position: relative;
+    top: 27%;
+    transform: translateY(-27%);
+}
--- a/ui/static/ui/js/listing.js
+++ b/ui/static/ui/js/listing.js
@@ -76,7 +76,87 @@ define('listing',
         });
     }
 
+    /**
+     * Change page number in edit mode. Navigate to new page of it is valid
+     *
+     * @param {Number} pageNum - current page number
+     * @param {Number} maxPages - total number of pages
+     * @param {JQuery} callerElem - Reference to pagination status link
+     */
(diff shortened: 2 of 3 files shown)


ShawnMilo:
Minor usability points:
- The text box is too small; two-digit numbers are cut off, and for three digit numbers, the third digit doesn't appear on the screen.

Maybe small, but worth mentioning:
- I don't think it's obvious that one is expected to click on the "page x of n" text, although that's a fairly minor training issue.
- After selecting the page number you want, you can presss `<ENTER>` or click away from the widget, but it doesn't seem obvious.

amir-qayyum-khan:
@ShawnMilo any suggestion on these points?

> - I don't think it's obvious that one is expected to click on the "page x of n" text, although that's a fairly  minor training issue.
> - After selecting the page number you want, you can presss <ENTER> or click away from the widget, but it doesn't seem obvious.

ShawnMilo:
For the first one, what if the box was always there, so it would have the left and right arrows, with "page x of n" in the center, with "x" always being a text box? It wouldn't look as pretty, but would be more obvious.

As for the second one, I don't know if it's a good idea, but maybe show "press `ENTER` to view page" or something could appear when the user has clicked into the box, and disappear when they leave the box or press Enter.

I'm not a UI or UX expert though; someone else should weigh in as well. Any recommendations, @pdpinch?

pdpinch:
I didn't want to delay this waiting for UX input, but it will eventually get UX input. 

For now, it can be an "advanced" feature that requires a little user training. 

I would like the box to be big enough to type 3 numbers though. 
- Peter

ShawnMilo:
Okay. So once the box is enlarged I think it'll be good to merge.

amir-qayyum-khan:
let me do that quickly.

amir-qayyum-khan:
Done @ShawnMilo

ShawnMilo:
It looks good. I'll merge after Travis passes. :o)

amir-qayyum-khan:
sure :)

PR #547, opened 2015-08-19
Fixed static asset download for local servers

### LINKED ISSUE #546: Static asset links don't work for local servers (when LORE_USE_S3 is disabled)
An empty response is received instead

noisecapella:
Fixes #546

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -287,7 +287,7 @@ def serve_static_assets(request, path):
         raise PermissionDenied()
     filename = os.path.basename(file_path)
     response = StreamingHttpResponse(
-        FileWrapper(open(file_path), 'rb'),
+        FileWrapper(open(file_path, 'rb')),
         content_type=mimetypes.guess_type(file_path)[0]
     )
     response['Content-Length'] = os.path.getsize(file_path)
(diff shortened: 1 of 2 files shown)


ui/views.py
```diff
@@ -287,7 +287,7 @@ def serve_static_assets(request, path):
         raise PermissionDenied()
     filename = os.path.basename(file_path)
     response = StreamingHttpResponse(
-        FileWrapper(open(file_path), 'rb'),
+        FileWrapper(open(file_path, 'rb')),
```
giocalitri, line 290: how did it work before?

noisecapella, line 290: It just returned an empty response. I think there was an exception since the `'rb'` was passed where a block size should have been, and somehow the error was just swallowed up.

noisecapella, line 290: Actually it looks like I missed a typo in the other function too

noisecapella:
Nevermind, false alarm

noisecapella:
Is this good to merge @giocalitri?

ui/views.py
giocalitri, line 290: is this still an issue?

giocalitri:
Missed the sequence of the messages: I think it's good to go

PR #550, opened 2015-08-19
Made creator of a repo an admin during repo creation.

### LINKED ISSUE #549: assign repository admin permissions upon save, not in api views
Currently, `assign_user_to_repo_group` is called in the view of the restful API (`rest/views.py`). This means than any other code which creates a repository must individually import and execute `roles.api.assign_user_to_repo_group`. This can be simplified by explicitly calling `roles.api.assign_user_to_repo_group` during `Repository.save()`, either directly or via a signal.

This is currently blocking development of #512. A quick workaround would be to import and execute `roles.api.assign_user_to_repo_group` in the test, but that just adds to the duplication.

The decision to make this change was made by @giocalitri, @noisecapella, and me via Hangout.

ShawnMilo:
No longer a separate step. Allowed for the deletion of code and
simplification of future tests, including for #512.

Closes #549.

FYI: @giocalitri @noisecapella

### DIFF
--- a/rest/views.py
+++ b/rest/views.py
@@ -93,16 +93,6 @@ class RepositoryList(ListCreateAPIView):
         repo_ids = [repo.id for repo in get_repos(self.request.user.id)]
         return Repository.objects.filter(id__in=repo_ids).order_by('id')
 
-    def perform_create(self, serializer):
-        """Create a new repository."""
-        # pylint: disable=protected-access
-        repo = serializer.save()
-        assign_user_to_repo_group(
-            self.request.user,
-            repo,
-            GroupTypes.REPO_ADMINISTRATOR
-        )
-
 
 class RepositoryDetail(RetrieveAPIView):
     """REST detail view for Repository."""
--- a/roles/signals.py
+++ b/roles/signals.py
@@ -3,7 +3,7 @@ Signals handlers
 """
 from __future__ import unicode_literals
 
-from django.db.models.signals import post_migrate
+from django.db.models.signals import post_migrate, post_save
 from django.dispatch import receiver
 
 from learningresources.models import Repository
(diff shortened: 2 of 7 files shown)


rest/views.py
```diff
@@ -95,13 +95,7 @@ def get_queryset(self):
...
     def perform_create(self, serializer):
         """Create a new repository."""
-        # pylint: disable=protected-access
-        repo = serializer.save()
-        assign_user_to_repo_group(
-            self.request.user,
-            repo,
-            GroupTypes.REPO_ADMINISTRATOR
-        )
+        serializer.save()
```
noisecapella: This is the default behavior of `perform_create` so you should be able to just remove this method

roles/signals.py
```diff
@@ -20,3 +20,24 @@ def my_callback(sender, **kwargs):
...
+    """
+    Give the creator of a repository admin permissions
+    over that repository.
+    """
+    from roles.api import assign_user_to_repo_group, roles_init_new_repo
+    from roles.permissions import GroupTypes
+    instance = kwargs.pop("instance", None)
+    if not kwargs["created"]:
+        return
+    if instance.__class__.__name__ != "Repository":
```
noisecapella: Since we're importing `Repository` anyway can't we compare this against `Repository.__name__` instead?

noisecapella: Or just compare classes for equality if that works as expected in Python

noisecapella:
Looks good other than the two minor comments

ShawnMilo:
@noisecapella: Thanks. Changes made. I left it as a separate commit for now; I'll flatten when it's good to merge.

ShawnMilo:
I suspect the coverage drop is due to the fact that this PR is mostly deleting code.

ShawnMilo:
Just refreshed from master and flattened. Will merge when Travis passes.

PR #551, opened 2015-08-20
Changes password hashing during tests.

ShawnMilo:
This speeds up test execution by about 10%.
On my machine, the full test suite dropped from 10:43 to 9:36 on a sample run.

@bdero 
@carsongee

### DIFF
--- a/learningresources/tests/base.py
+++ b/learningresources/tests/base.py
@@ -18,6 +18,7 @@ from django.core.files.storage import default_storage
 from django.core.management import call_command
 from django.test import Client
 from django.test.testcases import TestCase
+from django.test.utils import override_settings
 import haystack
 
 from learningresources.api import (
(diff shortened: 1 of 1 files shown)


noisecapella:
Picking two Travis builds at random there's not much of an improvement although this is just two data points: [my PR which is reasonably close to master](https://travis-ci.org/mitodl/lore/builds/76332526) and [this PR](https://travis-ci.org/mitodl/lore/builds/76478086). Did you find this slowdown during profiling?

ShawnMilo:
I didn't do any profiling. I just ran a bunch of A/B tests with different things modified that I learned from blog posts. This one seemed like the easiest big win.

One thing about the Travis builds is that their execution time has varied greatly depending upon time of day (presumably server load). If you have a chance and are interested, please try the same branch locally, with and without this change.

noisecapella:
Would it be possible to do something like this instead?

```
@override_settings(PASSWORD_HASHERS=('django.contrib.auth.hashers.MD5PasswordHasher',))
class LoreTestCase(TestCase):
    ...
```

ShawnMilo:
That makes sense. Did you try it?

noisecapella:
I just did, it has the same effect so I'd prefer it over the other solution. I also ran the tests locally and got a 22% reduction in time, so it seems worth it

ShawnMilo:
@noisecapella: Good to merge?

noisecapella:
Looks good to me :+1:

ShawnMilo:
lore_comment_timing

Ha ha ha. Timing.

Issue #516, opened 2015-08-13
Research Elasticsearch DSL


ShawnMilo:
The purpose of this issue is to evaluate the Elasticsearch DSL project for use in LORE. This would enable us to eliminate Haystack (which is broken and poorly maintained of late), and have access to all of Elasticsearch's functionality instead of the lowest common feature set exposed by Haystack for all its backends.

@ferdi found [this](http://elasticsearch-dsl.readthedocs.org/en/latest/index.html#), which is an abstraction of the the `elasticsearch` library we're already using.

In short, it wraps `elasticsearch`'s HTTP/JSON API, which is extremely verbose (requiring the creation of a lot of nested dictionaries), with something that looks a lot more like Django's ORM. It still requires a good understanding of how Elasticsearch works. It does not _replace_ Haystack's functionality. Instead, it throws it away in favor of better control and access to Elasticsearch's full feature set.

There was a [talk](https://www.youtube.com/watch?v=hE359wuHyhs) given about it at the latest DjangoCon given by the author of [the project](https://github.com/elastic/elasticsearch-dsl-py).

---

Example from their docs:

This JSON:

``` json
response = client.search(
    index="my-index",
    body={
      "query": {
        "filtered": {
          "query": {
            "bool": {
              "must": [{"match": {"title": "python"}}],
              "must_not": [{"match": {"description": "beta"}}]
            }
          },
          "filter": {"term": {"category": "search"}}
        }
      },
      "aggs" : {
        "per_tag": {
          "terms": {"field": "tags"},
          "aggs": {
            "max_lines": {"max": {"field": "lines"}}
          }
        }
      }
    }
)
```

is replaced by this Python:

``` python
s = Search(using=client, index="my-index") \
    .filter("term", category="search") \
    .query("match", title="python")   \
    .query(~Q("match", description="beta"))

s.aggs.bucket('per_tag', 'terms', field='tags') \
    .metric('max_lines', 'max', field='lines')

response = s.execute()
```

ShawnMilo:
Compared to Haystack letting a bug that actually makes basic functionality of the project _not work_ for nearly five months, the author of Elasticsearch DSL just closed a ticket after 17 days and [apologized](https://github.com/elastic/elasticsearch-dsl-py/pull/217#event-382529749) for the delay.

pdpinch:
Hardest part is learning the elastic search query language, since this is closer to it than how Haystack does it. 

Probably should change everything all at once: querying and indexing

[...]

PR #554, opened 2015-08-20
Added REST endpoint for search

### LINKED ISSUE #553: REST search endpoint
This would be necessary to allow our javascript to query for search updates without doing a page refresh each time.

noisecapella:
Fixes #553

### DIFF
--- a/rest/serializers.py
+++ b/rest/serializers.py
@@ -20,6 +20,7 @@ from rest_framework.serializers import (
     FileField,
     SerializerMethodField,
     IntegerField,
+    FloatField,
 )
 
 from rest.util import LambdaDefault, RequiredBooleanField
@@ -289,3 +290,18 @@ class LearningResourceExportTaskSerializer(Serializer):
     id = CharField()
     status = CharField()
     url = CharField()
+
+
+class RepositorySearchSerializer(Serializer):
+    """Serializer for search metadata."""
+    resource_type = CharField()
+    course = CharField()
+    run = CharField()
+    xa_nr_views = IntegerField(source="nr_views")
+    xa_nr_attempts = IntegerField(source="nr_attempts")
+    xa_avg_grade = FloatField(source="avg_grade")
+    lid = IntegerField()
+    title = CharField()
+    description = CharField()
+    description_path = CharField()
+    preview_url = CharField()
--- a/rest/tests/test_search.py
+++ b/rest/tests/test_search.py
@@ -1,0 +1,360 @@
+"""
+Tests for search endpoint.
+"""
+
+from __future__ import unicode_literals
+
+from six.moves import urllib_parse
+
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_405_METHOD_NOT_ALLOWED,
+)
+
+from learningresources.api import create_repo
+from learningresources.models import LearningResource, get_preview_url
+from rest.tests.base import (
+    RESTTestCase,
+    REPO_BASE,
+    as_json,
+)
+from taxonomy.models import Term, Vocabulary
+
+
+class TestSearch(RESTTestCase):
+    """
+    Tests for search endpoint.
+    """
+
+    def get_results(self, query="", selected_facets=None, sortby=""):
+        """Helper method to get search results."""
+        if selected_facets is None:
+            selected_facets = []
+
+        selected_facets_arg = ""
+        for facet in selected_facets:
+            selected_facets_arg += "&selected_facets={facet}".format(
+                facet=urllib_parse.quote_plus(facet)
+            )
+        resp = self.client.get(
+            "{repo_base}{repo_slug}/search/?q={query}{facets}"
+            "&sortby={sortby}".format(
+                repo_base=REPO_BASE,
+                repo_slug=self.repo.slug,
+                query=urllib_parse.quote_plus(query),
+                facets=selected_facets_arg,
+                sortby=sortby
+            )
+        )
+        self.assertEqual(HTTP_200_OK, resp.status_code)
+        return as_json(resp)
+
+    def assert_result_equal(self, result, resource):
+        """Helper method to assert result == resource."""
+
+        self.assertEqual(
+            {
+                'course': resource.course.course_number,
+                'description': resource.description,
+                'description_path': resource.description_path,
+                'lid': resource.id,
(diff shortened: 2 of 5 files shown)


noisecapella:
@ShawnMilo Any advice on the parsing of the query string? I'm doing a lot of adhoc stuff which might exist in haystack somewhere

ShawnMilo:
No, what you're doing looks like the same stuff Giovanni and I have already done in other places. If it works and you don't think it's too ugly, I'd leave it alone, considering we're planning on getting rid of Haystack anyway (see #516).

noisecapella:
Marking this work in progress to add a `facet_counts` key to the response

rest/tests/test_search.py
```diff
@@ -0,0 +1,303 @@
...
+                repo_base=REPO_BASE,
+                repo_slug=self.repo.slug,
+                query=urllib_parse.quote_plus(query),
+                facets=selected_facets_arg
+            )
+        )
+        self.assertEqual(HTTP_200_OK, resp.status_code)
+        return as_json(resp)
+
+    def assertResultEqual(self, result, resource):
```
giocalitri: this function name should be with underscores, not camelcase

noisecapella: It wouldn't be consistent with the `unittest` `assertEqual` functions then. I don't have a strong preference however

giocalitri: I think this should be changed, especially to make sure there is no confusion with the `unittest` methods

rest/tests/test_search.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Verify that a blank search will retrieve all results.
+        """
+        no_q_url = "{repo_base}{repo_slug}/search/".format(
+            repo_base=REPO_BASE,
+            repo_slug=self.repo.slug,
+        )
+
+        count = self.get_results()['count']
+        self.assertEqual(1, count)
```
giocalitri: I don't think that 1 is a good number to test "all results". You should have at least 2

giocalitri:
Code looks good: testing functionality

giocalitri:
I think the `apiary.apib` needs to be updated

rest/serializers.py
```diff
@@ -289,3 +290,19 @@ class LearningResourceExportTaskSerializer(Serializer):
     id = CharField()
     status = CharField()
     url = CharField()
+
+
+class RepositorySearchSerializer(Serializer):
+    """Serializer for search metadata."""
+    text = CharField()
```
giocalitri: Question: what is this field? Why do you need it?

noisecapella: This matches up with `text` in `LearningResourceIndex`. It's basically a concatenation of title, description, and content, see `prepare_text` in `LearningResourceIndex`

noisecapella: Actually, if we don't need it in the UI we might as well get rid of it. It's probably in the search index to aid in full text searches.

giocalitri: So it is useless for the interface, because for the panel you need the version with the XML and probably you need to make another request anyway.

giocalitri: unless we introduce highlighted snippets... in that case it can be useful, but it is not the case now

giocalitri: @noisecapella Yep: we wrote our comments at the same time :-)

rest/tests/test_search.py
```diff
@@ -0,0 +1,284 @@
...
+                'course': resource.course.course_number,
+                'description': resource.description,
+                'description_path': resource.description_path,
+                'resource_type': resource.learning_resource_type.name,
+                'lid': resource.id,
+                'xa_nr_attempts': resource.xa_nr_attempts,
+                'xa_nr_views': resource.xa_nr_views,
+                'preview_url': get_preview_url(resource),
+                'run': resource.course.run,
+                'title': resource.title
```
giocalitri: NIT: can you please sort or just group these elements for better readability?

giocalitri:
I think there is a bug with the sorting: I noticed that what you do in the rest form is sorting on line `809` of `rest/views.py`.

The current sorting is handled by `search.forms.SearchForm`: if you check the code there is a secondary `order_by` that uses the default field `LoreSortingFields.BASE_SORTING_FIELD`.

noisecapella:
Good catch, the default field is also wrong in my case (should be `xa_nr_views`, not `nr_views`). I'll fix both problems

giocalitri:
In general, your REST search view looks pretty different from the `RepositoryView` in `ui.views`.
I was expecting from this PR more re-use of that code with a serializer and less re-write (hope this is clear), even if I'm not sure that is even possible.

giocalitri:
I think what is left to do is:
- [ ] Sorting + tests
- [ ] Clean up the code for the old repository view (may require a new issue because the work on the interface is not done yet)
- [ ] update the `apiary` file

noisecapella:
The old repository view needs to stay for now since we're not yet at the point of replacing it. I'm not sure what we can do to increase the amount of code shared. Both use `SearchQuerySet` but we're due to get rid of it anyway when Haystack is replaced. We could also get rid of `RepositoryView` at some future date which would leave us with one code path for search.

noisecapella:
apiary should now be up to date, sorting should also be working. `sortby` takes indexing fields which aren't quite the same as the serializer so people will be using `nr_views` instead of `xa_nr_views`, for example. But this matches the query string for the repository page anyway so it's probably for the best.

rest/views.py
```diff
@@ -760,3 +766,109 @@ def get_object(self):
...
+
+class RepositorySearchList(GenericViewSet):
+    """
+    View of search results for repository search.
+    """
+    serializer_class = RepositorySearchSerializer
+    permission_classes = (ViewRepoPermission, IsAuthenticated)
+
+    def get_queryset(self):
+        query = self.request.GET.get('q')
```
giocalitri: can this be written as `self.request.GET.get('q', '')`

noisecapella: Yes, I'll make this change

noisecapella:
I'm going to squash these commits. Any other comments or something I missed before I merge?

giocalitri:
I'm checking

giocalitri:
Everything looks fine, functionality included. 
Feel free to merge as soon as tests pass

noisecapella:
:tada:

PR #555, opened 2015-08-21
Added support to the UI for free (on the fly) terms in the learningresource panel

### LINKED ISSUE #510: "Tag Style (on the fly)" terms don't work in the learning resource panel
As a curator, I'd like to be able create new terms while adding metadata to resources. Choosing "Tag Style" should allow me to create free form terms, but it apparently does not.

giocalitri:
Closes #510

### DIFF
--- a/ui/jstests/test-learning-resource.jsx
+++ b/ui/jstests/test-learning-resource.jsx
@@ -49,6 +49,12 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     "label": "notrequired",
     "weight": 1
   };
+  var termResponseMedium = {
+    "id": 5,
+    "slug": "medium",
+    "label": "medium",
+    "weight": 1
+  };
   var vocabularyResponsePrereq = {
     "id": 1,
     "slug": "prerequisite",
@@ -120,6 +126,11 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
         url: '/api/v1/repositories/repo/vocabularies/?type_name=course&page=2',
         type: 'GET',
         responseText: vocabulariesResponseSecond
+      });
+      TestUtils.initMockjax({
+        url: '/api/v1/repositories/repo/vocabularies/difficulty/terms/',
+        type: 'POST',
+        responseText: termResponseMedium
       });
     },
     afterEach: function() {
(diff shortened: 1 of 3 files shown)


ui/jstests/test-learning-resource.jsx
```diff
@@ -220,38 +231,88 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+            assert.equal($termsOptions[0].selected, true);
             assert.equal($termsOptions[1].selected, false);
-            // the second vocabulary can be a multi select
-            $(termsSelect)
-              .val(["hard", "easy"])
-              .trigger('change');
-            React.addons.TestUtils.Simulate.change(termsSelect);
+            assert.equal($(termsSelect).val(), "required");
+
+            // remove selection for this vocabulary to not interfer to the rest of the test
```
noisecapella: Typo: interfere

ui/jstests/test-learning-resource.jsx
```diff
@@ -220,38 +231,88 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
-              assert.equal($termsOptions[0].selected, true);
-              assert.equal($termsOptions[1].selected, true);
-              // be sure that the state reflects the selection
-              var terms = _.map(component.state.vocabulariesAndTerms,
-                function (tuple) {
-                  return tuple.selectedTerms;
+              // Switch to difficulty
+              $vocabSelect.val("difficulty").trigger('change');
+              component.forceUpdate(function() {
+                // re-fetch the selects to get the prerequisite one
```
noisecapella: It seems like you are refetching to get the difficulty selects not the prerequisite

ui/static/ui/js/learning_resources.jsx
```diff
@@ -120,8 +127,65 @@ define('learning_resources', [
...
         }), 'value');
-
-      this.props.setValues(this.props.selectedVocabulary.slug, selectedValues);
+      // check if the current vocabulary allows free tagging and in case add
+      // the new tags before proceeding
+      if (this.props.selectedVocabulary.vocabulary_type === 'f') {
+        // extract the list of terms not in the vocabulary existing terms
+        var termsSlugs = _.pluck(this.props.selectedVocabulary.terms, 'slug');
+        var termsToCreate = _.filter(selectedValues, function (slug) {
+          return _.indexOf(termsSlugs, slug) === -1;
```
noisecapella: You can use `_.includes` here instead

ui/static/ui/js/utils.jsx
```diff
@@ -116,6 +116,12 @@ define("utils", ["jquery", "lodash", "react", "select2"],
         allowClear = !isMultiTerms;
       }
 
+      // Allow terms on the fly if the vocabulary is defined to do so
+      var allowTags = false;
+      if (this.props.vocabularyType === 'f') {
```
noisecapella: Can you change `this.props.vocabularyType` to `this.props.allowTags` and make it a boolean? Ideally the `Select2` component should be reusable in many different contexts so we shouldn't use vocabulary or term-specific wording

ui/static/ui/js/learning_resources.jsx
```diff
@@ -120,8 +127,65 @@ define('learning_resources', [
...
+            .done(function(newTerm) {
+              // replace che current label in the selected
+              // values with the newly created slug
+              selectedValues.splice(
+                _.indexOf(selectedValues, termLabel),
+                1,
+                newTerm.slug
+              );
+              // add the term to the current selected vocabulary
+              thiz.props.selectedVocabulary.terms.push(newTerm);
```
noisecapella: We should not alter `props`, `props` are supposed to be immutable. Either use a callback to have the parent make this change in the props, or move this functionality into the state if necessary.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -120,8 +127,65 @@ define('learning_resources', [
...
+              // finally set the current selected valuse
```
noisecapella: Typo: values

ui/static/ui/js/learning_resources.jsx
```diff
@@ -120,8 +127,65 @@ define('learning_resources', [
...
+        var termsSlugs = _.pluck(this.props.selectedVocabulary.terms, 'slug');
+        var termsToCreate = _.filter(selectedValues, function (slug) {
+          return _.indexOf(termsSlugs, slug) === -1;
+        });
+        // if there are no new terms to create, set the state
+        if (!termsToCreate.length) {
+          this.props.setValues(
+            this.props.selectedVocabulary.slug, selectedValues
+          );
+        } else {
```
noisecapella, line 147: It is unlikely but what happens if a term is added and another term removed in the same event. Here the else condition doesn't handle the remove case.

This seems tricky since you would ideally want to call `setValues` once, otherwise one of the calls may overwrite the other call. You can do this with a promise, for example:

```
  var promises = _.map(termsToCreate, function(term) {
    return $.post() // post to create new term
  }
  return $.when.apply($, promises).then(function () {
    var args;
    if (promises.length === 1) {
      args = [arguments[0]];
    } else {
      args = arguments;
    }
    var newSlugs = ... // extract slugs from promise results in args
    thiz.props.setValues(thiz.props.selectedVocabulary.slug, newSlugs)
  });
```

giocalitri, line 147: I'm not sure I understood: If there is one term added and another removed in the same event (and I don't think this can even happen) it means that the selected terms will include the one added and not the one removed and everything else will work just fine, even if the added term is a new term.

noisecapella, line 147: I was thinking about it and you're right in that this case wouldn't happen since we're only inside this branch if there's at least one term to create.

I'm still uneasy about the `splice` below and how the new terms are mixed with the old slug values in the same array. I'd prefer if we remove the new terms from `selectedValues` after `termsToCreate` is set, and do the single `setValues` call after all POSTS complete where we add the new terms with the proper slugs. But I don't see a logical error in your approach, just my opinion on code cleanliness so feel free to go either way

noisecapella:
Functionality works as expected, terms with spaces work fine

giocalitri:
I updated the code, but I did not change it to have all the ajax calls run together: I think it is so unlikely that there will be more terms to create in the event, that it's not worth the time.

giocalitri:
I will squash as soon as you say the code looks good

noisecapella:
That's fine with me, feel free to merge

PR #557, opened 2015-08-21
Stripped caching out of vocabularies during indexing.

### LINKED ISSUE #556: Strip vocabulary caching out of Elasticsearch indexing.
It causes a bug which isn't consistently duplicated, and also slows down imports. Strip it out for now and we can worry about it later if it becomes a problem.

ShawnMilo:
Fixes #556.

### DIFF
--- a/search/tests/test_indexing.py
+++ b/search/tests/test_indexing.py
@@ -205,11 +205,11 @@ class TestIndexing(SearchTestCase):
             return count
 
         set_cache_timeout(0)
-        with self.assertNumQueries(23):
+        with self.assertNumQueries(20):
             self.assertEqual(get_count(), 0)
 
         set_cache_timeout(60)
-        with self.assertNumQueries(8):
+        with self.assertNumQueries(9):
             self.assertEqual(get_count(), 1)
 
     def test_course_cache(self):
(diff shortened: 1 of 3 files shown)


search/tests/test_indexing.py
```diff
@@ -205,11 +205,11 @@ def get_count():
             return count
 
         set_cache_timeout(0)
```
giocalitri, line 207: what's this for now?

giocalitri:
Looks there are mostly deletions, so it looks good.

search/tests/test_indexing.py
ShawnMilo, line 207: Since both course and vocabularies are queried in `get_count()` (above), there is still value in explicitly enabling and disabling caching for the `self.assertNumQueries` tests (below).

giocalitri, line 207: got it

noisecapella:
:+1:

ShawnMilo:
Thanks, @noisecapella. Once tests pass I'll merge, assuming @carsongee lets us know the result of the profiling was good.

carsongee:
Profile checks out

carsongee:
Profile by the way if you want to load it in: https://www.dropbox.com/s/ay9zvbh0t737624/import_file.profile?dl=0  It looks like we probably have a pretty good performance opportunity with get_or_create just being create during the import since get is taking up about 22% of the time on large course imports.

PR #563, opened 2015-08-24
Optimized Dockerfile to reduce build times

carsongee:
- Folded several commands into single RUNs
- Switched to COPY from ADD
- Moved COPY's that were grouped together to only get added at the last
  possible minute
- New flow of system -> python -> node -> source
- Copied .gitignore into .dockerignore

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,79 @@
-.git/
+__pycache__/
+*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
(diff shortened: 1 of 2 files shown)


carsongee:
@bdero since you were interested

noisecapella:
On my laptop the old build was 5m39 and the new build was 5m35. But some of the changes like using COPY, .dockerignore, and folding commands together are probably worth doing anyway. I'll try out the new build and merge if it looks good

PR #564, opened 2015-08-24
Refactored facet view as React component

### LINKED ISSUE #565: Use React for showing facets
This is necessary as part of #455

noisecapella:
Fixes #565


ui/static/ui/js/listing_resources.jsx
```diff
@@ -1,7 +1,138 @@
...
+              </ul>
+            </div>
+          </div>
+        </div>;
+      }
+    });
+
+    var Facets = React.createClass({
+      render: function() {
+        var thiz = this;
```
jamiefolsom, line 98: Remind me why we have to use thiz here again?

noisecapella, line 98: In the `makeFacetGroup` function it's used to refer to the `this` from the render function

jamiefolsom:
**General**
- [x] UI and functionality works.
- [x] No js errors in console.
- [x] No XHR errors.
- [x] All JS tests pass.

**Questions**
- [x] Is it the nesting of components which necessitates aliasing `var thiz = this`?
- [x] Generally speaking, what's your thought about when an index like this, with lots of sub-components should be broken down into smaller files organized by function/role? In React, are there advantages/disadvantages to doing that? We're almost at 500 lines in this one page now.

ui/static/ui/js/listing_resources.jsx
```diff
@@ -1,7 +1,138 @@
-define('listing_resources', ['react', 'jquery', 'lodash'],
-  function (React, $, _) {
+define('listing_resources', ['react', 'jquery', 'lodash', 'utils'],
+  function (React, $, _, Utils) {
     'use strict';
 
+    var ICheckbox = Utils.ICheckbox;
+
+    var getImageFile = function(resourceType) {
+      if (resourceType === 'chapter') {
```
jamiefolsom, line 8: Any advantage/disadvantage here to using a case statement instead?

jamiefolsom, line 8: Or rather, a switch statement...

noisecapella, line 8: I don't think there's a difference. In any case this code was already here, it was just moved up the page

jamiefolsom, line 8: :+1:

noisecapella:
I can move the new code into `facets.jsx` if you think it would make it cleaner. Regarding `thiz`, it will be necessary in most cases where we use closures. There are some alternatives:
- Use `bind` to set the inner function's `this` to the outer function's this. I think `thiz` is a little more explicit but I don't have a strong opinion there
- Create variables for whatever we refer to in `this` in the outer scope and then only refer to those variables in the inner scope.

jamiefolsom:
I don't have a strong opinion on either of these things. I am more just curious to hear your thoughts. Fine by me the way they are.

jamiefolsom:
@giocalitri I've done an initial review of this one -- see the checklist [above](https://github.com/mitodl/lore/pull/564#issuecomment-137242467). I see no issues, although the coverage check is failing at present.

giocalitri:
@jamiefolsom  I think this PR is well covered by you. I'm skipping it

jamiefolsom:
I say :+1:

PR #570, opened 2015-08-25
Added ElasticBeanstalk configuration

carsongee:


pdpinch:
@bdero do we still want to do this?

bdero:
Not at the moment, although this could come in handy for the future. It also serves as a good reference for ElasticBeanstalk config in general.

Issue #487, opened 2015-08-07
Export button visibility


Ferdi:
I'm not sure  #462 is right.  The export button should only be visible if there are resources to be exported.  By default does not show up at all.

ShawnMilo:
To clarify, do you mean the button never appears until after the user has clicked on some resources to populate their "shopping cart"? Regardless of whether any resources are currently shown on-screen. Once there are items in the cart, the button remains until the items are deleted or the cart is emptied. Is that correct?

Ferdi:
Right, that's what I meant. Makes sense ?

ShawnMilo:
Yes, that makes sense. Like a "check out" button not being available until there's something in the cart.

For uniformity of the UI, do we want to hide it or just disable it?

amir-qayyum-khan:
For this task I will create a PR on the top of George work. Because George is working on refactoring listing.
- In react component inside rendering function i will add check if **Learning resource** count is greater than 0  only then I will add button to container.

cc @pdpinch

PR #474, opened 2015-08-06
Refactored listing resources to use React

noisecapella:
Fixes #455

Issue #455, opened 2015-07-31
Rewrite listing page in React


noisecapella:
This would have several benefits:
- No operations would require a page refresh anymore
- Easier to test listing page (currently there are no tests covering this part)
- Cleaner code
- Easier integration with existing React components

Potential complications:
- Integration with haystack text caching

Any thoughts?

Ferdi:
Each learning resource will be a react component ?  What else will change ?

noisecapella:
There would be a component managing everything as well as other smaller components like the panels, learning resources, and so on. For example if a user wanted to search by a learning resource type, they would click the checkbox, that component would tell the main component that it needs to update the resources on the page, the main component would ask the API for a new set of information and it would rerender the resources in place instead of causing a page refresh like we do now.

amir-qayyum-khan:
In upcoming PR i am going to divide Learning resource in to react components like for example
- Editable description component 
- Learning resource 
- Learning resource list

A  will bind (reuse) panels like learning resource panel, export panel with Learning resource listing.
Will handle permissions and other events in react class.

What do you say @pdpinch @Ferdi

Ferdi:
Would it make sense to start with only "Editable description component" ?   And do the rest on separate PRs ?

noisecapella:
Would anyone mind if I did this instead? I've run out of things to do and I have an idea how to implement this

PR #573, opened 2015-08-28
Made export button only visible if there are resources to be exported

amir-qayyum-khan:
In this PR I made export button only visible if there are resources to be exported. By default does not show up at all.

Issue https://github.com/mitodl/lore/issues/487
PR https://github.com/mitodl/lore/pull/474/files
cc @noisecapella @pdpinch 

screen shot 2015-08-28 at 6 25 10 pm

screen shot 2015-08-28 at 6 25 15 pm


noisecapella:
Given the issue it makes sense to only show the export button if there's a count to show, but we should remove `show_export_button` because it's not necessary anymore. Can you remove `show_export_button` from `ui/views.py:259` and everything that depends on it, in both Python and Javascript?

jamiefolsom:
There's a slight jag when the spinner shows/hides. Hard coding it to show in the Chrome dev tools panel, it's possible to see that it's because in addition to "spinning", the whole gif is moving in a circular pattern.

spinner

noisecapella:
Actually I'll do that portion myself since the PR being merged into is in the middle review at the moment. PR #455 will have this functionality adopted into it

Issue #506, opened 2015-08-13
Editing a term in taxonomy panel


amir-qayyum-khan:
##### Requirement
- [ ] Editing a term should make the term editable in place in the taxonomies panel

Epic https://github.com/mitodl/lore/issues/380
##### Details

On clicking Edit term, I will add in place editing box. where user can edit and save that term. And there will be cancel button to exit edit mode. On save I will re-render list.

cc @pdpinch

PR #574, opened 2015-08-31
Added inline editing feature for terms in taxonomy panel.

amir-qayyum-khan:
Fixed https://github.com/mitodl/lore/issues/506

@pdpinch @carsongee 

screen shot 2015-08-31 at 5 53 12 pm
screen shot 2015-08-31 at 5 53 27 pm
screen shot 2015-08-31 at 5 53 32 pm
screen shot 2015-08-31 at 5 53 41 pm

### DIFF
--- a/ui/jstests/test_manage_taxonomies.jsx
+++ b/ui/jstests/test_manage_taxonomies.jsx
@@ -194,27 +194,252 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
   QUnit.test('Assert that TermComponent renders properly',
     function(assert) {
       assert.ok(TermComponent, "class object not found");
+      var done = assert.async();
       var term = {
         "id": 9,
         "slug": "test",
         "label": "test",
         "weight": 1
       };
-      var termComponentRendered = React.addons.TestUtils.
+      var parentUpdateCount = 0;
+      var updateTerm = function() {
+        parentUpdateCount += 1;
+      };
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/test/",
+        responseText: term,
+        type: "PATCH"
+      });
+      var afterMount = function(component) {
+        var labels = React.addons.TestUtils.
+        scryRenderedDOMComponentsWithTag(
+          component,
+          'label'
+        );
+
+        //testing label value render
+        var termLabel = labels[0];
+        var label = termLabel.getDOMNode();
+        assert.equal(label.innerHTML, term.label);
+        assert.equal(component.state.formatActionState, 'show');
+
+        var formatButton = React.addons.TestUtils.
+          findRenderedDOMComponentWithClass(
+            component,
+            'format-button'
+        );
+        var cancelButton = React.addons.TestUtils.
+          findRenderedDOMComponentWithClass(
+            component,
+            'revert-button'
+        );
+
+        //open edit mode
+        React.addons.TestUtils.Simulate.click(formatButton);
+        component.forceUpdate(function() {
+          assert.equal(component.state.formatActionState, 'edit');
+          //edit term
+          var editTermBox = React.addons.TestUtils.
+            findRenderedDOMComponentWithTag(
+            component,
+            'input'
+          );
+          React.addons.TestUtils.Simulate.change(
+            editTermBox, {target: {value: "TestB"}}
+          );
+          component.forceUpdate(function() {
+            assert.equal(component.state.label, "TestB");
+            //save term
+            React.addons.TestUtils.Simulate.click(formatButton);
+            component.forceUpdate(function() {
+              //after saved term using api
+              waitForAjax(1, function() {
+                //term state reset
+                assert.equal(component.state.formatActionState, 'show');
+                // term is update in parent
+                assert.equal(parentUpdateCount, 1);
+
+                // Edit term again
+                React.addons.TestUtils.Simulate.click(formatButton);
+                component.forceUpdate(function() {
+                  assert.equal(component.state.formatActionState, 'edit');
+                  editTermBox = React.addons.TestUtils.
+                    findRenderedDOMComponentWithTag(
+                    component,
+                    'input'
+                  );
+                  React.addons.TestUtils.Simulate.change(
+                    editTermBox, {target: {value: "TestB"}}
+                  );
+                  component.forceUpdate(function () {
+                    // press cancel button and assert term layout is reset.
+                    React.addons.TestUtils.Simulate.click(cancelButton);
+
+                    component.forceUpdate(function () {
+                      assert.equal(component.state.formatActionState, 'show');
+                      //assert editbox is hide (UI reset)
+                      var editTermBoxes = React.addons.TestUtils.
+                        scryRenderedDOMComponentsWithTag(
+                        component,
+                        'input'
+                      );
+
+                      assert.equal(editTermBoxes.length, 0);
+
+                      // Edit again term with same label
+                      React.addons.TestUtils.Simulate.click(formatButton);
+                      component.forceUpdate(function() {
+                        assert.equal(component.state.formatActionState, 'edit');
+                        editTermBox = React.addons.TestUtils.
+                          findRenderedDOMComponentWithTag(
+                          component,
+                          'input'
+                        );
+                        React.addons.TestUtils.Simulate.change(
+                          editTermBox, {target: {value: "test"}}
+                        );
+
+                        component.forceUpdate(function() {
+                          // save term with label equals to previous label
+                          React.addons.TestUtils.Simulate.click(formatButton);
+
+                          component.forceUpdate(function() {
+                            //assert editbox is hide (UI reset)
+                            editTermBoxes = React.addons.TestUtils.
+                              scryRenderedDOMComponentsWithTag(
+                              component,
+                              'input'
+                            );
+                            assert.equal(editTermBoxes.length, 0);
+                            done();
+                          });
+                        });
+                      });
+                    });
+                  });
+                });
+              });
+            });
+          });
+        });
+      };
+
+      React.addons.TestUtils.
         renderIntoDocument(
           <TermComponent
             term={term}
+            repoSlug="repo"
+            updateTerm={updateTerm}
+            vocabularySlug={vocabulary.slug}
+            ref={afterMount}
           />
         );
+    }
+  );
 
-      var labelComponent = React.addons.TestUtils.
-        findRenderedDOMComponentWithTag(
-          termComponentRendered,
+  QUnit.test('Assert that ajax fail on update term TermComponent',
+    function(assert) {
+      assert.ok(TermComponent, "class object not found");
+      var done = assert.async();
+      var term = {
+        "id": 9,
+        "slug": "test",
+        "label": "test",
+        "weight": 1
+      };
+      var parentUpdateCount = 0;
+      var updateTerm = function() {
+        parentUpdateCount += 1;
+      };
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/test/",
+        responseText: term,
+        type: "PATCH",
+        status: 400
+      });
+      var afterMount = function(component) {
+        var labels = React.addons.TestUtils.
+        scryRenderedDOMComponentsWithTag(
+          component,
           'label'
-      );
-      //testing label value render
-      var label = labelComponent.getDOMNode();
-      assert.equal(label.innerHTML, term.label);
+        );
+        //testing label value render
+        var termLabel = labels[0];
+        var label = termLabel.getDOMNode();
+        assert.equal(label.innerHTML, term.label);
+        assert.equal(component.state.formatActionState, 'show');
+        var formatButton = React.addons.TestUtils.
+          findRenderedDOMComponentWithClass(
+            component,
+            'format-button'
+        );
+        var cancelButton = React.addons.TestUtils.
+          findRenderedDOMComponentWithClass(
+            component,
+            'revert-button'
+        );
+        React.addons.TestUtils.Simulate.click(formatButton);
+        component.forceUpdate(function() {
+          assert.equal(component.state.formatActionState, 'edit');
+          var editTermBox = React.addons.TestUtils.
+          findRenderedDOMComponentWithTag(
+            component,
+            'input'
+          );
+          React.addons.TestUtils.Simulate.change(
+            editTermBox, {target: {value: "TestB"}}
+          );
+          component.forceUpdate(function() {
+            assert.equal(component.state.label, "TestB");
+            React.addons.TestUtils.Simulate.click(formatButton);
+            component.forceUpdate(function() {
+              waitForAjax(1, function() {
+                assert.equal(component.state.label, "TestB");
+                assert.equal(
+                  component.state.errorMessage, 'Unable to update term'
+                );
+                assert.equal(component.state.formatActionState, 'edit');
+                assert.equal(parentUpdateCount, 0);
+
+                editTermBox = React.addons.TestUtils.
+                  findRenderedDOMComponentWithTag(
+                  component,
+                  'input'
+                );
+                React.addons.TestUtils.Simulate.change(
+                  editTermBox, {target: {value: "TestB"}}
+                );
+                //after unable to save you can reset edit mode
+                component.forceUpdate(function () {
+                  React.addons.TestUtils.Simulate.click(cancelButton);
+                  component.forceUpdate(function () {
+                    assert.equal(component.state.label, "test");
+                    assert.equal(component.state.formatActionState, 'show');
+                    //assert editbox is hide (UI reset)
+                    var editTermBoxes = React.addons.TestUtils.
+                      scryRenderedDOMComponentsWithTag(
+                      component,
+                      'input'
+                    );
+                    assert.equal(editTermBoxes.length, 0);
+                    done();
+                  });
+                });
+              });
+            });
+          });
+        });
+      };
+      React.addons.TestUtils.
+        renderIntoDocument(
+          <TermComponent
+            term={term}
+            repoSlug="repo"
+            updateTerm={updateTerm}
+            vocabularySlug={vocabulary.slug}
+            ref={afterMount}
+          />
+        );
     }
   );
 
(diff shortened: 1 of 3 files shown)


jamiefolsom:
Hi @amir-qayyum-khan, looks good! Looking at UI/functionality now. First thing I notice is a somewhat distracting behavior on the edit/save buttons on individual terms, which is that on click, a text selection is created. See: 
screen recording 2015-09-01 at 09 52 am

This can be remedied using a CSS rule, like: 

```
.noselect {
    -webkit-touch-callout: none;
    -webkit-user-select: none;
    -khtml-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

Which can be applied to the React js className:

```
        this.setState({
          formatActionClassName: "fa fa-save noselect", // note additional class here
          formatActionState: 'save',
          editTextClass: 'form-control edit-term-box-show'
        });
```

and

```
      this.setState({
        formatActionClassName: "fa fa-pencil noselect", // note additional class here
        formatActionState: 'edit',
        editTextClass: 'form-control edit-term-box-hide',
        listClassName: 'form-group',
        errorMessage: ''
      });
```

See:

screen recording 2015-09-01 at 09 56 am

jamiefolsom:
Further comments for @amir-qayyum-khan.
- [x] JS tests all pass.
- [x] Tests for manage taxonomies are getting huge, which may not be avoidable, or an issue, but I would be curious to hear your thoughts on if/how they can be broken out into smaller chunks. Maybe @noisecapella has thoughts? See: https://github.com/mitodl/lore/pull/574/files#diff-0f3ee7bcc78eb4497e0092ea2083d2e5R236 and 50 lines following.
- [x] Editing works, but enter key doesn't save -- user must click save button. Not a huge issue, but would be nice.
- [x] Error styling/user input feedback on saving an empty term is good
- [ ] Message should say: "can't save empty term. If you want to delete a term, please click the delete button (x)" or something?
- [ ] On save: JS warning in the console: "Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op."
- [ ] On save: if the term is empty, the form actually submits a bad XHR request; that request should not be sent. Here's the error: "PUT http://192.168.59.103:8070/api/v1/repositories/test/vocabularies/item-one/terms/asdf1/ 400 (BAD REQUEST)"

noisecapella:
`test_manage_taxonomies` is a mess and should be refactored at some point, but for this PR we should focus on the cleanliness of the code being added. Descriptive comments are probably the most important thing here

jamiefolsom:
@noisecapella understood, thanks.

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+        editTextClass: 'form-control edit-term-box-hide',
+        listClassName: 'form-group',
+        errorMessage: ''
+      });
+      labelSelector.show();
+    },
+    updateTermSubmit: function() {
+      var API_ROOT_VOCAB_URL = '/api/v1/repositories/' + this.props.repoSlug +
+        '/vocabularies/' + this.props.vocabularySlug + '/terms/' +
+        this.props.term.slug;
```
noisecapella: Django URLs should end with a slash (at least for LORE) to avoid a redirect

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
         "label": "test",
         "weight": 1
       };
-      var termComponentRendered = React.addons.TestUtils.
+      var parentUpdateCount = 0;
+      var updateTerm = function() {
+        parentUpdateCount += 1;
+      };
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/test",
```
noisecapella: Add a slash to the end of this URL

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+      var thiz = this;
+      var termData = {
+        label: this.state.label,
+        weight: this.props.term.weight
+      };
+      $.ajax({
+        type: 'PUT',
```
noisecapella: Why not a PATCH?

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+    getInitialState: function() {
+      return {
+        formatActionClassName: 'fa fa-pencil',
+        formatActionState: 'edit',
+        editTextClass: 'form-control edit-term-box-hide',
+        listClassName: 'form-group',
+        label: ''
+      };
+    },
+    componentWillMount: function() {
```
noisecapella: This seems unnecessary. Why not just include this code in `getInitialState`?

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
-          </a> <a href="#">
+          <a onClick={this.formatHandler} className="format-button">
+            <i className={this.state.formatActionClassName}></i>
+          </a> <a onClick={this.revertHandler} className="revert-button">
             <i className="fa fa-remove"></i>
           </a>
-        </span> <label className="term-title"
-        htmlFor="minimal-checkbox-1-11">{this.props.term.label}</label></li>;
+        </span>
+        <label className="term-title" id={this.props.term.id}
```
noisecapella: We shouldn't assign numbers to HTML id attributes, it could easily collide with something else.

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+        htmlFor="minimal-checkbox-1-11">{this.state.label}</label> <input
+        name="exit_term_temp" type="text" id={editBoxId}
```
noisecapella: This name attribute will collide with other terms. If we don't need the name (we shouldn't if we're not submitting it as a form) then we should get rid of it

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+          valueLink={this.linkState('label')}
+          className={this.state.editTextClass}/> <label htmlFor={editBoxId}
```
noisecapella: It looks like you are setting the visibility using CSS for the input textbox. Why not store the visibility in the state as a boolean and use it to tell React whether to render the textbox?

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+        className="help-inline control-label">{this.state.errorMessage}</label>
```
noisecapella: If there is no error message there should be no label tag

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+      </li>;
+    },
+    formatHandler: function() {
+      var formatActionState = this.state.formatActionState;
+      var replaceWith = $('#edit_' + this.props.term.id);
+      var labelSelector = $('#' + this.props.term.id);
```
noisecapella: We should never be using global selectors in React.

For your purpose try using the ref attribute instead: https://facebook.github.io/react/docs/working-with-the-browser.html#refs-and-finddomnode

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+        htmlFor="minimal-checkbox-1-11">{this.state.label}</label> <input
+        name="exit_term_temp" type="text" id={editBoxId}
+          valueLink={this.linkState('label')}
+          className={this.state.editTextClass}/> <label htmlFor={editBoxId}
+        className="help-inline control-label">{this.state.errorMessage}</label>
+      </li>;
+    },
+    formatHandler: function() {
+      var formatActionState = this.state.formatActionState;
+      var replaceWith = $('#edit_' + this.props.term.id);
```
noisecapella: See below comment

ui/static/ui/js/manage_taxonomies.jsx
noisecapella: Actually it looks like `formatActionState` can be used for that purpose

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+    revertHandler : function() {
+      var formatActionState = this.state.formatActionState;
+      if (formatActionState === 'save') {
+        // user is in edit mode. Cancel edit if user presses cross icon.
+        this.resetUtilityFeatures();
+        this.setState({label: this.props.term.label});
+      }
+    },
+    resetUtilityFeatures: function() {
+      var labelSelector = $('#' + this.props.term.id);
```
noisecapella: See above comment about not using global selectors in React

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
   var ICheckbox = Utils.ICheckbox;
 
   var TermComponent = React.createClass({
+    mixins: [React.addons.LinkedStateMixin],
+    getInitialState: function() {
+      return {
+        formatActionClassName: 'fa fa-pencil',
```
noisecapella: It looks like this depends entirely on `formatActionState`, you should remove it from state and calculate it in `render` instead based on the value of `formatActionState`

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -6,24 +6,116 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+        formatActionState: 'edit',
+        editTextClass: 'form-control edit-term-box-hide',
```
noisecapella: You can probably remove this as mentioned in a comment below.

ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -338,11 +431,28 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils'],
...
+      var vocabIndex = _.findIndex(vocabularies, function(vocabularyObj) {
+        return vocabularyObj.vocabulary.slug === vocabSlug;
+      });
+      if (vocabIndex > -1) {
+        var terms = vocabularies[vocabIndex].terms;
+        var termIndex = _.findIndex(terms, function(termObj) {
+          return termObj.id === term.id;
+        });
+        if (termIndex > -1) {
+          vocabularies[vocabIndex].terms[termIndex] = term;
```
noisecapella: You can't ever modify `this.state` or anything attached to it, you have to clone the object first. Why not do something like:

```
  vocabularies = _.map(vocabularies, function(tuple) {
    if (tuple.vocabulary.slug === vocabSlug) {
      var terms = _.map(tuple.terms, function(tupleTerm) {
        if (tupleTerm.id === term.id) {
          return term;
        }
        return tupleTerm;
      });
      return {
        terms: terms,
        vocabulary: tuple.vocabulary
      };
    }
```

Then `setState`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        type: "PUT"
+      });
+      var afterMount = function(component) {
+        var labels = React.addons.TestUtils.
+        scryRenderedDOMComponentsWithTag(
+          component,
+          'label'
+        );
+        //testing label value render
+        var termLable = labels[0];
```
noisecapella: `termLable` should be `termLabel`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+          assert.equal(component.state.formatActionState, 'save');
+          //edit term
+          React.addons.TestUtils.Simulate.change(
+            editTermBox, {target: {value: "TestB"}}
+          );
+          component.forceUpdate(function() {
+            //save term
+            React.addons.TestUtils.Simulate.click(formatButton);
+            component.forceUpdate(function() {
+              //afer saved term using api
```
noisecapella: Change to `after`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        scryRenderedDOMComponentsWithTag(
+          component,
+          'label'
+        );
+        //testing label value render
+        var termLable = labels[0];
+        var label = termLable.getDOMNode();
+        assert.equal(label.innerHTML, term.label);
+        assert.equal(component.state.formatActionState, 'edit');
+        var formatButton = React.addons.TestUtils.
```
noisecapella, line 226: Please add some line breaks to break the test into logical pieces for easier readability

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        //open edit mode
+        React.addons.TestUtils.Simulate.click(formatButton);
+        component.forceUpdate(function() {
+          assert.equal(component.state.formatActionState, 'save');
+          //edit term
+          React.addons.TestUtils.Simulate.change(
+            editTermBox, {target: {value: "TestB"}}
+          );
+          component.forceUpdate(function() {
+            //save term
```
noisecapella, line 252: Add an assertion here for the change in state for the textbox.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+            React.addons.TestUtils.Simulate.click(formatButton);
+            component.forceUpdate(function() {
+              //afer saved term using api
+              waitForAjax(1, function() {
+                //term state reset
```
noisecapella, line 257: Can you assert that the textbox value was reset?

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        "slug": "test",
+        "label": "test",
+        "weight": 1
+      };
+      var parentUpdateCount = 0;
+      var updateTerm = function() {
+        parentUpdateCount += 1;
+      };
+      TestUtils.initMockjax({
+        url: "/api/v1/repositories/repo/vocabularies/difficulty/terms/test",
```
noisecapella: Add trailing slash

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        responseText: term,
+        type: "PUT",
```
noisecapella: Change to PATCH

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        scryRenderedDOMComponentsWithTag(
+          component,
           'label'
-      );
-      //testing label value render
-      var label = labelComponent.getDOMNode();
-      assert.equal(label.innerHTML, term.label);
+        );
+        //testing label value render
+        var termLable = labels[0];
```
noisecapella: Should be `termLabel`

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+          'input'
+        );
+        React.addons.TestUtils.Simulate.click(formatButton);
+        component.forceUpdate(function() {
+          assert.equal(component.state.formatActionState, 'save');
+          React.addons.TestUtils.Simulate.change(
+            editTermBox, {target: {value: "TestB"}}
+          );
+          component.forceUpdate(function() {
+            React.addons.TestUtils.Simulate.click(formatButton);
```
noisecapella, line 394: Assert change in the textbox

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -190,27 +190,217 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+                  component.state.errorMessage, 'Unable to update term'
+                );
+                assert.equal(component.state.formatActionState, 'save');
+                assert.equal(parentUpdateCount, 0);
+                React.addons.TestUtils.Simulate.change(
+                    editTermBox, {target: {value: "TestB"}}
+                );
+                //after unable to save you can reset edit mode
+                component.forceUpdate(function () {
+                  React.addons.TestUtils.Simulate.click(cancelButton);
```
noisecapella, line 414: Assert textbox state here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -972,6 +1167,93 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+            component.state.vocabularies.length,
+            1
+          );
+          assert.equal(
+            component.state.vocabularies[0].terms.length,
+            2
+          );
+          var updateTermUrl = "/api/v1/repositories/repo/vocabularies/" +
+              component.state.vocabularies[0].vocabulary.slug + "/terms/" +
+              component.state.vocabularies[0].terms[0].slug;
```
noisecapella: Add a trailing slash here

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -972,6 +1167,93 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+          TestUtils.initMockjax({
+            url: updateTermUrl,
+            responseText: term,
+            type: "PUT"
```
noisecapella: Use PATCH

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -972,6 +1167,93 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+          var editTermBox = editTermBoxes[0];
+          //open edit mode
+          React.addons.TestUtils.Simulate.click(formatButton);
+          component.forceUpdate(function() {
+            //edit term
+            React.addons.TestUtils.Simulate.change(
+              editTermBox, {target: {value: "TestB"}}
+            );
+            component.forceUpdate(function() {
+              //save term
```
noisecapella, line 1382: Assert the change in the textbox state

noisecapella:
Coverage looks good, functionality works for me. Can @ferdi or @pdpinch comment on whether this fits with their idea for look and feel?

ui/static/ui/js/manage_taxonomies.jsx
amir-qayyum-khan: As you see i need class `form-control` on input text, thats why instead if adding style -> visibility I am using className and adding second class

ui/static/ui/js/manage_taxonomies.jsx
amir-qayyum-khan: @noisecapella  Can you refer me to doc that states do not use jquery in react component. What if I dont use id instead use data-xyz?

amir-qayyum-khan:
Done @noisecapella @jamiefolsom

ui/static/ui/js/manage_taxonomies.jsx
noisecapella: The main problem with global selectors is that they may collide with other elements with the same id outside the react element. Since we can use the `ref` attribute to identify DOM elements instead, we don't have a need for `id` attributes here.

React may delete or modify DOM elements on any render so we should avoid touching the DOM as much as possible, instead using `state` to communicate these changes to the `render` method. This also makes the code cleaner since all rendering changes are in one place. Instead of calling `labelSelector.hide()` you should change the state using `setState` and update `render` to not display the label in that case.

For `replaceWith.focus()` you will need to access the node directly, try something like `$(React.findDOMNode(this.refs.exitTempTerm)).focus()`

noisecapella:
I've submitted a PR with suggested changes: #624 

I'm also wondering why you are using the `no-select` CSS attribute?

amir-qayyum-khan:
@noisecapella 
@jamiefolsom suggested me to use no-select 
https://github.com/mitodl/lore/pull/574#issuecomment-136730739

noisecapella:
I see, makes sense. Thanks for merging my PR @amir-qayyum-khan. I'll demonstrate this functionality to @pdpinch when he comes in then merge this into master.

PR #577, opened 2015-08-31
Added bootstrap as requirement for manage taxonomies

noisecapella:
The use of `$().tab()` in this file requires bootstrap. Depending on how tests are run this requirement may not be satisfied causing the tests to fail. At some point soon we should move that code into a callback passed via props so that it doesn't mess with the global test state, and so it's cleaner.


giocalitri:
Looks good to me: merging

PR #581, opened 2015-09-01
Fixed migration to bulk create rows in through table

noisecapella:
lore-ci is stuck because the migration is taking too long. This changes the insert to use `bulk_create` which will perform much more quickly


learningresources/migrations/0015_backfill_curator_vocabularies.py
```diff
@@ -64,8 +64,18 @@ def backfill_resources(apps, repo_id):
     resources = LearningResource.objects.filter(
         course__repository__id=repo_id
     )
-    for resource in resources.iterator():
-        resource.terms.add(empty_term)
+    empty_term_id = empty_term.id
```
bdero: Why is resolving empty_term.id upfront necessary?

bdero:
:+1:

PR #584, opened 2015-09-02
Added travis-ci build notifications for hipchat & slack

bdero:
The github integration build status notifications aren't very good in hipchat (they don't color code failures or offer much info about the build in general. The travis integration notifications look much better.


bdero:
The build that was triggered from this PR verified that both the hipchat and slack integrations worked. (Look at the "lore" rooms in either chat service. ;)

PR #496, opened 2015-08-11
Fixed counter in learning resouce exports panels header

amir-qayyum-khan:
Hi,

In this PR i made Learning resources select counter in Exports panel header consistent with the resources select in listing.

issue https://github.com/mitodl/lore/issues/471
https://github.com/mitodl/lore/pull/474
@pdpinch @noisecapella 

screen shot 2015-08-11 at 7 54 54 pm

screen shot 2015-08-11 at 7 54 05 pm

PR #586, opened 2015-09-02
Fixed counter in learning resouce exports panels header

### LINKED ISSUE #471: Export panel title's export count is out of sync with rest of page
If you have three exports and you go to the export panel you will see 'Export (3)' in the side. However if you click an export link and then look at the export panel again this number doesn't update.

noisecapella:
Fixes #471 
Originally #496


noisecapella:
Will merge when tests pass

PR #589, opened 2015-09-02
Changed behavior to use AJAX calls for listing page updates

### LINKED ISSUE #566: Use search REST API to populate listing page
Instead of refreshing the page on listing page changes, query the search API and rerender the relevant React components. This should be the last piece for #455

noisecapella:
Fixes #566


ui/static/ui/js/listing.js
```diff
@@ -126,7 +126,57 @@ define('listing',
...
+          queryMap[k] = [v];
+        } else {
+          queryMap[k] = v;
+        }
+      });
+
+      /**
+       * Iterates through query string to figure out which facets
+       * are checked, and return an object with this information.
+       */
```
giocalitri: this looks hacky: maybe we should make this in the backend and return a "tokenized" query parameter object together with the response

noisecapella: will do

ui/static/ui/js/listing.js
```diff
@@ -148,54 +175,121 @@ define('listing',
...
 
-      var facetIsSelected = function(facetId, valueId) {
-        var facetQuery = "selected_facets=" + facetId + "_exact:" + valueId;
-        if (location.search === undefined) {
-          return false;
-        }
-        return location.search.indexOf(facetQuery) !== -1;
+      refreshFromAPI = function() {
+        var newQuery = "?" + URI().search(queryMap).query();
+        // Let's hope nobody's using IE 9 or earlier...
```
giocalitri: @pdpinch is this a safe statement?

ui/static/ui/js/listing.js
```diff
@@ -148,54 +175,121 @@ define('listing',
...
+          listingOptions.repoSlug + "/search/" + newQuery;
+
+        $("#listing").css({opacity: 0.6});
+        return $.get(url).then(function(collection) {
+          listingOptions = $.extend({}, listingOptions);
+          listingOptions.resources = collection.results;
+          listingOptions.facetCounts = collection.facet_counts;
+          selectedFacets = collection.selected_facets;
+
+          numPages = Math.ceil(collection.count / 20);
```
giocalitri: Why we just don't do something like
`numPages = Math.ceil(collection.count / collection.results.length);` ?
If we change the number of learning resources per page in the backend this will update automatically.

ui/static/ui/js/listing.js
```diff
@@ -148,54 +175,121 @@ define('listing',
...
-        return location.search.indexOf(facetQuery) !== -1;
+      refreshFromAPI = function() {
+        var newQuery = "?" + URI().search(queryMap).query();
+        // Let's hope nobody's using IE 9 or earlier...
+        window.history.replaceState("currenturl", "MIT LORE", newQuery);
+
+        var url = "/api/v1/repositories/" +
+          listingOptions.repoSlug + "/search/" + newQuery;
+
+        $("#listing").css({opacity: 0.6});
```
giocalitri: What about putting the "opacity" handling in a specific function so we don't have to change it in multiple places?

giocalitri:
Functionality looks fine.
The only thing I don't like is that we keep loading the taxonomy panel during the search result, but since the behavior has not changed, I cannot call it a bug.

ui/static/ui/js/listing.js
noisecapella: `collection.results.length` may be less than 20 items if we're on the last page. But I'll update `listingOptions` to include this value from the server

ui/static/ui/js/listing.js
noisecapella: Actually it looks like we can use a [polyfill](https://github.com/browserstate/history.js) to support all reasonable browsers (Netscape users will be out of luck). I'll make this change

ui/static/ui/js/listing.js
noisecapella: Sounds good

giocalitri:
this PR has now 129 files changed: waiting for #630 to be merged before reviewing again

noisecapella:
#630 is merged

giocalitri:
After you squash, you can merge if tests pass

PR #597, opened 2015-09-04
Formatted average grade as fixed width number

noisecapella:
Overlooked this minor detail. If you look at lore-rc vs lore-ci you can see that the average grade is not formatted the same way.

### DIFF
--- a/ui/static/ui/js/listing_resources.jsx
+++ b/ui/static/ui/js/listing_resources.jsx
@@ -132,7 +132,7 @@ define('listing_resources', ['react', 'jquery', 'lodash'],
                     </span>
                     <span className="meta-item mi-col-3">
                       <i className="fa fa-graduation-cap">
-                      </i> {resource.xa_avg_grade}
+                      </i> {resource.xa_avg_grade.toFixed(1)}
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/listing_resources.jsx
```diff
@@ -132,7 +132,7 @@ define('listing_resources', ['react', 'jquery', 'lodash'],
                     </span>
                     <span className="meta-item mi-col-3">
                       <i className="fa fa-graduation-cap">
-                      </i> {resource.xa_avg_grade}
+                      </i> {resource.xa_avg_grade.toFixed(1)}
```
giocalitri, line 135: what is `toFixed` supposed to do?

noisecapella, line 135: It outputs the number formatted to that many decimal places. For example:

> 0.3.toFixed(3)
>    '0.300'

It's only here to match Django's default behavior when dealing with floats without a decimal component.

giocalitri, line 135: Ok, so you can merge it

PR #614, opened 2015-09-08
Added tests for listing page

### LINKED ISSUE #615: Add tests for listing.js
listing.js contains a bunch of functionality necessary for the listing page but we don't test any of it. We should test at least the low hanging fruit

noisecapella:
Fixes #615

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -58,8 +58,6 @@ module.exports = function(config) {
 
     // list of files to exclude from coverage and testing
     exclude: [
-      "ui/static/ui/js/listing.js",
-      "ui/static/ui/js/csrf.js"
     ],
 
     // preprocess matching files before serving them to the browser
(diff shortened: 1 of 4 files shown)


noisecapella:
I'm going to call this done for now, we should add more tests but we can do that in a future PR

noisecapella:
Coverage will decrease because previously listing.js and csrf.js were not counted in coverage results

giocalitri:
This has conflicts now

ShawnMilo:
In `ui/static/ui/js/csrf.js` and `ui/static/ui/js/listing.js`, all the function declarations have a space between `function` and `(`, but they don't in other files.

Just an observation. The [Google Javascript style guide](https://google.github.io/styleguide/javascriptguide.xml) doesn't seem to specifically address it, but there are over 100 examples without a space on the page, and only two with a space.

It looks good otherwise.

noisecapella:
This used to be enforced by JSCS but an exception was added to not check for that space (see `.jscsrc` for the rules enforced).

PR #621, opened 2015-09-08
Updated export to preserve static asset path

### LINKED ISSUE #594: Preserve subdirectories in static asset export
In courses like [8.01r](https://github.mit.edu/mitx/content-mit-801r) there are subdirectories in the static assets directory. These directories appear to be preserved on import, but are apparently flattened on export -- which breaks the links from the XML. 

See ZenDesk #2370

noisecapella:
Fixes #594

### DIFF
--- a/exporter/api.py
+++ b/exporter/api.py
@@ -4,6 +4,7 @@ Functions for exporting learning resources
 
 from __future__ import unicode_literals
 
+import errno
 import os
 import re
 import tarfile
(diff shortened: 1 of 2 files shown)


noisecapella:
A little premature, this also includes course and run in the path being produced

pwilkins:
I ran an export of 8.01CC in Docker.  If the goal is to reproduce the contents of the static folder (it's unclear from the issue description) then there still are some items missing.  

The course contains a directory at `static/textbooksummary` that contains four files.  It also contains a file `static/Syllabus.pdf`.  None of this content is in the export.  I can supply the course tarball if you want to try it.

pwilkins:
I discussed this with @noisecapella and it turns out that I was looking for the wrong result.  I had expected that exporting just the `course` learningresource would export all the static assets.  That's incorrect.  It exports just those static assets associated with the `course` learningresource.  This is what it is doing, reproducing the directory structure.

So, it works.

exporter/api.py
```diff
@@ -13,6 +14,25 @@
...
+    while os.path.exists(path):
+        collision = True
+        path = "{base}_{i}{ext}".format(
+            base=base,
+            ext=ext,
+            i=i
+        )
+        i += 1
+
+    return path, collision
```
pwilkins, line 35: Just a question: Why is `collision`  tracked?  It seems to be only used in the tests, and there it is asserted to be `True`.

noisecapella, line 35: The collision value is exposed in the REST API where it's used to alert the user that one of their static assets was renamed. See [this line](https://github.com/mitodl/lore/blob/master/ui%2Fstatic%2Fui%2Fjs%2Flr_exports.jsx#L116)

PR #623, opened 2015-09-09
Added sorting by "title"

### LINKED ISSUE #606: sort alphabetically by learning resource name
As a curator, sometimes I just want to sort alphabetically. 

Perhaps this should be the default sort

giocalitri:
Handles also the ascending sorting
Fixed unit tests

Closes #606

### DIFF
--- a/rest/tests/test_search.py
+++ b/rest/tests/test_search.py
@@ -181,16 +181,19 @@ class TestSearch(RESTTestCase):
         resource1.xa_avg_grade = 2.0
         resource1.xa_nr_attempts = 4
         resource1.xa_nr_views = 1000
+        resource1.title = '22222'
         resource1.save()
 
         resource2.xa_avg_grade = 4.0
         resource2.xa_nr_attempts = 1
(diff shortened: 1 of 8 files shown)


giocalitri:
@bdero This needs reindex after has been merged and after release

rest/tests/test_search.py
```diff
@@ -181,16 +181,19 @@ def test_sortby(self):
         resource1.xa_avg_grade = 2.0
         resource1.xa_nr_attempts = 4
         resource1.xa_nr_views = 1000
+        resource1.title = '00000'
         resource1.save()
 
         resource2.xa_avg_grade = 4.0
         resource2.xa_nr_attempts = 1
         resource2.xa_nr_views = 100
+        resource2.title = '11111'
```
noisecapella, line 190: Can you swap the titles such that they aren't in the same order as the resource id? That way we can confirm that it's the title being sorted and not the resource id

giocalitri:
Swapped the titles

noisecapella:
:+1:

PR #624, opened 2015-09-09
Remove jQuery selectors, add hasLabel state and ref attribute

noisecapella:
This changes a few things:
- `formatActionState` now represents the state of the component at the moment rather than the next action. If set to `show` the component will show the term, if set to `edit` the term will be shown in a textbox for editing.
- `edit-term-box-hide` is removed, `edit-term-box-show` renamed to `edit-term-box`.
- Pieces of the state like `formatActionClassName`, `editTextClass`, and `listClassName` are removed from state and calculated in `render`. See (the React docs)[https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html]: "Whenever possible, compute values on-the-fly to ensure that they don't get out of sync later on and cause maintenance trouble."
- `labelSelector.hide()` and `.show` removed, `render` now renders the element depending on the value of `formatActionState`
- `ref` tag added to textbox, `focus()` now uses the `ref` tag to find the DOM node to focus.
- data attributes removed because they aren't necessary anymore

### DIFF
--- a/ui/jstests/test_manage_taxonomies.jsx
+++ b/ui/jstests/test_manage_taxonomies.jsx
@@ -217,7 +217,7 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
         var termLabel = labels[0];
         var label = termLabel.getDOMNode();
         assert.equal(label.innerHTML, term.label);
-        assert.equal(component.state.formatActionState, 'edit');
+        assert.equal(component.state.formatActionState, 'show');
 
         var formatButton = React.addons.TestUtils.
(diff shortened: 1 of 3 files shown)


noisecapella:
@amir-qayyum-khan please look over the changes and merge them into your PR if they look good to you.

amir-qayyum-khan:
Will you fix the Jshint issue or should i fix?

amir-qayyum-khan:
let me fix issue after merge on my side. To windup these PR ASP. Thanks

PR #626, opened 2015-09-09
Added Roles module docstrings to documentation

pwilkins:
Added Roles to docs/api.rst to include it in documentation.  Also
updated docstrings in files that generated Sphinx errors during build.

@pdpinch @noisecapella


noisecapella:
Looks good to me if tests pass

PR #629, opened 2015-09-10
Added capability to facet by missing Vocabulary terms in REST API search

giocalitri:
First part: backend only 

The facet for this must look like:
/api/v1/repositories/small/search/?selected_facets=_missing_:3_exact

### DIFF
--- a/rest/views.py
+++ b/rest/views.py
@@ -26,6 +26,7 @@ from celery.states import FAILURE, SUCCESS, REVOKED
 from celery.result import AsyncResult
 from haystack.inputs import Exact
 from haystack.query import SearchQuerySet
+from statsd.defaults.django import statsd
 
 from exporter.tasks import export_resources
 from roles.permissions import GroupTypes, BaseGroupTypes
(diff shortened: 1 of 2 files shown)


giocalitri:
Need to write a couple of tests

giocalitri:
Ready to be reviewed

rest/views.py
```diff
@@ -819,7 +838,17 @@ def make_facet_counts(self, queryset):
...
         facet_counts = queryset.facet_counts()
-        ret = get_vocabularies(facet_counts)
+        vocabs = get_vocabularies(facet_counts)
+
+        # return dictionary
+        ret_dict = {}
+
+        # process vocabularies
+        for key, values in vocabs.items():
+            missing_count = queryset.filter(
```
noisecapella, line 849: Is there any potential performance impact from this? (cc @ShawnMilo) It seems like it would execute a query for every vocabulary, not sure if that's correct or if it's anything significant

ShawnMilo, line 849: Right now, it has to do a query per vocabulary. I don't know what the performance effect is (have you timed it, @giocalitri?). In any case, this needs to be in before the Elasticsearch-DSL work is done, so it's only temporary.

noisecapella, line 849: Ok, I'm fine just merging it then

noisecapella:
:+1:

giocalitri:
Actually I have been asked to plug statsd after I add these queries, so will do and update the PR

giocalitri:
plugged statsd

rest/views.py
giocalitri, line 849: I did not measure the difference in performance, but right now we don't have another choice, so unless the website becomes unusable, I guess this is the way to go.

noisecapella:
Functionality works, feel free to merge

PR #637, opened 2015-09-10
Fixed bug with sorting by title being case sensitive

### LINKED ISSUE #631: Sorting by title does not work properly
It seems to be case sensitive

giocalitri:
Fixes #631

### DIFF
--- a/search/search_indexes.py
+++ b/search/search_indexes.py
@@ -85,9 +85,8 @@ class LearningResourceIndex(indexes.SearchIndex, indexes.Indexable):
     avg_grade = indexes.FloatField(model_attr="xa_avg_grade")
 
     lid = indexes.IntegerField(model_attr="id", indexed=False)
-    # title set in this way because of a known bug in haystack
-    # http://bit.ly/1ENeElY
-    title = indexes.CharField(model_attr="title", indexed=False, stored=True)
+    title = indexes.CharField(model_attr="title", indexed=False)
+    titlesort = indexes.CharField(indexed=False)
     description = indexes.CharField(model_attr="description", indexed=False)
     description_path = indexes.CharField(
         model_attr="description_path",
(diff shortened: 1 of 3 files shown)


search/search_indexes.py
```diff
@@ -137,6 +136,15 @@ def prepare_preview_url(self, obj):  # pylint: disable=no-self-use
             run=data["run"],
         )
 
+    def prepare_titlesort(self, obj):  # pylint: disable=no-self-use
+        """Define what goes into the "titlesort" index."""
+        title = obj.title.strip()
+        # hack to handle empty titles
+        # to show up at the bottom of the sorted list
+        if not title:
+            title = 'z' * 30
```
noisecapella: Why not prepend a '0' to all titles and then make the title a '1' if it's empty? That guarantees it's at the bottom. We can't rely on ASCII comparisons like this, what if the title starts with a non alphanumeric character, or something unicode?

noisecapella: To clarify, I mean relying on 'z' having a high value relative to everything else. It's reasonable to expect '0' to be before '1' and 'a' to be before 'z'

search/sorting.py
```diff
@@ -13,7 +13,7 @@ class LoreSortingFields(object):
     SORT_BY_NR_VIEWS = ('nr_views', 'Number of Views (desc)', '-')
     SORT_BY_NR_ATTEMPTS = ('nr_attempts', 'Number of Attempts (desc)', '-')
     SORT_BY_AVG_GRADE = ('avg_grade', 'Average Grade (desc)', '-')
-    SORT_BY_TITLE = ('title', 'Title (asc)', '')
+    SORT_BY_TITLE = ('titlesort', 'Title (asc)', '')
```
noisecapella, line 16: nit: It seems a bit odd to have `?sortby=titlesort` in the URL, is there some way to use `title` in the URL but use the `titlesort` field in the index? Feel free to ignore if this complicates things unnecessarily

search/search_indexes.py
giocalitri: I agree with your solution: will do

search/sorting.py
giocalitri, line 16: yes, but it means changing many more things:
what about a different string instead of `titlesort`? Any idea?

noisecapella, line 16: If we can't use `title`, `titlesort` is fine with me

giocalitri:
Updated the `prepare_titlesort` function

noisecapella:
:+1: after tests pass

PR #639, opened 2015-09-11
Use xbundle v. 0.3.1, which fixes a bug that caused incomplete import

jamiefolsom:
Connected to mitodl/xbundle#43 and mitodl/xbundle#44

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ git+https://github.com/mitodl/django-guardian@de7ca5c39dfd0870067497f5843581e9cd
 git+https://github.com/mingchen/django-cas-ng@69463fe7c23f025be7165c2967f4ceb983d60472#egg=django-cas-ng==3.4.3-odl
 
 # Importer application requirements
-xbundle==0.3.0
+git+https://github.com/mitodl/xbundle@0e6d87b6d87bf0c366b3e87865df2327d57c6525#egg=xbundle
 git+https://github.com/gdub/python-archive.git@61b9afde21621a8acce964d3336a7fb5d2d07ca6#egg=python-archive==0.3.1-odl
 django-storages-redux==1.3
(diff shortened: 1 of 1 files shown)


bdero:
Please do `egg=xbundle==0.3.1`, to ensure that pip identifies this as a new version of xbundle. Other than that, :+1:

jamiefolsom:
@bdero Done. Thanks!

requirements.txt
```diff
@@ -28,7 +28,7 @@ git+https://github.com/mitodl/django-guardian@de7ca5c39dfd0870067497f5843581e9cd
 git+https://github.com/mingchen/django-cas-ng@69463fe7c23f025be7165c2967f4ceb983d60472#egg=django-cas-ng==3.4.3-odl
 
 # Importer application requirements
-xbundle==0.3.0
+git+https://github.com/mitodl/xbundle@0e6d87b6d87bf0c366b3e87865df2327d57c6525#egg=xbundle==0.3.1
```
pdpinch, line 31: Since this is not (yet) released to pyPi, do we need some additional text for the version, like the -odl used elsewhere?

jamiefolsom, line 31: My understanding is that that's just an alias for the specified commit hash, and since the repo is an mitodl repo, that additional modifier might be unneccessary?

bdero:
@pdpinch Since that commit includes release notes for 0.3.1, I assumed that commit _will_ be version 0.3.1 on the cheeseshop shortly. If we're not going to release soon and there are other XBundle changes down the pipe, we definitely should append `-odl` or something else to the end as you suggest.

pdpinch:
Thanks Brandon. I think we'll release soon-ish, and it will be before any additional changes, so let's go with this.

pdpinch:
Good to merge @jamiefolsom

PR #643, opened 2015-09-11
Cleared message box on changes to the learning resource panel

### LINKED ISSUE #609: Remove save confirmation when learning resource panel is updated
As a curator, I find it confusing when I:
1. make a change
2. save my change
3. make another change, and
the save confirmation message is still in the same location? Was my change saved?

The confirmation messages should be removed when more changes are made to the learning resource panel.

noisecapella:
Fixes #609

### DIFF
--- a/ui/jstests/test-learning-resource.jsx
+++ b/ui/jstests/test-learning-resource.jsx
@@ -488,6 +488,67 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
     }
   );
 
+  QUnit.test('If terms or description are changed, clear error messages',
+    function (assert) {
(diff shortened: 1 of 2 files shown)


pdpinch:
functionality is good. 

no tests? It's ok, but I wanted to check.

noisecapella:
That's an oversight, I'll add a test

noisecapella:
Ok, test is in place

giocalitri:
~~I just verified that this PR introduced the bug #725~~

giocalitri:
I take it back: the weird behavior of #725 is caused by the `firebug` plugin.

This PR can be merged

Issue #507, opened 2015-08-13
Deleting a term in taxonomy panel


amir-qayyum-khan:
##### Requirements
- [ ] deleting a term should remove it from all learning resources

Epic https://github.com/mitodl/lore/issues/380
##### Details

On clicking delete term, I will show a confirmation dialog. On confirming I will remove that term and re-render list.

cc @pdpinch

amir-qayyum-khan:
@noisecapella are u working on this task?

noisecapella:
No, feel free to work on it

amir-qayyum-khan:
ok, thanks

pdpinch:
See also https://odl.zendesk.com/agent/tickets/2513

amir-qayyum-khan:
@pdpinch i dont have access to zendesk.

pdpinch:
Sorry about that. It was mostly for my reference. Here's the user's feedback:

> The "X" button in the taxonomy view doesn't allow you to delete an individual taxonomy item (not the entire vocabulary, but just a single item inside the vocabulary).

pdpinch:
closed via #648

PR #648, opened 2015-09-14
Delete term(s) in taxonomy panel

amir-qayyum-khan:
Hi

In this PR i added code to delete term(s) of a vocabulary.

Issue https://github.com/mitodl/lore/issues/507

@pdpinch @noisecapella

### DIFF
--- a/ui/jstests/test_manage_taxonomies.jsx
+++ b/ui/jstests/test_manage_taxonomies.jsx
@@ -181,7 +181,7 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
     }
   });
 
-  QUnit.test('Assert that TermComponent renders properly',
+  QUnit.test('Assert that edit TermComponent renders properly',
     function(assert) {
       assert.ok(TermComponent, "class object not found");
       var done = assert.async();
@@ -475,7 +475,9 @@ define(['QUnit', 'jquery', 'lodash', 'manage_taxonomies', 'react',
               );
               component.forceUpdate(function() {
                 assert.equal(component.state.label, "TestB");
-
+                assert.equal(component.state.formatActionState, 'edit');
+                assert.equal(parentUpdateCount, 0);
+                // listing page was not asked to refresh
                 assert.equal(refreshCount, 0);
                 React.addons.TestUtils.Simulate.click(saveButton);
                 component.forceUpdate(function() {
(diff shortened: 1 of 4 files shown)


ui/views.py
```diff
@@ -172,7 +172,9 @@ def get_vocabularies(facets):
         vocab = (vocabulary_id, vocabs[int(vocabulary_id)])
         vocabularies[vocab] = []
         for t_id, count in term_data:
```
amir-qayyum-khan, line 173: On deleting a term, that particular term get delete from **Term model**. But not from **term_data**. Can any one help me understanding this. I have added a check #L176

> `if t_id_int in terms:`

Can any one help me understanding @ShawnMilo @pdpinch

noisecapella, line 173: `term_data` comes from elasticsearch. I'm thinking that the delete completed before elasticsearch had a chance to be updated. Can you confirm @ShawnMilo?

I think you should leave this piece of code alone because there are better places to fix this, maybe by waiting for elasticsearch at the point of the API call.

ui/jstests/test_manage_taxonomies.jsx
```diff
@@ -465,6 +465,137 @@ define(['QUnit', 'jquery', 'manage_taxonomies', 'react',
...
+        );
+        //select delete
+        React.addons.TestUtils.Simulate.click(deleteButton);
+        component.forceUpdate(function() {
+          assert.equal(
+            component.state.errorMessage, ''
+          );
+          assert.equal(component.state.formatActionState, 'show');
+          waitForAjax(1, function() {
+            // check sate of error message
```
noisecapella: `sate` should be `state`

ui/views.py
ShawnMilo, line 173: @amir-qayyum-khan @noisecapella 

It _sounds_ like a race condition. The `Term` is deleted, but Elasticsearch isn't yet updated. That suggests one of two things:
1. Simple race condition.
2. Deleting a term doesn't trigger index updates for the `LearningResource`s. 

If it's the first, waiting a few seconds and trying again should "fix" it. If it's the second, then some work would need to be done to fix it. However, that work should be put into the new `elasticsearch-dsl` code, and not here.

noisecapella, line 173: Actually it looks like we don't update the index at all for term deletes

ShawnMilo, line 173: If it's not automatically happening in the `m2m_updated` signal then it's not happening. I just don't know if deleting it, which does remove the m2m relation, triggers that signal.

noisecapella, line 173: If we need to wait for the elasticsearch DSL work to update the index then we should hold off on this PR until that work is done.

pdpinch:
Last I heard, this was blocked by elastic search DSL. Since that's merged, can work on this start again?

@ShawnMilo @noisecapella 

We've had a user request for this, https://odl.zendesk.com/agent/tickets/2513

noisecapella:
I created and merged #746 to add that functionality and fix another issue I saw. I'll merge this once tests pass and PR #739  is merged

amir-qayyum-khan:
@noisecapella  sounds good to me. :+1:

noisecapella:
Thanks @amir-qayyum-khan I'll merge when tests pass

pdpinch:
closes #507

PR #650, opened 2015-09-14
Fixed clear export bug

### LINKED ISSUE #644: Export button doesn't clear after export
This is a regression due to it being overlooked in recent React work. It should be fairly straightforward to solve

noisecapella:
Fixes #644


pdpinch:
LGTM. I tested it locally and it functions as well. Merge away 

(something weird going on with coverage)

PR #656, opened 2015-09-14
Increased requirejs timeout

### LINKED ISSUE #463: Javascript timeouts
This is something that's also present on the UI mockup to a lesser degree. The page stalls for a moment then all the javascript for the listing page is executed. You can see this in the checkboxes which become iCheck checkboxes after a noticeable lag.

I've also occasionally gotten this from lore-ci when the delay is so long that it quits:

```
Error: Load timeout for modules: jquery,lodash,reactaddons,select2,bootstrap,icheck
http://requirejs.org/docs/errors.html#timeout
```

noisecapella:
Fixes #463 

This is more of a workaround than a solution but it should at least help user experience a little

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -113,7 +113,8 @@
       require.config({
         baseUrl: "{% static "bower"%}",
         shim: SHIMS,
-        paths: REQUIRE_PATHS
+        paths: REQUIRE_PATHS,
+        waitSeconds: 200
       });
       require(["jquery", "bootstrap"]);
     </script>


ShawnMilo:
Is there an issue or anything for what this works around?

noisecapella:
The issue is #463. Basically, requirejs waits for all the javascript dependencies to be obtained and set up before executing any code. If a few seconds pass without all dependencies being fulfilled it gives up and errors on the javascript console. To the user the page just doesn't work randomly which is not good. Increasing the timeout would help the page load eventually.

A better solution is to get our javascript compression working so that all our dependencies come in the same file, but that's another issue.

ShawnMilo:
Okay, fair enough. Please open an issue for the compression work we need to do, if we don't already have one.

ShawnMilo:
:+1:

PR #657, opened 2015-09-14
Hotfix/0.10.1

pwilkins:


bdero:
Since 0.10.0 didn't make it to production, we didn't complete that release. This should probably be release candidate # 2 of the 0.10.0 release, not a new bugfix release 0.10.1.

bdero:
Not that it's a huge deal, but maybe something to note for next time. @pwilkins

pwilkins:
@bdero What may have made this more difficult is that I already tagged this as 0.10.0.  I suppose I could have edited the tag, or deleted and re-created it.

Issue #613, opened 2015-09-08
DataTable.js


pdpinch:
As a user, I'd like a way to view and sort by terms and other fields on assets, without having to open the learning resource panel. 

We've had many requests to use DataTable.js, possibly as an alternate "curator" view of the learning resource listing page.

jamiefolsom:
I heard this request first hand last week! After a little research, it appears there is some concern that datatable.js is too closely bound to the DOM to be a good fit with React. There are some alternatives however. See: [Griddle](http://griddlegriddle.github.io/Griddle/) and [React-Datagrid](http://zippyui.com/react-datagrid/)

pdpinch:
MVP: alternate view of repository with static view of data

PR #659, opened 2015-09-14
Implemented page_size parameter to allow users to set page size

noisecapella:
Required for #613 

Allows us to optionally set a `page_size` parameter which lets us gulp down data more efficiently

### DIFF
--- a/rest/pagination.py
+++ b/rest/pagination.py
@@ -1,0 +1,16 @@
+"""
+Pagination classes for Lore's REST API.
+"""
+
+from __future__ import unicode_literals
+
+from rest_framework.pagination import PageNumberPagination
+
+
+class LorePagination(PageNumberPagination):
+    """
+    Pagination class for Lore's REST API.
+    """
+    page_size = 20
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
--- a/rest/tests/test_misc.py
+++ b/rest/tests/test_misc.py
@@ -4,19 +4,47 @@ Unit tests for REST api
 from __future__ import unicode_literals
 
 from rest_framework.status import (
+    HTTP_200_OK,
     HTTP_404_NOT_FOUND,
 )
 
+from learningresources.models import LearningResourceType
 from rest.tests.base import (
     RESTTestCase,
     API_BASE,
+    as_json,
 )
+from rest.pagination import LorePagination
 
 
 class TestMisc(RESTTestCase):
     """
     REST test
     """
+
+    def assert_page_size(self, page_size, expected_num_results):
+        """
+        Helper function to assert len(results) == expected_num_results.
+        """
+        num_types = LearningResourceType.objects.filter().count()
+
+        if page_size is None:
+            page_size_param = ""
+        else:
+            page_size_param = "?{query_param}={page_size}".format(
+                query_param=LorePagination.page_size_query_param,
+                page_size=page_size
+            )
+        resp = self.client.get(
+            "{api_base}learning_resource_types/{page_size_param}".format(
+                api_base=API_BASE,
+                page_size_param=page_size_param
+            )
+        )
+        self.assertEqual(resp.status_code, HTTP_200_OK)
+        resources = as_json(resp)
+        self.assertEqual(len(resources['results']), expected_num_results)
+        self.assertEqual(resources['count'], num_types)
 
     def test_root(self):
         """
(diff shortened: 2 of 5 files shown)


rest/pagination.py
```diff
@@ -0,0 +1,16 @@
...
+from rest_framework.pagination import PageNumberPagination
+
+
+class LorePagination(PageNumberPagination):
+    """
+    Pagination class for Lore's REST API.
+    """
+    page_size = 20
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
```
giocalitri, line 16: where are these numbers coming from?

noisecapella, line 16: 20 is the default page size, 1000 is just some number I considered to be big enough for one request, a guess

giocalitri, line 16: is 20 the same number in the configuration? if so you should take it from there

rest/tests/test_misc.py
```diff
@@ -48,3 +53,41 @@ def test_root(self):
...
+        """
+        Test page_size query parameter.
+        """
+        self.import_course_tarball(self.repo)
+
+        num_resources = LearningResource.objects.filter(
+            course__repository__id=self.repo.id
+        ).count()
+
+        def assert_page_size(page_size):
```
giocalitri: nit: I would move this function at the beginning, after the doctoring of the parent function

rest/pagination.py
noisecapella, line 16: Sorry, it's not the default, it was a value we added. It's removed from lore/settings.py above so this is the only place it exists now.

giocalitri, line 16: sorry, now I understand the first file (that I completely skipped)

rest/tests/test_misc.py
```diff
@@ -48,3 +53,41 @@ def test_root(self):
...
+                )
+            )
+            self.assertEqual(resp.status_code, HTTP_200_OK)
+            resources = as_json(resp)
+            self.assertTrue(len(resources['results']) <= page_size)
+            self.assertEqual(resources['count'], num_resources)
+
+        assert_page_size(2)
+        assert_page_size(10)
+        assert_page_size(None)
```
giocalitri: what happens with `assert_page_size(LorePagination.max_page_size +1)` ?

giocalitri: what about invalid page numbers?

ui/tests/test_repository_views.py
```diff
@@ -0,0 +1,102 @@
...
+        Test that page size context is set properly.
+        """
+        resp = self.client.get(
+            "/repositories/{slug}/".format(slug=self.repo.slug))
+        self.assertEqual(json.loads(resp.context['page_size_json']), 20)
+
+        # Invalid params are ignored.
+        resp = self.client.get(
+            "/repositories/{slug}/?page_size=xyz".format(slug=self.repo.slug))
+        self.assertEqual(json.loads(resp.context['page_size_json']), 20)
```
giocalitri: ok, fount the answer to my previous question (invalid page number)

ui/tests/test_repository_views.py
```diff
@@ -0,0 +1,102 @@
...
+
+        resp = self.client.get(
+            "/repositories/{slug}/?page_size=300".format(slug=self.repo.slug))
+        self.assertEqual(json.loads(resp.context['page_size_json']), 300)
```
giocalitri: I guess you need an assert for `page_size` > max allowed

rest/tests/test_misc.py
noisecapella: I'll add this, it works fine locally

rest/tests/test_misc.py
noisecapella: Will do

ui/tests/test_repository_views.py
```diff
@@ -0,0 +1,102 @@
...
+            'manage_repo_users',
+            'add_edit_metadata',
+            'manage_taxonomy',
+            'view_repo',
+            'import_course'
+        ]
+        self.assertEqual(
+            sorted(resp.context["perms_on_cur_repo"]),
+            sorted(expected_perms)
+        )
```
giocalitri, line 91: why do you need this test?

giocalitri, line 91: sorry, now I understand, it's for the context

giocalitri:
This is good to go as soon as you squash and tests pass

Issue #604, opened 2015-09-08
Use elipsis (...) instead of MISSING in learning resource path when display_name isn't set


pdpinch:
As a user, the repeated MISSING that appears in learning resource paths in the list view is distracting. Can we use something like `...` instead?

![screenshot-lore-ci herokuapp com 2015-09-08 11-11-08](https://cloud.githubusercontent.com/assets/430126/9738638/6ea0de2e-561a-11e5-99bc-4a82e1ec9987.png)

we only want to change the path, not the learning resource name. It would be best to change the existing path data, as well as any new path data.

PR #664, opened 2015-09-15
Replaced MISSING with elipsis in learning resource description_path

amir-qayyum-khan:
In this PR I replaced **MISSING** to  **elipsis(...)**. 
**MISSING** was used as default value for  description_path if display_name was not set to
##### Only applicable to new courses we import.

issue https://github.com/mitodl/lore/issues/604

@pdpinch @noisecapella


pdpinch:
Is there a management command for updating existing data? Maybe @giocalitri or @ShawnMilo knows.

importer/api/__init__.py
```diff
@@ -148,7 +148,7 @@ def import_children(course, element, parent, parent_dpath):
         None
     """
     # pylint: disable=too-many-locals
-    title = element.attrib.get("display_name", "MISSING")
+    title = element.attrib.get("display_name", "...")
```
giocalitri, line 151: I'm pretty sure that is not a `…` but `...`

giocalitri:
@pdpinch Yes, the command is `python manage.py update_description_paths`

amir-qayyum-khan:
I am not familiar with that. But there must be.
Is there any doc to understand which database we are using for LORE. I mean which DBMS, set of docker commands to access db and any software reset of team is using to view database manually?

@pdpinch @ShawnMilo @noisecapella

giocalitri:
The problem is that this will not work with existing learning resources already imported: we need a migration to update the existing learning resources

importer/api/__init__.py
amir-qayyum-khan, line 151: @giocalitri  what do u mean by that? spaces in each dot ?

giocalitri, line 151: no, I mean the ellipsis you put is a sequence of 3 dots and not the unicode ellipsis character U+2026

pdpinch, line 151: Good point. Do you think … (U+2026) is safe for this case?

giocalitri, line 151: It should be: in theory all the code we wrote should be unicode safe: if it's not, we have a bug that we have to address, so maybe it's a good idea to use this case to verify everything.

noisecapella:
The migration should also handle the `update_description_paths` task, probably by calling whatever function the management command uses.

amir-qayyum-khan:
Should I create the migration in this PR ?
@noisecapella @giocalitri  @pdpinch

pdpinch:
yes please

giocalitri:
@amir-qayyum-khan there has been another migration to update the description paths: just use the same code.

amir-qayyum-khan:
ok, i will do it. thanks

giocalitri:
just FYI: you may need to force the update of the description path: I'm pretty sure the function I wrote that can already do it.

importer/api/__init__.py
amir-qayyum-khan, line 151: @giocalitri @pdpinch 
Getting issues using unicode ellipsis
- unable to create LearningResource in test case `<exporter.tests.test_export.TestExport testMethod=test_export_resources_to_directory>`
  https://gist.github.com/amir-qayyum-khan/dc97cac3a38326a48450
- haystack issue
  https://gist.github.com/amir-qayyum-khan/5fba27592c8020d5f7f8

#### The issue is here

https://github.com/mitodl/lore/blob/fix/aq/fix_display_name_lr/search/search_indexes.py#L121
if I change it to : 
Ref http://stackoverflow.com/questions/3235386/python-using-format-on-a-unicode-escaped-string

```
        return u"{0} {1} {2}".format(
            obj.title.decode('utf-8'),
            obj.description.decode('utf-8'),
            content
        )
```

then issue appears here 

> File "/usr/local/lib/python2.7/dist-packages/haystack/fields.py", line 165, in convert
>     return six.text_type(value)

what you guys suggest ?

giocalitri, line 151: I do not understand why 
https://github.com/mitodl/lore/blob/fix/aq/fix_display_name_lr/search/search_indexes.py#L121
gives a problem.
In theory having `from __future__ import unicode_literals` at the beginning means all the strings are in unicode.
Can you make sure that all the modules where `title` is used make that import?

did you try to use the string `"…"` directly instead of `"MISSING"` or did you use the unicode code?

amir-qayyum-khan, line 151: I was using unicode, But I revert to "...". As per our discussion on hangout today. I will open an issue for haystack is not accepting unicode character.

giocalitri:
I partially re-wrote the code in #707: closing this PR

PR #665, opened 2015-09-15
Removed react-addons bower package, addons actually live in react package

noisecapella:


ShawnMilo:
Code deleted, tests passing. Looks like fine work to me.

PR #666, opened 2015-09-15
Switched to minimized JS for libraries

noisecapella:
This should make our javascript code run slightly faster.


pdpinch:
Does django-compressor not minimize these already?

ShawnMilo:
@pdpinch: Django-compressor does have some code that looks like it can minify JavaScript. I don't know if we have to tweak any settings or something to take advantage of it. However, assuming we don't plan on modifying these third-party libraries, I like the idea of using the official minified versions.

Looks straightforward to me.

noisecapella:
Unfortunately django-compressor does not minimize these other libraries, we only compress our own javascript code. I think it was the intention to at one point and we should put some effort in setting up `RequireJSCompressor` to work properly, but that's another issue.

PR #667, opened 2015-09-15
Added datatable proof of concept

### LINKED ISSUE #613: DataTable.js

noisecapella:
Fixes #613


noisecapella:
This is ready for review once #659 is merged

pwilkins:
Increasing the page size using the `Page Size` widget to 200 at the bottom of the datatable works, but choosing a smaller size once a larger size has been selected leaves the datatable empty.

pwilkins:
Consider adding the breadcrumb text as a datatable column, for example, `Advanced Introductory Classical Mechanics / 1: Newton's Laws of Motion / Newton's First Law / Inertial and Non Inertial Frame of References`.  It would be a useful sort field.

noisecapella:
@pwilkins That should be available under Description Path. You will probably have to resize the column, you can do that by clicking on the right edge in the header and dragging it.

noisecapella:
I can't reproduce the `Page Size` problem you mentioned. Can you note what the current page size was at each point?

pwilkins:
Steps to reproduce `Page Size` problem: 
Starting with an initial page size of 20, 
Change the page size to 200,
Scroll to the bottom and change the page size back to 20.

This occurs in FireFox.

noisecapella:
I see the problem now, I'll look into it

ui/static/ui/js/lore_data_table.jsx
```diff
@@ -0,0 +1,197 @@
...
+          name: "title",
+          title: "Title",
+        },
+        {
+          name: "description",
+          title: "Description"
+        },
+        {
+          name: "description_path",
+          title: "Description Path",
```
pdpinch: Can we label this "Path" or "Learning Resource Path" if you don't think that's clear enough?

"Description Path" doesn't make sense to me.

ui/static/ui/js/lore_data_table.jsx
```diff
@@ -0,0 +1,197 @@
...
+          name: "course",
+          title: "Course"
+        },
+        {
+          name: "run",
+          title: "Run"
+        },
+        {
+          name: "xa_nr_views",
+          title: "Number of Views",
```
pdpinch: Shorten to "Views"

ui/static/ui/js/lore_data_table.jsx
```diff
@@ -0,0 +1,197 @@
...
+          type: "number"
+        },
+        {
+          name: "xa_nr_attempts",
+          title: "Number of Attempts",
```
pdpinch: shorten to "Attempts"

ui/static/ui/js/require_config.js
```diff
@@ -5,8 +5,9 @@ var REQUIRE_PATHS = {
   bootstrap: "bootstrap/dist/js/bootstrap.min",
   icheck: "icheck/icheck.min",
   retina: "retina.js/dist/retina.min",
-  react: "react/react-with-addons.min",
+  React: "react/react-with-addons",
```
pdpinch: Why did you revert the minification?

noisecapella: That was unintentional, I'll change it back

pdpinch:
Just my 3 comments. Otherwise this looks good to me as a proof of concept. 

@Ferdi ?

pwilkins:
@noisecapella I didn't immediately recognize that `Description Path` was the same as what I was calling "breadcrumbs".  Now that I look at it, it's not as useful in practice as I had imagined.

noisecapella:
@pwilkins I filed an issue for the scrolling problem, given that it's a proof of concept I won't hold up the merge for that though

pwilkins:
@noisecapella I agree.  This shouldn't delay the merge.

Ferdi:
I had the impression we would specifically use datatable.js library.   Do we know how these 2 libraries (react-datagrid vs datatablejs) compare in terms of features?  My only concern is that the users who requested these feature might expect some specific features from datatablejs.

jamiefolsom:
Hi @Ferdi, 

My understanding (@noisecapella, correct me if I'm wrong) is that datatable.js is one of several libraries that offer similar functionality, and is more mature than the others. 

However, while it's possible to use Datatable,js alongside (and separate from) react components, it's not possible to use it within a react component without losing the features that make it better than the alternatives, since it modifies the DOM directly, which is in conflict with the react approach.

PR #671, opened 2015-09-16
(WIP) Added collectstatic step for builds and deploys

bdero:


pdpinch:
Is there an issue that goes with this? 

Does this have to do with @noisecapella's work on #613?

And is it actually ready for review, or did waffle add that label automatically?

noisecapella:
This was thought to be blocking that issue but it turns out to be a file I forgot to commit instead, see #672

PR #672, opened 2015-09-16
Removed lib/ from gitignore

noisecapella:
This hides files inside all `lib` directories, it seems like the intention was to just hide certain build files. We can add those back on an as needed basis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ develop-eggs/
 /dist/
 downloads/
 eggs/
-lib/
 lib64/
 parts/
 sdist/


ShawnMilo:
This was added (along with dozens of other lines) by default by Carson. So it doesn't have an explicitly-stated purpose, and clearly came from a template.

@bdero: Do you have any reason why this shouldn't be merged?

pdpinch:
It's part of the [standard github .gitgnore for python](https://github.com/github/gitignore/blob/master/Python.gitignore#L18)

But if it's causing trouble, might as well remove it. Easy enough to add back in when necessary.

pdpinch:
:+1:

ShawnMilo:
@noisecapella: What trouble is it causing? Are you adding a new `lib` folder in a branch? I removed `lib/` locally and it's not showing any new untracked files, so it looks like it's fine to me.

noisecapella:
In the datatable branch I added `ui/static/lib` for static files which weren't part of a bower package.

PR #673, opened 2015-09-16
Adds elasticsearch-dsl and adds it alongside Haystack for now.

ShawnMilo:
Haystack will removed once all the features are added and tested. This requires
the addition of searching by facets, React.js updates, and updates to the
restful API.

Before code review, please watch [this](https://youtu.be/uU66q-8Zofg), which explains how the new libraries are being used.

The `search/utils.py` file should be read first, in this order:
-  `create_mapping`
- `resource_to_dict`
- `index_resources`
- `search_index`
- `class SearchResults`
- the rest

This should make the tests easy to follow.

### DIFF
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -25,8 +25,10 @@ from learningresources.api import (
     create_static_asset,
     get_video_sub,
     join_description_paths,
+    get_resources,
 )
 from learningresources.models import StaticAsset, course_asset_basepath
+from search.utils import index_resources
 
 log = logging.getLogger(__name__)
 
@@ -119,7 +121,6 @@ def import_course(bundle, repo_id, user_id, static_dir):
         learningresources.models.Course
     """
     src = bundle.course
-
     course = create_course(
         org=src.attrib["org"],
         repo_id=repo_id,
@@ -130,6 +131,10 @@ def import_course(bundle, repo_id, user_id, static_dir):
     import_static_assets(course, static_dir)
     import_children(course, src, None, '')
     populate_xanalytics_fields.delay(course.id)
+    # This triggers a bulk indexing of all LearningResource instances
+    # for the course at once.
+    index_resources(
+        get_resources(repo_id).filter(course__id=course.id))
     return course
 
 
--- a/search/utils.py
+++ b/search/utils.py
@@ -1,0 +1,387 @@
+"""
+Search functions.
+"""
+
+from __future__ import unicode_literals
+
+from collections import defaultdict
+from lxml import etree
+import logging
+
+from django.conf import settings
+from elasticsearch.helpers import bulk
+from elasticsearch.exceptions import NotFoundError
+from elasticsearch_dsl.connections import connections
+from elasticsearch_dsl import Search, Mapping, query
+
+from statsd.defaults.django import statsd
+
+from learningresources.models import get_preview_url, LearningResource
+from search.search_indexes import get_course_metadata
+
+log = logging.getLogger(__name__)
+
+DOC_TYPE = "learningresource"
+INDEX_NAME = settings.HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"]
+URL = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
+_CONN = None
+PAGE_LENGTH = 10
+
+
+def get_conn():
+    """
+    Lazily create the connection.
+    """
+    # pylint: disable=global-statement
+    # This is ugly. Any suggestions on a way that doesn't require "global"?
+    global _CONN
+    if _CONN is None:
+        _CONN = connections.create_connection(hosts=[URL])
+    return _CONN
+
+
+def get_resource_terms(resource_ids):
+    """
+    Returns taxonomy metadata for LearningResources.
+    Args:
+        resource_ids (iterable of int): Primary keys of LearningResources
+    Returns:
+        data (dict): Vocab/term data for course.
+    """
+    resource_data = defaultdict(lambda: defaultdict(list))
+    rels = LearningResource.terms.related.through.objects.select_related(
+        "term__vocabulary").filter(learningresource__id__in=resource_ids)
+    for rel in rels.iterator():
+        obj = resource_data[rel.learningresource_id]
+        obj[rel.term.vocabulary.name].append(rel.term.label)
+    # Replace the defaultdicts with dicts.
+    return {k: dict(v) for k, v in resource_data.items()}
+
+
+def search_index(tokens=None, repo_slug=None, sort_by=None, terms=None):
+    """
+    Perform a search in Elasticsearch.
+
+    Args:
+        tokens (unicode): string of one or more words
+        repo_slug (unicode): repository slug
+        sort_by (string): field to sort by
+        terms: (dict): {"vocabulary name": ["term1" [, "term2"]]}
+    Returns:
+        results (SearchResults)
+    """
+    if terms is None:
+        terms = {}
+    if tokens is None:
+        # Retrieve everything.
+        search = Search(index=INDEX_NAME, doc_type=DOC_TYPE).query()
+    else:
+        # Search on title, description, and content_xml (minus markup).
+        search = Search(index=INDEX_NAME, doc_type=DOC_TYPE)
+        multi = query.MultiMatch(
+            query=tokens, fields=["title", "description", "content_stripped"])
+        search = search.query(multi)
+    if terms != {}:
+        # Filter further on taxonomy terms.
+        search = search.query("match", **terms)
+    if repo_slug is not None:
+        # Filter further on repository.
+        search = search.query("match", repository=repo_slug)
+    if sort_by is not None:
+        # Temporary workaround; the values in sorting.py should be updated,
+        # but for now Haystack is still using them. Also, the hyphen is
+        # required because we sort the numeric values high to low.
+        if not sort_by.startswith("title"):
+            sort_by = "-xa_{0}".format(sort_by)
+        search = search.sort(sort_by, "id")
+    return SearchResults(search)
+
+
+@statsd.timer('lore.elasticsearch.bulk_index')
+def index_resources(resources):
+    """Add/update records in Elasticsearch."""
+
+    if hasattr(resources, "values_list"):
+        # If it's a queryset, get the IDs the smart way.
+        resource_ids = resources.values_list("id", flat=True)
+    else:
+        resource_ids = [x.id for x in resources]
+    # Terms assigned to the resources.
+    term_info = get_resource_terms(resource_ids)
+
+    if hasattr(resources, "iterator"):
+        # If it's a queryset, be efficient.
+        resources = resources.iterator()
+
+    ensure_vocabulary_mappings(term_info)
+
+    # Perform bulk insert using Elasticsearch directly.
+    conn = get_conn()
+    insert_count, errors = bulk(
+        conn,
+        (resource_to_dict(x, term_info.get(x.id, {})) for x in resources),
+        index=INDEX_NAME,
+        doc_type=DOC_TYPE
+    )
+
+    if errors != []:
+        log.error("Error during bulk insert: %s", errors)
+
+    return insert_count
+
+
+@statsd.timer('lore.elasticsearch.delete_index')
+def delete_index(resource):
+    """Delete a record from Elasticsearch."""
+    conn = get_conn()
+    try:
+        conn.delete(
+            index=INDEX_NAME, doc_type=DOC_TYPE, id=resource.id)
+        refresh_index()
+    except NotFoundError:
+        # We tried to delete something that wasn't in the index.
+        # In that case, we don't care; the desired outcome is identical.
+        pass
+
+
+def resource_to_dict(resource, term_info=None):
+    """
+    Retrieve important values from a LearningResource to index.
+
+    This dict corresponds to the mapping created in Elasticsearch.
+
+    The titlesort bits, with the "0" and "1" prefixes, were copied from
+    the prepare_titlesort function in search/search_indexes.py. It was there
+    to make blank titles sort to the bottom instead of the top.
+
+    Args:
+        resource (LearningResource): Item to convert to dict.
+        term_info (dict): Vocabulary terms assigned to resource.
+    Returns:
+        rec (dict): Dictionary representation of the LearningResource.
+    """
+
+    if term_info is None:
+        term_info = {}
+    rec = {
+        "title": resource.title,
+        # The zero is for sorting blank items to the bottom. See below.
+        "titlesort": "0{0}".format(resource.title.strip()),
+        "id": resource.id,
+        "_id": resource.id,  # The ID used by Elasticsearch.
+        "resource_type": resource.learning_resource_type.name,
+        "description": resource.description,
+        "content_xml": resource.content_xml,
+        "content_stripped": strip_xml(resource.content_xml),
+        "xa_nr_views": resource.xa_nr_views,
+        "xa_nr_attempts": resource.xa_nr_attempts,
+        "xa_avg_grade": resource.xa_avg_grade,
+        "xa_histogram_grade": resource.xa_histogram_grade,
+    }
+
+    course = get_course_metadata(resource.course_id)
+    rec["preview_url"] = get_preview_url(
+        resource,
+        org=course["org"],
+        course_number=course["course_number"],
+        run=course["run"],
+    )
+    rec["run"] = course["run"]
+    rec["course"] = course["course_number"]
+    rec["repository"] = course["repo_slug"]
+
+    # Index term info. Since these fields all use the "not_analyzed"
+    # index, they must all be exact matches.
+    for key, vals in term_info.items():
+        rec[key] = vals
+
+    # If the title is empty, sort it to the bottom. See above.
+    if rec["titlesort"] == "0":
+        rec["titlesort"] = "1"
+
+    # Keys that may have unicode issues.
+    text_keys = (
+        'title', 'titlesort', 'resource_type', 'description', 'content_xml',
+        'content_stripped',
+    )
+    for key in text_keys:
+        try:
+            # Thanks to unicode_literals above, this works in
+            # Python 2 and Python 3. Avoid trying to decode a string
+            # if it's already unicode.
+            if not isinstance(rec[key], type("")):
+                rec[key] = rec[key].decode('utf-8')
+        except AttributeError:
+            pass  # Python 3
+    return rec
+
+
+def clear_index():
+    """Wipe the index."""
+    conn = get_conn()
+    if conn.indices.exists(INDEX_NAME):
+        conn.indices.delete(INDEX_NAME)
+        conn.indices.create(INDEX_NAME)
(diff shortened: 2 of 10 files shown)


noisecapella:
That's a neat video, it does seem to clean up a lot of little messy things about Haystack

ShawnMilo:
@noisecapella: Thanks for checking it out. It's definitely more straightforward. Haystack was keeping things simple for the benefit of the Haystack authors, which made our use of Haystack pretty clumsy.

We can now get it to work the way we want it to _and_ it's simpler to use. Everybody wins.

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+
+from learningresources.models import get_preview_url, LearningResource
+from search.search_indexes import get_course_metadata
+
+log = logging.getLogger(__name__)
+
+DOC_TYPE = "learningresource"
+INDEX_NAME = settings.HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"]
+URL = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
+CONN = connections.create_connection(hosts=[URL])
```
noisecapella: If we initialize the connection here there will be a network connection happening we `import search.utils`. The other globals are just strings but this one does a whole lot more. My main concern is that this may cause an exception if the connection fails and it wouldn't be clean to put a try block around an import.

Could you make this a lazy connection instead where functions call something like `get_connection` which may initialize? This would follow how databases work in Django where the connection is made on first request.

ShawnMilo: So, keep CONN as a global variable, defaulting to `None`, and have each function in `utils.py` call `get_connection` (or maybe `initialize_connection`) which checks if it's none, initializes if necessary, and returns CONN?

noisecapella: Yep

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+PAGE_LENGTH = 10
```
noisecapella, line 28: Would it make sense to move this to `lore.settings`, or is this something that we probably wouldn't need to tweak?

ShawnMilo, line 28: I'm completely open to suggestions. It's only there because 10 is the actual Elasticsearch default, so when doing pagination I needed it in a variable.

If we change the value, it won't stop the default result from Elasticsearch (we'd have to change Elasticsearch's config file), but we would change the behavior of `SearchResults.page_count` and `get_page`.

I'd bet we change it once (as a part of getting it merged) or never, then never again. Based on that I don't think we should bother putting it in `settings.py` (because we'll never override it), but if someone thinks it should be in there I have no moral or philosophical objections.

importer/api/__init__.py
```diff
@@ -130,6 +131,10 @@ def import_course(bundle, repo_id, user_id, static_dir):
     import_static_assets(course, static_dir)
     import_children(course, src, None, '')
     populate_xanalytics_fields.delay(course.id)
+    # This triggers a bulk indexing of all LearningResource instances
```
noisecapella, line 134: Are we still doing the non-bulk indexing?

ShawnMilo, line 134: The old code is still doing the non-bulk indexing, so sadly you won't be able to watch it run faster. However, I did add statsd decorators, so it should still be possible to evaluate.

The new code does _only_ the bulk indexing for newly-created items (see `search/signals.py`  function `handle_resource_update`, which quits if the `created` flag is `True`). Then, individual indexing is only done upon update (by that `handle_resource_update` function).

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        self._search = search
+
+    def count(self):
+        """Total records matching the query."""
+        return self._search.count()
+
+    def page_count(self):
+        """Total number of result pages."""
+        count = self._search.count() / PAGE_LENGTH
+        if self._search.count() % PAGE_LENGTH > 0:
```
noisecapella: If `self._search.count()` is called twice does that ask Elasticsearch for the count twice or this is thing cached?

ShawnMilo: I don't know, but I think both occurrences should just be changed to `self.count()`.

ShawnMilo: I'm not aware of what caching Elasticsearch may or may not be doing, but I'm inclined to ignore it because I'm sure it's being handled correctly.

My guess is that it's cached by Elasticsearch after the first call after `.execute()`, but if it can possibly change then I don't want to cache anything on our end.

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def all(self):
+        """Return all results in a generator."""
+        for i in range(self.page_count()):
+            for hit in self.get_page(i+1):
+                yield hit
+
+    def __getitem__(self, i):
+        """Return result by index."""
+        return self._search[i:i+1].execute().hits[0]
```
noisecapella: You can just use `_search[i]` here, it seems to work the same way

search/utils.py
noisecapella: In this case we should use a local variable for `self._search.count()` so that the page count returned is consistent with a single count value. If `self._search.count()` returns two different numbers then the page count calculation doesn't make sense.

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        CONN.indices.delete_mapping(index=INDEX_NAME, doc_type=DOC_TYPE)
+
+    mapping = Mapping(DOC_TYPE)
+
+    mapping.field("course", "string", index="not_analyzed")
+    mapping.field("description_path", "string", index="no")
+    mapping.field("description", "string", index="analyzed")
+    mapping.field("preview_url", "string", index="no")
+    mapping.field("repository", "string", index="not_analyzed")
+    mapping.field("resource_type", "string", index="analyzed")
```
noisecapella: If we're faceting on resource_type wouldn't we want an exact match here? Not sure if it matters, I don't think any resource types have spaces at all

ShawnMilo: Yes, good catch.

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+    mapping.field("xa_histogram_grade", "float")
+    mapping.field("xa_nr_attempts", "integer")
+    mapping.field("xa_nr_views", "integer")
+
+    mapping.save(INDEX_NAME)
+
+    # If we've done this, we always want to re-index all existing
+    # LearningResource instances. This function will probably only
+    # ever be called by migrations.
+    index_resources(LearningResource.objects.all())
```
noisecapella, line 315: If the function is called by migrations I believe they would need to use their version of the LearningResource model. What if the user passed in `LearningResource.objects.all()` as an argument?

ShawnMilo, line 315: This is possible, but any concerns regarding using the real ORM are only possible during schema migrations. This is just a data migration.

There are other potential problems with using the fake (migration) models. In the event of any changes to the mapping or the model itself, the schema migrations would have run prior to the data migration, so everything will always be in the correct state.

I lean towards leaving it as is.

noisecapella, line 315: What if the next migration added is a schema migration? There's no guarantee that a data migration runs after all other schema migrations.

search/utils.py
```diff
@@ -0,0 +1,366 @@
...
+    vocab_names = set()
+    for result in term_info.values():
+        for key in result.keys():
+            vocab_names.add(key)
+    updated = False
+    # Add vocabulary to mapping if necessary.
+    for name in vocab_names:
+        if name in existing_vocabs:
+            continue
+        mapping.field(name, "string", index="not_analyzed", null_value="empty")
```
noisecapella, line 360: Is `null_value` necessary here since we don't use `null` as a value for vocabs or terms? From the documentation it seems like it's only useful for cases where we want to distinguish between actual `null` and `null` due to missing values.

search/utils.py
ShawnMilo, line 315: The migrations are run in a known order, and each individual migration is its own transaction.

There is no reason to put an "update mapping" data migration in the middle of other migrations; it can always go last.

Also, updating the indexes will do additional database lookups (for the terms and course information), so I think hits to the regular database can't be avoided.

search/utils.py
ShawnMilo, line 360: By using the `null_value` we can get counts via the aggregations (as in the demo video). That helps with what @giocalitri worked on recently.

search/utils.py
noisecapella, line 315: What if the update mapping data migration was added, then another schema migration was added? We can't just reorder the migrations or remove them after they're run.

I don't have a problem leaving this as it is however, if it causes a problem we can deal with it at that time.

ShawnMilo, line 315: I don't believe that can ever be a problem. Because:

Any change (to `LearningResource`, probably) which would require the mapping to be updated would have the mapping data migration in the same branch.

There will never be a case where two arbitrary branches are in the same release and one of them breaks the mapping in the other. If two branches made incompatible changes, their schema migrations would have to be coordinated anyway, so the migration ordering/numbering would already be handled.

search/utils.py
noisecapella, line 360: Makes sense, I was thinking we could just query with `None` but it looks like elasticsearch doesn't like that, and actual `empty` values will show up different from missing values so there's no confusion there.

noisecapella:
Almost through with this, looking at tests now

ShawnMilo:
Okay. The first round of issues were addressed and pushed. Diff:

``` diff
diff --git search/utils.py search/utils.py
index ffe01a1..2b8759b 100644
--- search/utils.py
+++ search/utils.py
@@ -24,10 +24,22 @@ log = logging.getLogger(__name__)
 DOC_TYPE = "learningresource"
 INDEX_NAME = settings.HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"]
 URL = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
-CONN = connections.create_connection(hosts=[URL])
+_CONN = connections.create_connection(hosts=[URL])
 PAGE_LENGTH = 10


+def get_conn():
+    """
+    Lazily create the connection.
+    """
+    # pylint: disable=global-statement
+    # This is ugly. Any suggestions on a way that doesn't require "global"?
+    global _CONN
+    if _CONN is None:
+        _CONN = connections.create_connection(hosts=[URL])
+    return _CONN
+
+
 def get_resource_terms(resource_ids):
     """
     Returns taxonomy metadata for LearningResources.
@@ -104,8 +116,9 @@ def index_resources(resources):
     ensure_vocabulary_mappings(term_info)

     # Perform bulk insert using Elasticsearch directly.
+    conn = get_conn()
     insert_count, errors = bulk(
-        CONN,
+        conn,
         (resource_to_dict(x, term_info.get(x.id, {})) for x in resources),
         index=INDEX_NAME,
         doc_type=DOC_TYPE
@@ -120,8 +133,9 @@ def index_resources(resources):
 @statsd.timer('lore.elasticsearch.delete_index')
 def delete_index(resource):
     """Delete a record from Elasticsearch."""
+    conn = get_conn()
     try:
-        CONN.delete(
+        conn.delete(
             index=INDEX_NAME, doc_type=DOC_TYPE, id=resource.id)
         refresh_index()
     except NotFoundError:
@@ -201,10 +215,11 @@ def resource_to_dict(resource, term_info=None):

 def clear_index():
     """Wipe the index."""
-    if CONN.indices.exists(INDEX_NAME):
-        CONN.indices.delete(INDEX_NAME)
-        CONN.indices.create(INDEX_NAME)
-        CONN.indices.refresh()
+    conn = get_conn()
+    if conn.indices.exists(INDEX_NAME):
+        conn.indices.delete(INDEX_NAME)
+        conn.indices.create(INDEX_NAME)
+        conn.indices.refresh()
     create_mapping()

     # re-index all existing LearningResource instances:
@@ -228,8 +243,9 @@ class SearchResults(object):

     def page_count(self):
         """Total number of result pages."""
-        count = self._search.count() / PAGE_LENGTH
-        if self._search.count() % PAGE_LENGTH > 0:
+        total = self._search.count()
+        count = total / PAGE_LENGTH
+        if total % PAGE_LENGTH > 0:
             count += 1
         return int(count)

@@ -247,7 +263,7 @@ class SearchResults(object):

     def __getitem__(self, i):
         """Return result by index."""
-        return self._search[i:i+1].execute().hits[0]
+        return self._search[i].execute().hits[0]


 def create_mapping():
@@ -262,11 +278,12 @@ def create_mapping():
     """

     # Create the index if it doesn't exist.
-    if not CONN.indices.exists(INDEX_NAME):
-        CONN.indices.create(INDEX_NAME)
+    conn = get_conn()
+    if not conn.indices.exists(INDEX_NAME):
+        conn.indices.create(INDEX_NAME)
     # Delete the mapping if an older version exists.
-    if CONN.indices.exists_type(index=INDEX_NAME, doc_type=DOC_TYPE):
-        CONN.indices.delete_mapping(index=INDEX_NAME, doc_type=DOC_TYPE)
+    if conn.indices.exists_type(index=INDEX_NAME, doc_type=DOC_TYPE):
+        conn.indices.delete_mapping(index=INDEX_NAME, doc_type=DOC_TYPE)

     mapping = Mapping(DOC_TYPE)

@@ -275,7 +292,7 @@ def create_mapping():
     mapping.field("description", "string", index="analyzed")
     mapping.field("preview_url", "string", index="no")
     mapping.field("repository", "string", index="not_analyzed")
-    mapping.field("resource_type", "string", index="analyzed")
+    mapping.field("resource_type", "string", index="not_analyzed")
     mapping.field("content_xml", "string", index="no")
     mapping.field("content_stripped", "string", index="analyzed")
     mapping.field("run", "string", index="not_analyzed")
@@ -293,7 +310,7 @@ def create_mapping():
     # LearningResource instances. This function will probably only
     # ever be called by migrations.
     index_resources(LearningResource.objects.all())
-    CONN.indices.refresh()
+    conn.indices.refresh()


 def refresh_index():
@@ -301,7 +318,8 @@ def refresh_index():
     Force a refresh instead of waiting for it to happen automatically.
     This should only be necessary during tests.
     """
-    CONN.indices.refresh(index=INDEX_NAME)
+    conn = get_conn()
+    conn.indices.refresh(index=INDEX_NAME)


 def ensure_vocabulary_mappings(term_info):

```

noisecapella:
I got an error running your migration:

```
(lore_venv)george@george-laptop:~/Projects/lore$ docker-compose run web ./manage.py migrate
Operations to perform:
  Synchronize unmigrated apps: audit, exporter, roles, compressor, cas, xanalytics, messages, rest, staticfiles, ui, rest_framework, storages, haystack, bootstrap3
  Apply all migrations: learningresources, search, sessions, admin, guardian, auth, contenttypes, taxonomy
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying search.0001_initial...No handlers could be found for logger "elasticsearch.trace"
Traceback (most recent call last):
  File "./manage.py", line 16, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 330, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 393, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 444, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/migration.py", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/special.py", line 183, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/src/search/migrations/0001_initial.py", line 19, in create_learning_resource_mapping
    create_mapping()
  File "/src/search/utils.py", line 295, in create_mapping
    index_resources(LearningResource.objects.all())
  File "/usr/local/lib/python2.7/dist-packages/statsd/client.py", line 30, in _wrapped
    return_value = f(*args, **kwargs)
  File "/src/search/utils.py", line 111, in index_resources
    doc_type=DOC_TYPE
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/helpers/__init__.py", line 182, in bulk
    for ok, item in streaming_bulk(client, actions, **kwargs):
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/helpers/__init__.py", line 110, in streaming_bulk
    for action, data in chunk:
  File "/src/search/utils.py", line 109, in <genexpr>
    (resource_to_dict(x, term_info.get(x.id, {})) for x in resources),
  File "/src/search/utils.py", line 195, in resource_to_dict
    rec[key] = rec[key].decode('utf-8')
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 47: ordinal not in range(128)
```

noisecapella:
If the connection in `get_conn` fails, does Elasticsearch retry until it succeeds or are we responsible for that? It may be a problem for celery workers which will hold onto the same connection for a while.

noisecapella:
Test coverage looks good

noisecapella:
Is there any reasonable way to test functionality, or should that wait for the next PR?

ShawnMilo:
The migration and `_CONN = None` are fixed. Next I want to figure out having `get_conn` test the connection and establish a new connection if necessary. As for whether we can reasonably test functionality, what did you have in mind beyond what the unit tests (and demo script) already demonstrate?

noisecapella:
I tried some DSL queries using the shell already and they worked, that's probably enough for this phase

ShawnMilo:
On Wed, Sep 16, 2015 at 4:17 PM, George Schneeloch <notifications@github.com

> wrote:
> If the connection in get_conn fails, does Elasticsearch retry until it
> succeeds or are we responsible for that? It may be a problem for celery
> workers which will hold onto the same connection for a while.

Created a stand-alone script to test this:

``` python
from elasticsearch_dsl.connections import connections
from elasticsearch.exceptions import ConnectionError

conn = connections.create_connection(hosts=["localhost:10200"])

def print_info():
    try:
        print conn.info()
    except ConnectionError:
        print "failed"
        return False
    return True


assert print_info() == True
print "now, stop the service"
raw_input()
assert print_info()  == False
print "now, start the service"
raw_input()
assert print_info() == True
```

This works, so it does repair itself after a temporary loss of
communication.

ShawnMilo:
@noisecapella:

I think the outstanding issues have been addressed: What happens when the connection dies and the migration error.

noisecapella:
:+1: after squash

PR #679, opened 2015-09-17
Removed /node directory, removed symlinks from node_modules

### LINKED ISSUE #677: Remove /node directory from Dockerfile
Code already populates /src/node_modules, /node/node_modules is basically a duplicate. I think the rationalle when I talked with Carson was that the user may want to have a set of node packages different from whatever is on /src, but that hasn't been very useful in my experience

### LINKED ISSUE #678: Remove symlinks from node_modules
By adding '--no-bin-links' to npm install we can direct it to not create sym links ([from here](http://askubuntu.com/questions/269727/npm-errors-when-installing-packages-on-windows-share)). This works around @pwilkins's issue with boot2docker. We would need to also change our code to point to the actual file but this should only be in a few places.

noisecapella:
Fixes #678 
Fixes #677

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,13 +28,6 @@ RUN pip install -r requirements.txt &&\
 RUN pip3 install -r requirements.txt &&\
     pip3 install -r test_requirements.txt
 
-# Install node development and production packages
-RUN mkdir /node
-COPY package.json /node/package.json
-RUN cd /node &&\
-    npm install &&\
-    npm install --production
-
 # Add project
 COPY . /src
 WORKDIR /src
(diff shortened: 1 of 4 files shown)


ShawnMilo:
Is this expected to make rebuilds faster as well?

noisecapella:
It should, we were doing this twice so there's no need for it there. We already do `npm install --production` on `docker-compose up`, and if you run javascript tests via tox it will install the test packages too at that time.

ShawnMilo:
Looks good, works locally. Good to merge after squash (I fixed a trivial pylint issue) and once Travis passes.

pwilkins:
I checked out this branch in an existing container, removed `/src/node_modules`, and ran `docker-compose build`.  No errors occurred and the app worked.

Yeah!  I have a working Docker again.

pwilkins:
I'd say this is OK to merge once the tests pass.

PR #683, opened 2015-09-17
Fixed spacing between edit and delete buttons

noisecapella:
This looks like a whitespace change but it translates differently from JSX.

### DIFF
--- a/ui/static/ui/js/manage_taxonomies.jsx
+++ b/ui/static/ui/js/manage_taxonomies.jsx
@@ -141,8 +141,7 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
               <span className="utility-features">
                 <a href="#">
                   <i className="fa fa-pencil"></i>
-                </a>
-                <a data-toggle="modal" data-target="#confirm-delete"
+                </a> <a data-toggle="modal" data-target="#confirm-delete"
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/manage_taxonomies.jsx
```diff
@@ -141,8 +141,7 @@ define('manage_taxonomies', ['react', 'lodash', 'jquery', 'utils', 'bootstrap'],
               <span className="utility-features">
                 <a href="#">
                   <i className="fa fa-pencil"></i>
-                </a>
-                <a data-toggle="modal" data-target="#confirm-delete"
+                </a> <a data-toggle="modal" data-target="#confirm-delete"
```
giocalitri, line 144: what is the difference? The original spacing should be collapsed to 1 space, shouldn't it?

noisecapella, line 144: No actually, they're just put right next to each other. It's a quirk of JSX

noisecapella, line 144: You can confirm this by going to https://facebook.github.io/react/jsx-compiler.html and trying out:

```
<div>
<a /> <a />
</div>
```

and comparing the output to

```
<div>
<a /> 
<a />
</div>
```

giocalitri:
this is good to be merged

PR #684, opened 2015-09-18
Switched django local storage to overwrite

giocalitri:
The default django local storage does not allow to overwrite files with
the same name and this is a problem for our static assets.

### DIFF
--- a/learningresources/tests/test_models.py
+++ b/learningresources/tests/test_models.py
@@ -145,7 +145,7 @@ class TestModels(LoreTestCase):
                 FILE_PATH_MAX_LENGTH -
                 # minus the length of the base path of the Django storage and
                 # the base path of the file location temporary dir
-                len(os.path.join(base_path, temp_dir_path)) -
+                len('/'.join([base_path, temp_dir_path])) -
                 # minus 1 for the extra "/" to joint the paths
                 1
             )
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -237,7 +237,9 @@ if(
 if LORE_USE_S3:
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
 else:
-    DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
+    # by default use django.core.files.storage.FileSystemStorage with
+    # overwrite feature
+    DEFAULT_FILE_STORAGE = 'storages.backends.overwrite.OverwriteStorage'
 
 
 # Lore preview settings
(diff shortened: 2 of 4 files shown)


learningresources/tests/test_models.py
```diff
@@ -145,7 +145,7 @@ def test_static_asset_filename_length(self):
                 FILE_PATH_MAX_LENGTH -
                 # minus the length of the base path of the Django storage and
                 # the base path of the file location temporary dir
-                len(os.path.join(base_path, temp_dir_path)) -
+                len('/'.join([base_path, temp_dir_path])) -
```
ShawnMilo, line 148: Pedantic: `/` should be `os.sep`.

ShawnMilo, line 148: Although, why not `os.path.join` instead?

ShawnMilo:
This looks fine once tests pass.

learningresources/tests/test_models.py
giocalitri, line 148: it was giving a problem in that context: for some reason it was ignoring the first variable `base_path`, maybe because it does not start with a `/`, but it does not make much sense.
Maybe I need to investigate more, but considering that it is in a test and just to calculate a length, I think it does not matter

giocalitri, line 148: Here is the explanation.

The join function works as it should: the problem is that `temp_dir_path` starts with `/` so all the previous paths are ignored.
Honestly I thought that `os.path.join` was doing a simple string concatenation, so I must be careful in the future. 

```
def join(a, *p):
    """Join two or more pathname components, inserting '/' as needed.
    If any component is an absolute path, all previous path components
    will be discarded.  An empty last part will result in a path that
    ends with a separator."""
    path = a
    for b in p:
        if b.startswith('/'):
            path = b
        elif path == '' or path.endswith('/'):
            path +=  b
        else:
            path += '/' + b
    return path
```

lore/settings.py
```diff
@@ -237,7 +237,9 @@ def get_var(name, default):
 if LORE_USE_S3:
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
 else:
-    DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
+    # by default use django.core.files.storage.FileSystemStorage with
+    # overwrite feature
+    DEFAULT_FILE_STORAGE = 'storages.backends.overwrite.OverwriteStorage'
```
pdpinch, line 242: This should fix the issue with tests, but do we need to make some changes to the S3 configuration as well?

giocalitri, line 242: After thinking more about it, I don't think there is a problem with S3, since the problem was related to the local storage, but we may want to make some tests

PR #685, opened 2015-09-18
Added REST api view to get and delete a course

### LINKED ISSUE #469: API for deleting a course
As a tester, I'd like to be able to delete courses from a repository. Curators may also wish to do this.
- [ ] Create API for deleting courses.
  - [ ] This is presumably a cascading delete.
  - [ ] permissions should be the same as for import: administrators and curators
  - [ ] need to delete static files as well

---

A privileged LORE user (administrator, curator?) should be able to delete learning resources from a repository.

**Questions:**
1. What user roles can perform this action?
2. How will the UI support this feature?
3. Is implementing it as web service API sufficient initially?
4. Do deletes cascade?  That is, if a course is deleted, are all its learningresources also deleted?

@pdpinch @carsongee

giocalitri:
Closes #469

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -178,6 +178,29 @@ Create a repository by providing its JSON representation.
             ]
         }
 
+## Course [/repositories/{repo_slug}/courses/{course_id}/]
+
++ Parameters
+    + repo_slug: `physics-1` (string, required) - slug for the repository
+    + course_id: `1` (number, required) - ID of the course
+
+### Retrieve a Course [GET]
+
++ Response 200 (application/json)
+
+        {
+            "id": 1,
+            "org": "MITx",
+            "course_number": "1.Mech",
+            "run": "1T2010"
+        }
+
+### Delete Course [DELETE]
+
+Deletes a course, all learning resources and static assets in it.
+
++ Response 204 (application/json)
+
 ## Group SearchResults
 
 ## SearchResult Collection [/repositories/{repo_slug}/search/{?q,sortby,selected_facets}]
--- a/rest/tests/test_course.py
+++ b/rest/tests/test_course.py
@@ -4,11 +4,20 @@ REST tests for courses
 from __future__ import unicode_literals
 
 from rest.tests.base import (
+    REPO_BASE,
     RESTAuthTestCase,
     RESTTestCase,
 )
 
 from learningresources.api import create_course
+from learningresources.models import (
+    Course,
+    LearningResource,
+    LearningResourceType,
+    Repository,
+    StaticAsset,
+)
+from taxonomy.models import Term, Vocabulary
 
 
 class TestCourse(RESTTestCase):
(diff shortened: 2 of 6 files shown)


giocalitri:
Missed a couple of lines in the coverage

rest/tests/test_course.py
```diff
@@ -53,6 +61,169 @@ def test_list_courses(self):
...
+        self.count_resources(
+            repositories=1,
+            courses=1,
+            learning_resources=1,
+            learning_resource_types=8,
+            terms=1,
+            vocabularies=1
+        )
+
+        # the course remaining is the one not deleted
```
noisecapella, line 201: These comments are great, it makes it really easy to follow this test

rest/tests/test_course.py
```diff
@@ -53,6 +61,169 @@ def test_list_courses(self):
...
+        courses = Course.objects.all()
+        self.assertEqual(courses[0].id, self.course.id)
+
+        # and the learning resource remaining is associated with the remaining
+        # course
+        learning_resources = LearningResource.objects.all()
```
noisecapella: Can you iterate through all of these instead of just checking the first one?

rest/tests/test_course.py
```diff
@@ -85,3 +256,95 @@ def test_get_courses(self):
...
+
+        # recreate course
+        self.course = create_course(**course_params)
+
+        # administrator
+        self.login(self.user.username)
+        self.delete_course(
+            self.repo.slug,
+            self.course.id
+        )
```
noisecapella, line 379: Can you add a test for PATCH and PUT on a course? It would be fine to just assert a 405 status code since they're not implemented, no need to get complicated.

noisecapella, line 379: Also POST on `courses`

rest/views.py
```diff
@@ -132,6 +133,40 @@ def get_queryset(self):
...
+    def get_object(self):
+        """
+        Return details about a course in a repo
+        """
+        try:
+            course = Course.objects.get(id=int(self.kwargs['course_id']))
+        except (Course.DoesNotExist, ValueError):
+            raise Http404()
+        try:
+            assert course.repository.slug == self.kwargs['repo_slug']
```
noisecapella: We shouldn't be using `assert` for checks in production, this could theoretically be skipped if the server ran `python -O`, and from a clean code point of view asserts verify something we know to be true but assert to be on the safe side. Since this could reasonably be false it should just be an if statement instead.

apiary.apib
```diff
@@ -178,6 +178,29 @@ Create a repository by providing its JSON representation.
             ]
         }
 
+## Course [/repositories/{repo_slug}/courses/{course_id}]
```
noisecapella: Can you add a trailing `/` to be consistent with the other endpoints?

rest/tests/test_course.py
```diff
@@ -53,6 +61,169 @@ def test_list_courses(self):
...
+        }
+        repo_res = self.create_repository(another_repo_dict)
+
+        # try to get the right course but on wrong repo
+        self.delete_course(
+            repo_slug=repo_res['slug'],
+            course_id=self.course.id,
+            expected_status=404
+        )
+
```
noisecapella, line 242: Do you have a test to check that, given a course c1 in repo r1 and course c2 in repo c2, that /repository/r1/course/c2 and /repository/r2/course/c1 gets a 404?

rest/views.py
```diff
@@ -132,6 +133,40 @@ def get_queryset(self):
...
+            raise Http404()
+        return course
+
+    @statsd.timer('lore.rest.course_detail_delete')
+    def delete(self, request, *args, **kwargs):
+        """
+        Deletes a course in a repo with all the related learning resources and
+        static assets
+        """
+        course = self.get_object()
```
noisecapella: It looks like you can just call the super method here

rest/tests/test_course.py
giocalitri: there is only one: see line `191`

rest/tests/test_course.py
giocalitri, line 242: I guess I miss the latter: will do

rest/views.py
giocalitri: makes sense

rest/views.py
giocalitri: will do

apiary.apib
giocalitri: will do

rest/tests/test_course.py
giocalitri: changed anyway in case we want to change the test in future

noisecapella:
:+1:

PR #704, opened 2015-09-22
Pinned PyTest to a version < 2.8

giocalitri:
pytest-capturelog has problems with the latest version and
raises warnings

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,6 +1,6 @@
 # Core project/common application testing requirements
 
-pytest
+pytest<2.8
 pytest-pep8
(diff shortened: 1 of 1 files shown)


noisecapella:
What do the warnings look like? Would they show up in travis logs?

giocalitri:
no, they do not show up anywhere because the `pytest-capturelog` uses a deprecated function

giocalitri:
BTW: this is the issue 
https://github.com/eisensheng/pytest-catchlog/issues/3

noisecapella:
Sounds good, feel free to merge

PR #709, opened 2015-09-22
Static assets are served again using local storage

### LINKED ISSUE #698: Links to static assets don't work locally

giocalitri:
Closes #698

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -17,7 +17,6 @@ from django.http import Http404, StreamingHttpResponse
 from django.shortcuts import (
     render,
     redirect,
-    get_object_or_404,
 )
 from guardian.decorators import permission_required_or_403
 from guardian.shortcuts import get_perms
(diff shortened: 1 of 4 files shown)


ui/views.py
```diff
@@ -266,7 +265,10 @@ def serve_static_assets(request, path):
     # first check if the user has access to the file
     media_path = os.path.join(STATIC_ASSET_PREFIX, path)
     file_path = os.path.join(settings.MEDIA_ROOT, media_path)
-    static_asset = get_object_or_404(StaticAsset, asset=media_path)
+    static_assets = StaticAsset.objects.filter(asset=media_path)
```
noisecapella, line 268: Is it allowed for two static assets to have the same path?

giocalitri, line 268: now yes: there is not check during the import that the path should be unique for a course id

noisecapella:
:+1:

Issue #662, opened 2015-09-15
Free tagging terms are not loaded in the taxonomy panel until page refreshes


giocalitri:
This used not to be a big deal because the search result page was reloading terms at each refresh (or search, or faceting) of the results. 
Now that we don't reload the page that often any more it is more visible.
Terms need to be reloaded each time we open the taxonomy panel.

noisecapella:
#642 should have fixed this

noisecapella:
Actually I misread this

noisecapella:
This functionality is included in #522

PR #710, opened 2015-09-23
Refresh Taxonomy panel each time user open it

amir-qayyum-khan:
> Terms need to be reloaded each time we open the taxonomy panel.

Issue https://github.com/mitodl/lore/issues/662

@giocalitri @pdpinch @noisecapella

### DIFF
--- a/ui/static/ui/js/listing.js
+++ b/ui/static/ui/js/listing.js
@@ -88,6 +88,14 @@ define('listing',
         $('.cd-panel').removeClass('is-visible');
       }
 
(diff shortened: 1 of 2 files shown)


noisecapella:
I think loading the taxonomy panel on click is a good change, but why not use `React.unmountComponentAtNode(container)` in the loader instead of `componentWillReceiveProps`? This is what the other panels do

noisecapella:
I vote to close this since this would be fixed in #522 which will hopefully be merged soon

giocalitri:
go for it

PR #711, opened 2015-09-23
Implemented verification of resource types on models

noisecapella:
Related to #504
Requirement for #522 

This doesn't include an API to check what term resource links would be deleted, that will be in a separate PR

### DIFF
--- a/rest/views.py
+++ b/rest/views.py
@@ -219,15 +219,15 @@ class VocabularyDetail(RetrieveUpdateDestroyAPIView):
         removed. Note that partial_update will call this function too.
         """
         vocab = self.get_object()
-        if 'learning_resource_types' in self.request.data:
+        new_types = self.request.data.get('learning_resource_types', None)
+        if new_types is not None:
             # Since LearningResourceTypes indicate which relationships
             # are valid between Terms and LearningResources, we need to
             # remove the newly invalid relationships caused by this update.
-            new_types = set(self.request.data['learning_resource_types'])
             old_types = set(
                 t.name for t in vocab.learning_resource_types.all()
             )
-            removed_types = old_types - new_types
+            removed_types = old_types - set(new_types)
 
             with transaction.atomic():
                 for term in vocab.term_set.all():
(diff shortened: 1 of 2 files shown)


rest/views.py
```diff
@@ -212,6 +213,43 @@ def get_queryset(self):
...
+            for resource in term.learning_resources.all():
+                if resource.learning_resource_type.name in removed_types:
+                    term.learning_resources.remove(resource)
+
+    def update(self, request, *args, **kwargs):
+        """
+        Override to remove resource term links if resource type is
+        removed.
+        """
+        vocab = get_object_or_404(Vocabulary, slug=kwargs['vocab_slug'])
```
giocalitri: if you fetch the vocabulary in this way here, why the `get_queryset` does it in a different way?

Wouldn't be better to get rid of `get_queryset`, override `get_object` and use `self.get_object` here?

rest/views.py
```diff
@@ -212,6 +213,43 @@ def get_queryset(self):
...
+        if 'learning_resource_types' in self.request.data:
+            new_types = set(self.request.data['learning_resource_types'])
```
giocalitri: I noticed that you do not do a sanity check of the `learning_resource_types` in the `request`. Do you need to do it?

noisecapella: I'll add an assertion for this case

rest/views.py
noisecapella: will do

rest/views.py
```diff
@@ -212,6 +212,30 @@ def get_queryset(self):
...
+            old_types = set(
+                t.name for t in vocab.learning_resource_types.all()
+            )
+            removed_types = old_types - new_types
+
+            for term in vocab.term_set.all():
+                for resource in term.learning_resources.all():
+                    if resource.learning_resource_type.name in removed_types:
+                        term.learning_resources.remove(resource)
```
giocalitri: this will trigger one signal per learning resource: do we want a bulk update also here?

noisecapella: We do need the reindexing to happen so we can't do this efficiently until the elasticsearch DSL work is in place. But I should add a transaction just in case it fails halfway through

giocalitri:
Feel free to merge as soon as tests pass

noisecapella:
Looks like I forgot to squash this... whoops

PR #717, opened 2015-09-23
Fixed id attribute for vocab and term select

noisecapella:
This change prevents id attribute conflicts for more than one component on the same page

### DIFF
--- a/ui/static/ui/js/learning_resources.jsx
+++ b/ui/static/ui/js/learning_resources.jsx
@@ -62,14 +62,15 @@ define('learning_resources',
 
       var slug = this.props.selectedVocabulary.slug;
 
+      var vocabId = "vocab-" + slug;
       return <div className="form-group">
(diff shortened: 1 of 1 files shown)


ui/static/ui/js/learning_resources.jsx
```diff
@@ -62,14 +62,15 @@ define('learning_resources',
 
       var slug = this.props.selectedVocabulary.slug;
 
+      var vocabId = "vocab-" + slug;
```
giocalitri, line 65: extremely low importance nit: why not `"vocabulary-" + slug`?

giocalitri:
Only that comment, but it can be merged also as it is

ui/static/ui/js/learning_resources.jsx
noisecapella, line 65: No particular reason, it just matches what was there before

PR #721, opened 2015-09-24
Changed get_or_create to create

### LINKED ISSUE #562: Investigate eliminating "get" overhead on course imports.
From Carson's [comment](https://github.com/mitodl/lore/pull/557#issuecomment-133473829):

> It looks like we probably have a pretty good performance opportunity with get_or_create just being create during the import since get is taking up about 22% of the time on large course imports.

noisecapella:
Fixes #562 

Before: 29.013169265s
After: 31.766057494s

I'm not sure there's much of a gain here but I personally think it's slightly cleaner this way.

### DIFF
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -73,7 +73,7 @@ def create_course(org, repo_id, course_number, run, user_id):
         "repository_id": repo_id,
     }
     with transaction.atomic():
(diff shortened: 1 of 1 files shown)


giocalitri:
I guess that if we avoid making a useless query it's always better.

you can merge it
:+1:

PR #722, opened 2015-09-24
Fixed newrelic not reporting

bdero:
The newrelic agent is running, but it's not reporting because it's unable to create background threads to perform monitoring tasks. Uwsgi disables threading in python interpreters by default.

For more info, see the "Mandatory options" section of: https://docs.newrelic.com/docs/agents/python-agent/hosting-mechanisms/python-agent-uwsgi

@pdpinch


pdpinch:
:+1: once tests pass

PR #729, opened 2015-09-25
Fixing memory usage in migration

giocalitri:


bdero:
Looks good once tests pass. :+1:

noisecapella:
What does this do? Wouldn't we want our migrations to be wrapped in transactions?

giocalitri:
we are having problems with the memory during transactions

bdero:
We might want to take the conversation over here instead of in the Lore room, but @giocalitri is currently testing if the large transaction being built by this migration is the cause of a steep linear memory increase I'm seeing on the dyno trying to run this migration.
Since the actions of this migration are idempotent, there shouldn't be problems if the migration is killed while in the middle of running.

That being said, this shouldn't be merged until we know that it's the transaction and not something else.

giocalitri:
The atomic kwarg itself does not solve the problem, I modified the code to avoid to load the entire queryset in memory
I tested it and it should work: before the memory was spiking; now it raises but does not go at the same levels I saw before

giocalitri:
We may want to remove the `atomic=False` since I think the new code should take care of the problem

bdero:
Ah, good - let's add the transaction back, then.

giocalitri:
Done

noisecapella:
The `.all` is still there. You may be able to just do `objects.order_by('id').iterator()`, not sure if it makes a difference however.

giocalitri:
@noisecapella as far as I remember the `.all` should not actually pull the data until you access the objects, so it just builds a queryset and querysets are lazy (see [here](https://docs.djangoproject.com/en/1.8/topics/db/optimization/#understand-queryset-evaluation))

noisecapella:
Sounds good to me

giocalitri:
actually you are right

giocalitri:
I'm going to change the code following @noisecapella suggestion (thanks!)

giocalitri:
Reading the documentation carefully, I think I was right (just misunderstood one part of it), but in any case I removed the `.all` because it's useless. Merging now.

PR #732, opened 2015-09-25
Added debug_toolbar

### LINKED ISSUE #730: Install and use Django Debug Toolbar
This would help in quick investigation of performance problems. I think it requires a Python package and a name in `INSTALLED_APPS`

noisecapella:
Fixes #730

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,7 @@ web:
   build: .
   command: >
     /bin/bash -c '
+    export HOST_IP=$(netstat -nr | grep ^0\.0\.0\.0 | awk "{print \$2}") &&
     sleep 3 &&
     npm install --production --no-bin-links &&
     python manage.py migrate &&
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -81,6 +81,8 @@ INSTALLED_APPS = (
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'debug_toolbar',
+    'elastic_panel',
     'storages',
     'compressor',
     'bootstrap3',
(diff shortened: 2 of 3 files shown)


noisecapella:
BTW I needed to add 

```
INTERNAL_IPS = ('172.17.42.1',)
```

to get it to turn on and work with the docker container. I could add this as part of the docker environment variables if it works for boot2docker too.

giocalitri:
I'm not sure that IP is stable.

noisecapella:
It looks like we can figure out the IP using this command (from [here](http://blog.michaelhamrah.com/2014/06/accessing-the-docker-host-server-within-a-container/)):

```
netstat -nr | grep '^0\.0\.0\.0' | awk '{print $2}'
```

I'll look into updating this PR to create an environment variable for that and populating `INTERNAL_IPS`

noisecapella:
This is basically done except that the session expires really frequently, need to investigate before this gets merged.

ShawnMilo:
In my experience, session problems like that are often due to a clock issue.

If not, check out what the [SESSION_COOKIE_AGE](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SESSION_COOKIE_AGE) setting is, maybe `log.debug` it or something. Perhaps it's something we'd want to add an environment variable for, so we can make sessions for developers last longer.

Also, maybe check the [SESSION_EXPIRE_AT_BROWSER_CLOSE](https://docs.djangoproject.com/en/1.8/ref/settings/#session-expire-at-browser-close) setting. Maybe closing a browser or a tab is killing the session.

noisecapella:
I'm having trouble reproducing it. Whoever reviews this should keep a look out but I feel like the session expiration issue is unrelated

noisecapella:
Just updated the commit to also show elasticsearch queries!

ShawnMilo:
That's cool!

noisecapella:
Any objects to merging this?

ShawnMilo:
If I understand correctly that the toolbar never loads in production (as the setting comment implies), then the code looks good. I haven't tested it locally. I can do it now, unless someone has done it already.

noisecapella:
That's correct, it shouldn't be seen in production. Can you try it out locally? It will require a docker rebuild

ShawnMilo:
Yeah, I was just realizing it's going to cause me to completely rebuild again to work on my Elasticsearch version, but I found `docker save` and `docker load`, so I'm just backing it up and I'll restore it afterwards. :smile:

ShawnMilo:
Yep, runs fine for me locally. And I was able to do other stuff (import course, create, assign, and filter by taxonomies). Good to merge.

PR #735, opened 2015-10-01
Fixed resource ids

noisecapella:
The results returned from the server have changed slightly, this PR updates our example data. This should fix React complaining about duplicate key values due to `resource.id` not existing.


giocalitri:
This can be merged

PR #700, opened 2015-09-21
rc/0.11.0

pwilkins:
Release 0.11.0
- [x] Reorganize Learning Resource panel to use three tabs
  [Jamie Folsom](jamiefolsom@gmail.com)
- [x] Added datatable proof of concept
  [George Schneeloch](gschneel@mit.edu)
- [x] Added REST api view to get and delete a course
  [Giovanni Di Milia](gdimilia@mit.edu)
- [x] Added calls to get_conn() where conn is used implicitly.
  [Shawn Milochik](shawn@milochik.com)
- [x] Added more detail to confirmation message for delete vocabulary
  [Amir Qayyum Khan](amir.qayyum.khan@gmail.com)
- [x] Moved save button to right of the term for edit term inside taxonomy panel
  [Amir Qayyum Khan](amir.qayyum.khan@gmail.com)
- [x] Switched django local storage to overwrite
  [Giovanni Di Milia](gdimilia@mit.edu)
- [x] Implemented page_size parameter to allow users to set page size
  [George Schneeloch](gschneel@mit.edu)
- [x] Fixed spacing between edit and delete buttons
  [George Schneeloch](gschneel@mit.edu)
- [x] Removed /node directory, removed symlinks from node_modules
  [George Schneeloch](gschneel@mit.edu)
- [x] Adds elasticsearch-dsl and adds it alongside Haystack for now.
  [Shawn Milochik](shawn@milochik.com)
- [x] Added save and close button to learning resource panel
  [Amir Qayyum Khan](amir.qayyum.khan@gmail.com)
- [x] Added tests for listing page
  [George Schneeloch](gschneel@mit.edu)
- [x] Removed lib/ from gitignore
  [George Schneeloch](gschneel@mit.edu)
- [x] Switched to minimized JS for libraries
  [George Schneeloch](gschneel@mit.edu)
- [x] Added REST API view to list courses in repository
  [Giovanni Di Milia](gdimilia@mit.edu)
- [x] Removed react-addons bower package, addons actually live in react package
  [George Schneeloch](gschneel@mit.edu)
- [x] Fixed pagination links
  [George Schneeloch](gschneel@mit.edu)
- [x] Increased requirejs timeout
  [George Schneeloch](gschneel@mit.edu)
- [x] Fixed exact repository search bug
  [George Schneeloch](gschneel@mit.edu)
- [x] Fixed clear export bug
  [George Schneeloch](gschneel@mit.edu)

PR #739, opened 2015-10-02
Chunked bulk indexing to lower memory footprint

### LINKED ISSUE #680: Add Haystack <-> Elasticsearch-DSL converters
Add converter to transform existing requests from what Haystack expects to what Elasticsearch-DSL uses.

Add a converter to transform the results of an Elasticsearch-DSL query to what the existing front-end/API expects.

After this is merged, we can start ripping out the Haystack dependency. In the future, the code should then be cleaned up to remove this work, so we can directly use the cleaner output of Elasticsearch-DSL.

Part of #668.

noisecapella:
Related to #680, still need to remove Haystack to finish it


noisecapella:
PR #745 must be merged before the tests will pass

search/utils.py
```diff
@@ -47,13 +47,28 @@ def get_conn():
...
+
     # Make sure everything exists.
     if not _CONN.indices.exists(INDEX_NAME):
-        _CONN.indices.create(INDEX_NAME)
+        raise ReindexException("Unable to find index {index_name}".format(
+            index_name=INDEX_NAME
+        ))
+
+    mapping = _CONN.indices.get_mapping()
+    if INDEX_NAME not in mapping:
```
ShawnMilo: This is checking for the index a second time, not the mapping. 

[This](https://github.com/mitodl/lore/pull/700#issuecomment-142340863) script contains the syntax to look `DOC_TYPE`.

search/tests/test_es_indexing.py
```diff
@@ -255,3 +265,36 @@ def test_sorting(self):
...
+            # Mapping doesn't exist yet
+            get_conn()
+
+        conn = get_conn(verify=False)
+        conn.indices.create(INDEX_NAME)
+        conn.indices.refresh()
+        create_mapping()
+
+        with self.assertRaises(Exception):
+            self.client.get(search_url)
```
ShawnMilo: Why does this raise an exception, and what is the exception type? Given that the index and mapping exist and are empty, why doesn't it just return no error and no results?

ShawnMilo:
This looks good, and tests pass for me locally after merging master.

I refreshed the index three times, and recreated it three times locally. All six runs took 44-49 seconds and consumed 390 MB of RAM.

Good to merge.

search/tests/test_es_indexing.py
noisecapella: I'm not sure, I'll look into it

noisecapella: I'm getting a TransportError here with `No mapping found for [id] in order to sort on`. It only goes away when I index the resources. I'll investigate a fix for this

noisecapella:
After investigating I think the problem is that `id` is dynamically mapped and it complains when being asked to sort on a mapping which doesn't exist yet since there are no resources at first. I tried just adding `id` to the mapping but this collides with Haystack somehow. I added a workaround to set `ignore_unmapped: True` on sort by `id`. I'll file an issue to reinvestigate this workaround, I feel like we will be able to remove it after getting rid of Haystack entirely.

ShawnMilo:
Okay. It works and tests pass, so as long an Travis completes successfully again then it's good to merge.

noisecapella:
One test was failing due to the mock complaining about a function not being called in a management command, even though it seems like it was called and it passed locally. I changed the test to just assert the reindexing instead, will merge when they pass

PR #740, opened 2015-10-02
Added limit to memory for Docker

noisecapella:
Having memory limits enforced in our development Docker containers may help catch errors. Note that these limits aren't enforced by default. I needed to set this line in `/etc/default/grub` for Ubuntu 15.04:

```
GRUB_CMDLINE_LINUX_DEFAULT="quiet splash cgroup_enable=memory swapaccount=1"
```

This causes a 10% performance hit according to Docker. I'm not sure how boot2docker is configured by default.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,7 @@ stats:
     - "8071:80"
 web:
   build: .
+  mem_limit: 384m
   command: >
     /bin/bash -c '
     export HOST_IP=$(netstat -nr | grep ^0\.0\.0\.0 | awk "{print \$2}") &&
@@ -53,6 +54,7 @@ web:
     - stats
 celery:
   image: lore_web
+  mem_limit: 384m
   command: >
     /bin/bash -c '
     sleep 3;


noisecapella:
Note that when you run `docker-compose run web bash` or something else, the other containers that get started will use whatever memory limit they used to have. You will need to do `docker-compose up` to set new memory limits, then `docker-compose run web bash` or whatever will have the correct memory limits.

giocalitri:
Did you (try to) figure out if with docker-compose we need to set that configuration line in `/etc/default/grub`?

noisecapella:
That line is for the host, for me it's Ubuntu 15.04 and I think most other people it would be boot2docker. It was necessary for me due to a warning message regarding swap limits being turned off. I'll investigate whether it's necessary for boot2docker.

ShawnMilo:
This looks straightforward enough.

However, could we just add the `mem_limit` setting in docker-compose.yml as described [here](https://docs.docker.com/compose/yml/)?

That could work on more systems without the need to make local config settings to the OS.

noisecapella:
This PR is about editing `docker-compose.yml`. Is it not? The OS settings are required (for me at least) to enforce the `mem_limit` that is specified. Otherwise you can see the memory limit in effect using `docker stats` but it won't stop you from using up more memory than is available. To have Ubuntu enforce the memory limit I needed the grub settings mentioned above which come from Docker's own docs: [Adjust memory and swap accounting](https://docs.docker.com/installation/ubuntulinux/#adjust-memory-and-swap-accounting)

ShawnMilo:
Yeah, brain fart. Sorry.

ShawnMilo:
Good to merge.

PR #741, opened 2015-10-05
Added warning message when user try to close learning resource panel without saving

### LINKED ISSUE #608: Warn user when closing learning resource panel without saving
As a curator, sometimes I accidentally close the learning resource panel before I've saved my changes. It would be nice to have a warning so I don't do that.

amir-qayyum-khan:
In this PR I added a message to ask user to save changes before exit. User can only exit LR panel if he save new changes or undo new changes.

Closes #608 
@pdpinch @noisecapella 

screen shot 2015-10-05 at 6 17 42 pm
screen shot 2015-10-05 at 6 17 46 pm
screen shot 2015-10-05 at 6 17 52 pm
screen shot 2015-10-05 at 6 18 14 pm
screen shot 2015-10-05 at 6 24 21 pm

### DIFF
--- a/ui/jstests/learningresources/test_learning_resource_panel.jsx
+++ b/ui/jstests/learningresources/test_learning_resource_panel.jsx
@@ -248,11 +248,136 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
 
         React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
           repoSlug="repo"
+          markDirty={function() {}}
           closeLearningResourcePanel={closeLearningResourcePanel}
           learningResourceId="1"
           ref={afterMount}/>);
       }
     );
+
+    QUnit.test(
+      'Assert that mark dirty sets properly when description change.',
+      function (assert) {
+        var done = assert.async();
+
+        var markDirtyState = false;
+        var markDirty = function (state) {
+          markDirtyState = state;
+        };
+
+        var afterMount = function (component) {
+          // wait for calls to populate form
+          waitForAjax(3, function () {
+            var textarea = React.addons.TestUtils.
+              findRenderedDOMComponentWithTag(
+                component,
+                'textarea'
+            );
+            React.addons.TestUtils.Simulate.change(
+              textarea, {target: {value: "x"}}
+            );
+            component.forceUpdate(function () {
+              assert.equal(component.state.description, "x");
+              assert.ok(markDirtyState, "Marked dirty");
+              done();
+            });
+          });
+        };
+        React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
+          repoSlug="repo"
+          markDirty={markDirty}
+          closeLearningResourcePanel={function() {}}
+          learningResourceId="1"
+          ref={afterMount}/>);
+      }
+    );
+
+    QUnit.test(
+      'Assert that mark dirty sets properly when terms change.',
+      function (assert) {
+        var done = assert.async();
+        var markDirtyState = false;
+        var markDirty = function (state) {
+          markDirtyState = state;
+        };
+        var closeLearningResourcePanel = function () {
+        };
+
+        var afterMount = function (component) {
+          // wait for calls to populate form
+          waitForAjax(3, function () {
+            // two menus: vocabulary and terms.
+            var $node = $(React.findDOMNode(component));
+
+            var $allSelects = $node.find("#vocabularies select");
+            assert.equal($allSelects.size(), 2);
+
+            var $vocabSelect = $node.find($allSelects).first();
+            assert.equal($vocabSelect.size(), 1);
+
+            // first vocab, two options
+            var $vocabOptions = $vocabSelect.find("option");
+            assert.equal($vocabOptions.size(), 2);
+
+            assert.equal($vocabOptions[0].selected, true);
+            assert.equal($vocabOptions[1].selected, false);
+
+            // TestUtils.Simulate.change only simulates a change event,
+            // we need to update the value first ourselves
+            $vocabSelect.val("prerequisite").trigger('change');
+            component.forceUpdate(function () {
+              assert.equal($vocabOptions[0].selected, false);
+              assert.equal($vocabOptions[1].selected, true);
+              // re-fetch the selects to get the prerequisite one
+              $allSelects = $node.find("#vocabularies select");
+              var termsSelect = $allSelects[1];
+              var $termsOptions = $(termsSelect).find("option");
+              assert.equal($termsOptions.size(), 2);
+              assert.equal($termsOptions[0].selected, true);
+              assert.equal($termsOptions[1].selected, false);
+
+              // remove selection for this vocabulary to not interfere to the rest of the test
+              $(termsSelect).val("").trigger('change');
+              component.forceUpdate(function () {
+                // Switch to difficulty
+                $vocabSelect.val("difficulty").trigger('change');
+                component.forceUpdate(function () {
+                  // re-fetch the selects to get the difficulty one
+                  $allSelects = $node.find("#vocabularies select");
+                  // update the term select
+                  termsSelect = $allSelects[1];
+                  $termsOptions = $(termsSelect).find("option");
+
+                  assert.equal($termsOptions.size(), 2);
+                  assert.equal($termsOptions[0].selected, false);
+                  assert.equal($termsOptions[1].selected, false);
+                  // the second vocabulary can be a multi select
+                  $(termsSelect)
+                    .val(["hard", "easy"])
+                    .trigger('change');
+                  markDirtyState = false;
+                  component.forceUpdate(function () {
+                    // when terms are change then markDirtyState flag is set.
+                    assert.equal(markDirtyState, true);
+                    assert.equal($termsOptions[0].selected, true);
+                    assert.equal($termsOptions[1].selected, true);
+                    done();
+                  });
+                });
+              });
+            });
+          });
+        };
+
+        React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
+          repoSlug="repo"
+          markDirty={markDirty}
+          closeLearningResourcePanel={closeLearningResourcePanel}
+          learningResourceId="1"
+          ref={afterMount}/>);
+      }
+    );
+
     QUnit.test(
       'Assert that LearningResourcePanel saves properly',
       function (assert) {
@@ -300,6 +425,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
         React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
           repoSlug="repo"
           learningResourceId="1"
+          markDirty={function() {}}
           closeLearningResourcePanel={closeLearningResourcePanel}
           refreshFromAPI={refreshFromAPI}
           ref={afterMount}/>);
@@ -353,6 +479,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
 
         React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
           repoSlug="repo"
+          markDirty={function() {}}
           closeLearningResourcePanel={closeLearningResourcePanel}
           learningResourceId="1"
           refreshFromAPI={refreshFromAPI}
@@ -386,6 +513,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
         };
         React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
           repoSlug="repo"
+          markDirty={function() {}}
           closeLearningResourcePanel={closeLearningResourcePanel}
           learningResourceId="1"
           ref={afterMount}/>);
@@ -418,6 +546,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
         };
         React.addons.TestUtils.renderIntoDocument(<LearningResourcePanel
           repoSlug="repo"
+          markDirty={function() {}}
           closeLearningResourcePanel={closeLearningResourcePanel}
           learningResourceId="1"
           ref={afterMount}/>);
@@ -482,6 +611,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
           <LearningResourcePanel
             repoSlug="repo"
             closeLearningResourcePanel={function() {}}
+            markDirty={function() {}}
             learningResourceId="1"
             ref={afterMount}/>
         );
--- a/ui/jstests/listing/test_listing.jsx
+++ b/ui/jstests/listing/test_listing.jsx
@@ -971,5 +971,63 @@ define(['QUnit', 'jquery', 'react', 'test_utils', 'utils', 'listing'],
       );
 
     });
+
+    QUnit.test('Open and close learning resource panel', function(assert) {
+      var done = assert.async();
+      TestUtils.initMockjax({
+        url: '/api/v1/repositories/test/learning_resources/' +
+        '1/?remove_content_xml=true',
+        type: 'GET',
+        responseText: learningResourceResponse
+      });
+      TestUtils.initMockjax({
+        url: '/api/v1/repositories/test/vocabularies/?type_name=course',
+        type: 'GET',
+        responseText: vocabularyResponse
+      });
+
+      var afterMount = function(component) {
+        waitForAjax(1, function() {
+          assert.notOk($('.cd-panel').hasClass("is-visible"));
+          component.openResourcePanel(1);
+          waitForAjax(2, function() {
+            assert.equal(component.state.currentResourceId , 1);
+            assert.ok($('.cd-panel').hasClass("is-visible"));
+            component.setState({
+              isLearningResourcePanelDirty: true
+            }, function () {
+              component.closeLearningResourcePanel();
+              assert.ok($('.cd-panel').hasClass("is-visible"),
+                "LR Panel should not close because it is marked dirty");
+              component.setState({
+                isLearningResourcePanelDirty: false
+              }, function() {
+                component.closeLearningResourcePanel();
+                assert.notOk($('.cd-panel').hasClass("is-visible"),
+                  "LR Panel should close");
+                done();
+              });
+            });
+          });
+        });
+      };
+
+      var options = {
+        allExports: listingOptions.allExports,
+        sortingOptions: listingOptions.sortingOptions,
+        imageDir: listingOptions.imageDir,
+        pageSize: 25,
+        repoSlug: listingOptions.repoSlug,
+        loggedInUsername: listingOptions.loggedInUsername,
+        updateQueryString: updateQueryString,
+        getQueryString: getQueryString,
+        showConfirmationDialog: function() {},
+        ref: afterMount
+      };
+
+      React.addons.TestUtils.renderIntoDocument(
+        React.createElement(Listing.ListingContainer, options)
+      );
+    });
   }
 );
--- a/ui/static/ui/js/learningresources/learning_resource_panel.jsx
+++ b/ui/static/ui/js/learningresources/learning_resource_panel.jsx
@@ -35,6 +35,8 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
 
         this.setState({
           vocabulariesAndTerms: newVocabulariesAndTerms
+        }, function () {
+          this.updateDirty();
         });
       },
 
@@ -115,8 +117,7 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
                 <button className="btn btn-lg btn-primary"
                         onClick={this.saveLearningResourcePanel}>
                   Save
-                </button>
-                <button className="btn btn-lg btn-success"
+                </button> <button className="btn btn-lg btn-success"
                         onClick={this.saveAndCloseLearningResourcePanel}>
                   Save and Close
                 </button>
@@ -133,7 +134,13 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
         this.setState({
           description: event.target.value,
           message: undefined
+        }, function() {
+          this.updateDirty();
         });
+      },
+      closeLearningResourcePanel: function(event) {
+        event.preventDefault();
+        this.props.closeLearningResourcePanel();
       },
       saveLearningResourcePanel: function (event) {
         event.preventDefault();
@@ -165,8 +172,12 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
           data: JSON.stringify(data)
         }).then(function () {
           thiz.setState({
-            message: "Form saved successfully!"
+            message: "Form saved successfully!",
+            descriptionOriginal: thiz.state.description,
+            vocabulariesAndTermsOriginal: thiz.state.vocabulariesAndTerms
           });
+          // user can close panel
+          thiz.updateDirty();
           if (closePanel) {
             thiz.props.closeLearningResourcePanel();
           }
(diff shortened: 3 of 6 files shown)


pdpinch:
Sorry Amir, we have a UX problem here. Now there is no way to exit the learning resource panel without saving. I think we need a cancel button... but I'm not sure where/what color.

amir-qayyum-khan:
@pdpinch we can use warning color here http://www.w3schools.com/bootstrap/bootstrap_buttons.asp

The button will be on right of **Save and Close**. And label can be
- Cancel
- Undo and Close

pdpinch:
Sounds good. Let's go with `Cancel`

amir-qayyum-khan:
@pdpinch added cancel button

screen shot 2015-10-12 at 1 34 28 pm
screen shot 2015-10-12 at 1 34 20 pm

amir-qayyum-khan:
@ShawnMilo can you review this.

ShawnMilo:
@amir-qayyum-khan: I found a small bug in testing.

When I edit text and then try to exit, I get the warning as expected. But then, if I open the same item _or_ the one beneath it and try to close without saving, I get the warning even if I don't change anything. This does not seem to happen for a repository with no vocabularies.

Video: https://youtu.be/oyNRdusH5Vw

amir-qayyum-khan:
Done @ShawnMilo

Ferdi:
I think it would have been better to have a popup , like the one we show when users try to delete a vocab.

"You have unsaved changes, are you sure you want to close?  yes/ no"

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +555,57 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
       assert.ok($(div).html().length === 0);
-      LearningResources.loader("repo", "1", function() {}, function() {}, div);
+      LearningResources.loader(
+        "repo", "1", function() {}, function() {}, false, div
+      );
       assert.ok($(div).html().length > 0);
     }
   );
+
+  QUnit.test('Assert learning resource panel do close',
```
ShawnMilo: Should this be  "does close" instead of "do close"?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -294,6 +295,32 @@ define('learning_resources',
...
+          Save
+        </button>  <button className="btn btn-lg btn-success"
+                onClick={this.saveAndCloseLearningResourcePanel}>
+          Save and Close
+        </button>
+      </p>;
+      if (!_.isEmpty(this.state.message) &&
+        !_.isEmpty(this.state.message.error)) {
+        // show cancel button if error message
+        actionButtons = <p>
```
ShawnMilo: nit: Is there a way to remove the duplication (lines 299-305 and lines 311-317)?

ShawnMilo:
It looks fine to me. I made two tiny comments, but neither is urgent. The wording should be fixed if it is wrong. 

Also needs to be squashed to a single commit. If the wording is correct (or when it is fixed), good to merge.

ShawnMilo:
Sorry, @amir-qayyum-khan -- I wrote my last comment before I saw @Ferdi's note. Please don't merge until a decision has been made.

amir-qayyum-khan:
Fixed
@ShawnMilo @Ferdi please review
screen shot 2015-10-15 at 5 09 44 pm
screen shot 2015-10-15 at 5 09 50 pm
screen shot 2015-10-15 at 5 10 01 pm

ShawnMilo:
@amir-qayyum-khan Thanks for the screenshots.

@Ferdi Do you like this UI change?

Ferdi:
Looks good. You say "click save" but save is not there, in the popup. It says "exit" -- exit what ? the app ?

I think what I suggested is simpler and more intuitive : 
"You have unsaved changes, are you sure you want to close? yes/ no"

ShawnMilo:
The functionality works and the code change from the previous version is minimal. Functionally it's good. Please change the wording as requested by Ferdi.

amir-qayyum-khan:
Done @ShawnMilo @Ferdi 

screen shot 2015-10-16 at 2 10 26 pm
screen shot 2015-10-16 at 2 11 42 pm

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,56 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+            );
+            done();
+          });
+        }
+      });
+
+      var afterMount = function (component) {
+        component.setState({closeLRPanel: true}, function () {
+          component.forceUpdate(function() {
+            assert.equal(component.state.closeLRPanel, true);
```
noisecapella: You must put `done` inside the last executed callback

ui/static/ui/js/listing.js
```diff
@@ -225,18 +230,26 @@ define('listing',
...
+          showConfirmationDialog,
           closeLearningResourcePanel,
+          closeLRPanel,
+          lastSelectedResourceId !== resourceId, // Is new resource
           $("#tab-1")[0]
         );
-        $('.cd-panel').addClass('is-visible');
-        StaticAssets.loader(repoSlug, resourceId, $("#tab-3")[0]);
-        XmlPanel.loader(repoSlug, resourceId, $("#tab-2")[0]);
+        // If use is not closing panel
```
noisecapella: Can you reword this?

amir-qayyum-khan:
Done @noisecapella

ui/static/ui/js/learning_resources.jsx
```diff
@@ -294,6 +295,16 @@ define('learning_resources',
         />;
       }
 
+      actionButtons = <p>
```
noisecapella: Can you move the `var` down to this line and remove it above?

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+          lrMountCount += 1;
+          return <LearningResourcePanel
+            repoSlug="repo"z
+            ref={this.afterMount}
+            closeLearningResourcePanel={this.props.closeLearningResourcePanel}
+            closeLRPanel={this.state.closeLRPanel}
+            learningResourceId="1" />;
+        },
+        afterMount: function () {
+          waitForAjax(9, function () {
```
noisecapella: What AJAX calls are happening here? This seems like a large number

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+      };
+
+      var Container = React.createClass({
+        getInitialState: function() {
+          return {closeLRPanel: false};
+        },
+        render: function() {
+          lrMountCount += 1;
+          return <LearningResourcePanel
+            repoSlug="repo"z
```
noisecapella: Stray character here

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+            if (lrMountCount > 1) {
+              done();
+            }
+          });
+        }
+      });
+
+      var afterMount = function (component) {
+        component.setState({closeLRPanel: true}, function () {
+          component.forceUpdate(function() {
```
noisecapella: This `forceUpdate` is unnecessary since the second callback above does the same thing

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+              closeLearningResourcePanelCount > 0, "Close panel was called"
+            );
+            if (lrMountCount > 1) {
+              done();
+            }
+          });
+        }
+      });
+
+      var afterMount = function (component) {
```
noisecapella: It seems like you can move the contents of this function inside the React class above

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+      var closeLearningResourcePanel = function() {
+        closeLearningResourcePanelCount += 1;
+      };
+
+      var Container = React.createClass({
+        getInitialState: function() {
+          return {closeLRPanel: false};
+        },
+        render: function() {
+          lrMountCount += 1;
```
noisecapella: We can't rely on counting how many times this function is called. What is the actual condition that you are waiting for?

ui/jstests/test-learning-resource.jsx
```diff
@@ -554,8 +561,60 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+          return <LearningResourcePanel
+            repoSlug="repo"z
+            ref={this.afterMount}
+            closeLearningResourcePanel={this.props.closeLearningResourcePanel}
+            closeLRPanel={this.state.closeLRPanel}
+            learningResourceId="1" />;
```
noisecapella: This is passing in a string here, can you pass it as a number instead to match the behavior of the rest of the code?

ui/jstests/test-learning-resource.jsx
amir-qayyum-khan: line 597 will be called at the end, because each time we change stage if container new LR panel renders.

ui/jstests/test-learning-resource.jsx
amir-qayyum-khan: when closeLRPanel = false then three calls.
when closeLRPanel = true then three calls on `componentDidMount` and three on closing LR panel `componentWillReceiveProps`

noisecapella: Why are we making AJAX calls when the user closes the panel?

amir-qayyum-khan:
Done @noisecapella

noisecapella:
It looks like there is still a AJAX request when the user attempts to close the panel. Can you look into this?

ui/jstests/test-learning-resource.jsx
noisecapella: Can you move the next line into the above `afterMount` function then delete this function? There is also no `done` in this function which would be necessary if you were to keep it

amir-qayyum-khan:
yes I do because
- state change
- we dont have prev state 

But let me look into that.

amir-qayyum-khan:
@noisecapella updated logic to avoid ajax call on close. I am not sure it is ok, please review it.
I am just backing up state in *Copy variables on first time load and when user try to exit i am comparing current state with previous one.

noisecapella:
It looks like this uses a message inside the panel instead of a popup. Can you change it to a popup like @ferdi mentioned in an earlier comment?

ui/jstests/test-learning-resource.jsx
```diff
@@ -681,8 +682,62 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+            refreshFromAPI={function() {}}
+            ref={this.props.afterMountCallback}/>;
+        }
+      });
+
+      var afterMount = function (component) {
+        component.setState({closeLRPanel: true}, function () {
+          assert.equal(component.state.closeLRPanel, true);
+        });
+      };
```
noisecapella: You are missing a `done` here. You can make two `done` functions by calling `assert.async()` twice.

ui/static/ui/js/learning_resources.jsx
noisecapella: Yes, you should be able to just assign the element to variables and reuse them that way.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -478,15 +532,18 @@ define('learning_resources',
...
+      if (isNewResource) {
+        // Unmount and remount the component to ensure that its state
+        // is always up to date with the rest of the app.
+        React.unmountComponentAtNode(container);
+      }
       React.render(<LearningResourcePanel
         repoSlug={repoSlug}
         learningResourceId={learningResourceId}
         refreshFromAPI={refreshFromAPI}
+        closeLRPanel={closeLRPanel}
```
noisecapella: Can you rename `closeLRPanel` to `requestedPanelClose` everywhere in this file? Otherwise it looks too similar to `closeLearningResourcePanel`

noisecapella: Actually see my comment below instead

noisecapella:
Instead of passing `closeLRPanel` into React components to indicate that we want to close them, we should pass the panel components a callback, `markDirty` for example, which says whether or not the panel contents have changed. The code in `listing.js` should listen for anything that would close the panel and if the dirty flag is set, pop up the dialog before closing the panel, else just close the panel.

amir-qayyum-khan:
Done @noisecapella please test

noisecapella:
This will require a messy rebase due to my reorganization in #821, sorry Amir!

ui/static/ui/js/learning_resources.jsx
```diff
@@ -476,19 +529,13 @@ define('learning_resources',
...
-      // is always up to date with the rest of the app.
-      React.unmountComponentAtNode(container);
-      React.render(<LearningResourcePanel
-        repoSlug={repoSlug}
-        learningResourceId={learningResourceId}
-        refreshFromAPI={refreshFromAPI}
-        closeLearningResourcePanel={closeLearningResourcePanel}
-        />, container);
+    loader: function (options, isNewResource, container) {
+      if (isNewResource) {
```
noisecapella: We don't need to know whether the resource is new or not, just whether it's dirty or not. I'm fine with the `options` change but we should remove `isNewResource` unless it's necessary for this task. If it's there only to save an API call, that should go in a separate PR

ui/static/ui/js/learning_resources.jsx
```diff
@@ -465,7 +516,9 @@ define('learning_resources',
     getInitialState: function () {
       return {
         description: "",
+        descriptionCopy: "",
```
noisecapella: Instead of `description` and `descriptionCopy` can you name these `originalDescription` and `description`? Otherwise it's not obvious which one a developer would use. It seems like the only use case of `originalDescription` is to compare with `description`, where `description` would be used everywhere else. Also rename `vocabulariesAndTerms` and `vocabulariesAndTermsCopy` the same way

ui/static/ui/js/learning_resources.jsx
```diff
@@ -390,6 +415,30 @@ define('learning_resources',
         thiz.setState({loaded: true});
       });
     },
+    componentWillReceiveProps: function(nextProps) {
+      var unSavedChangesMessage = "You have unsaved changes. " +
```
noisecapella: Can you use a dialog box for the message as @Ferdi mentioned earlier? I think this will also simplify the code in this class since the confirmation dialog code would be mostly outside this class, you wouldn't need to implement `componentWillReceiveProps` anymore.

ui/static/ui/js/learning_resources.jsx
```diff
@@ -378,9 +399,13 @@ define('learning_resources',
         thiz.setState({
           message: "Form saved successfully!"
         });
+
+        // Exit panel
         if (closePanel) {
           thiz.props.closeLearningResourcePanel();
         }
+
+        thiz.props.markDirty(false);
```
noisecapella: Can you move this inside `closeLearningResourcePanel()`? Once we decide to close the panel it would make sense to always clear the dirty flag there

ui/jstests/test-learning-resource.jsx
```diff
@@ -681,8 +688,65 @@ define(['QUnit', 'jquery', 'react', 'lodash', 'learning_resources',
...
+              refreshFromAPI: function() {},
+              markDirty: function(markDirty) { isMarkDirty = markDirty; },
+              closeLearningResourcePanel: function() {},
+              ref:afterMountTestDataChange
+            };
+            React.render(<LearningResourcePanel {...options} />, div);
+          });
+        });
+      };
+      var options = {
```
noisecapella: Why not specify these as React attributes?

ui/static/ui/js/learning_resources.jsx
```diff
@@ -390,6 +415,30 @@ define('learning_resources',
         thiz.setState({loaded: true});
       });
     },
+    componentWillReceiveProps: function(nextProps) {
+      var unSavedChangesMessage = "You have unsaved changes. " +
+        "Please save and exit.";
+
+      if (!_.eq(this.state.description, this.state.descriptionCopy)) {
```
noisecapella: We should do this comparison at the point that the state changes, `handleDescription` and all of the functions that update the terms.

amir-qayyum-khan:
Done @noisecapella 
screen shot 2015-11-17 at 7 13 59 pm

ui/static/ui/js/listing/listing.jsx
```diff
@@ -186,6 +191,36 @@ define('listing',
         };
         return React.createElement(ListingResources.ListingPage, options);
       },
+      componentDidUpdate: function () {
```
noisecapella: Why does this have to be implemented? It seems like you can move `closeLearningResourcePanel` inside this class and handle the dialog box completely within `closeLearningResourcePanel`

ui/static/ui/js/listing/listing.jsx
```diff
@@ -9,6 +9,7 @@ define('listing',
     'use strict';
 
     var EMAIL_EXTENSION = '@mit.edu';
+    var lastSelectedResourceId;
```
noisecapella: After `closeLearningResourcePanel` is moved inside the class you should be able to just use `this.state.currentResourceId` which already exists

ui/static/ui/js/learningresources/learning_resources.jsx
```diff
@@ -4,17 +4,20 @@ define('learning_resources',
...
-          />, container);
+      /** Current list of options are:
+        repoSlug
+        learningResourceId
+        refreshFromAPI
+        markDirty
+        closeLearningResourcePanel
+      */
+      loader: function (options, isNewResource, container) {
+        if (isNewResource) {
```
noisecapella: Why is this check necessary? Why not just always unmount if loader is called?

ui/static/ui/js/listing/listing.jsx
```diff
@@ -158,7 +161,9 @@ define('listing',
           // Keep track of resource id so we can lazy load panels.
           currentResourceId: undefined,
           // What panels are already loaded
-          loadedPanels: {}
+          loadedPanels: {},
+          isLearningResourcePanelDirty: false,
+          requestedPanelClose: false
```
noisecapella: If you handle the close logic inside `this.closeLearningResourcePanel` you shouldn't need this piece of state anymore

amir-qayyum-khan:
@noisecapella I am confuse. My logic is
- When user tries to close LR panel. I am sending new properties to LR panel without un-mounting it. This is to avoid rerunning `componentDidMount` call which updates state from server.
- By doing this when i call loader then control goes to `componentWillReceiveProps` in LR panel.
- Here i am checking the current state with original state (original state  saved in *original variables in  `componentDidMount`)
- If state is change then I am calling  `nextProps.markDirty(true);`
- After calling `nextProps.markDirty(true);`, state inside listing resource will be change afterword inside  `componentDidUpdate` I am checking if user's operation was close LR panel panel is marked dirty then show confirmation dialog else close panel.
- If I dont user [componentWillReceiveProps](https://github.com/mitodl/lore/pull/741/files#diff-9ee85d23b5ea32c8623184379671b176R186) where can i check that state is change from original state?
- if I dont user [isNewResource](https://github.com/mitodl/lore/pull/741/files#diff-f7f9a660eff4a5f7610ce3c1cdeda1ffR15) then how can i save original state so that i can compare new state vs original state? Because if i do unmount panel before mounting it then it will download and update state from server.

noisecapella:
> When user tries to close LR panel. I am sending new properties to LR panel

I think the panel shouldn't need the new properties when the user attempts to close the panel. When `this.props.closeLearningResourcePanel` is called it may immediately close the panel or otherwise pop up a dialog box for confirmation. But in either case the resource panel does not need any new props, at least in my understanding.

amir-qayyum-khan:
Done @noisecapella

ui/jstests/learningresources/test_learning_resource_panel.jsx
```diff
@@ -482,6 +487,7 @@ define(["QUnit", "react", "test_utils", "jquery", "lodash",
           <LearningResourcePanel
             repoSlug="repo"
             closeLearningResourcePanel={function() {}}
+            markDirty={function() {}}
```
noisecapella, line 614: Can you add another test which tests `markDirty`? The test can just make changes to the term state, and the description, and assert that the dirty flag was set properly.

ui/static/ui/js/learningresources/learning_resource_panel.jsx
```diff
@@ -204,6 +228,7 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
               message: undefined,
               description: description,
               previewUrl: previewUrl,
+              descriptionOriginal: _.cloneDeep(description),
```
noisecapella: It is not necessary to use `cloneDeep` here since Javascript strings are immutable

ui/static/ui/js/learningresources/learning_resource_panel.jsx
```diff
@@ -179,6 +194,15 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
           thiz.setState({loaded: true});
         });
       },
+      componentWillUpdate: function(nextProps, nextState) {
```
noisecapella: The logic here marks the dirty flag false and the logic in the callbacks above marks it true. Can we move this to a function like this:

```
updateDirty: function() {
    var isDirty = !_.eq(this.state.description, this.state.descriptionOriginal) || !_.eq(this.state.vocabulariesAndTerms, this.state.vocabulariesAndTermsOriginal);
    this.props.markDirty(isDirty);
}
```

Then just call `updateDirty` when there's any change to `vocabulariesAndTerms` or `description`

ui/static/ui/js/learningresources/learning_resource_panel.jsx
```diff
@@ -167,6 +180,8 @@ define("learning_resource_panel", ['react', 'jquery', 'lodash',
           thiz.setState({
             message: "Form saved successfully!"
           });
+          // user can close panel
+          thiz.props.markDirty(false);
```
noisecapella: `descriptionOriginal` and `vocabulariesAndTermsOriginal` should be updated here, it is not sufficient to just set the dirty flag to false. Please add the `updateDirty` method as described below, then replace this call with changes to `descriptionOriginal` and `vocabulariesAndTermsOriginal` and call `updateDirty`

amir-qayyum-khan:
Done @noisecapella

noisecapella:
Sorry for the long delay in reviewing your PRs @amir-qayyum-khan, I'll get to this soon. Thanks for your work!

ui/jstests/listing/test_listing.jsx
```diff
@@ -971,5 +971,146 @@ define(['QUnit', 'jquery', 'react', 'test_utils', 'utils', 'listing'],
...
+        '1/?remove_content_xml=true',
+        type: 'GET',
+        responseText: learningResourceResponse
+      });
+      TestUtils.initMockjax({
+        url: '/api/v1/repositories/test/vocabularies/?type_name=course',
+        type: 'GET',
+        responseText: vocabulariesResponseFirst
+      });
+      TestUtils.initMockjax({
```
noisecapella: Why not use the vocabulary data and mocks already defined at the top of the file?

ui/static/ui/js/listing/listing.jsx
```diff
@@ -187,6 +185,37 @@ define('listing',
...
+       * Use by close learning resource confirmation popup.
+       */
+      resetAndHideLearningResourcePanel: function() {
+        $('.cd-panel').removeClass('is-visible');
+        this.setState({
+          currentResourceId: undefined,
+          isLearningResourcePanelDirty: false
+        });
+      },
+      conformCloseLearningResourcePanel: function(status) {
```
noisecapella: Please rename to `confirmCloseLearningResourcePanel`

amir-qayyum-khan:
Done @noisecapella

noisecapella:
Thanks @amir-qayyum-khan. Remember to squash before you merge :+1:

PR #745, opened 2015-10-05
Updated status tests to use settings context

noisecapella:
This is necessary because during my work refactoring the elasticsearch PR the status test settings leaked into other tests which ran afterwards. I don't know why exactly but you can reproduce it by running the tests in PR #739 without this code

### DIFF
--- a/status/tests/test_status.py
+++ b/status/tests/test_status.py
@@ -34,29 +34,11 @@ class TestStatus(TestCase):
         self.url = reverse("status")
         super(TestStatus, self).setUp()
 
-        # Remember settings in case they're mutilated during a test.
-        self.original_settings = (
-            settings.BROKER_URL,
-            deepcopy(settings.DATABASES),
-            deepcopy(settings.HAYSTACK_CONNECTIONS),
-        )
-
-    def tearDown(self):
-        """Restore settings after each test."""
-        (
-            settings.BROKER_URL,
-            settings.DATABASES,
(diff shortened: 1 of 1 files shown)


noisecapella:
You can reproduce this locally by running these two tests in this order on master: `docker-compose run web tox -e py27 ./status/tests/test_status.py::TestStatus::test_broken_settings ./ui/tests/test_learningresources_views.py::TestViews::test_get_home`

The second test will fail due to the elasticsearch URL being set to `not a chance` which is done in `TestStatus`

ShawnMilo:
It might be cleaner to use `override_settings` or `modify_settings` as described [here](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#overriding-settings).

It's not obvious from that documentation whether you can chain the decorators to change multiple values at once, so if not it would probably be better to change the tests to check one server at a time. If that's the case, let me know if you want me to take over the branch, since I wrote these tests originally.

noisecapella:
The problem with using decorators is that the test modifies a piece of the `DATABASES` and `HAYSTACK_CONNECTIONS` dicts which we can't do cleanly outside of the function.

From what I can gather `modify_settings` only works with lists of values, not dictionaries, and nothing allows you to remove a setting entirely. I do use `self.settings` which uses `override_settings` under the hood so it avoids directly editing the setting values at least.

ShawnMilo:
I played around with it a bit and I understand your explanation now. It's great with the `self.settings` context manager. Good to merge.

PR #747, opened 2015-10-06
Fixed reindexing for edited vocabularies

### LINKED ISSUE #723: Terms in index not updated when vocabulary is edited
Introduced by #522, might get fixed by #718. If not we probably need to figure out a signal which handles this case to update the index

noisecapella:
Fixes #723


ShawnMilo:
Thumbs up.

PR #750, opened 2015-10-06
Added loader to listing, resource panel and taxonomy panel

### LINKED ISSUE #736: React components should load a spinner while data is loading
Some components takes a couple of seconds to load. Two examples:
- Listing page
- The list of vocabularies / terms when a specific learning resource is loaded in the sidebar.

It would be great to have a spinner while data is loading.

noisecapella:
Fixes #736


giocalitri:
It's good for me as it is

PR #753, opened 2015-10-07
Removed haystack

noisecapella:
Part of #680. Hoping this will make performance analysis easier

This does not do any refactoring of the facet counts JSON, that can happen in a future PR. This removes haystack plus a few other things:
- specify fields we want from elasticsearch to avoid getting content_xml blob when not needed
- add `search_type=count` to tell elasticsearch to only return aggregations
- cache search results inside `SearchResults` object to avoid duplicate calls
- add `id` to mapping, maybe this will improve performance over dynamic mapping? Not sure yet


noisecapella:
This should be ready for review now. I tried it on lore-branch-test and it seems reasonably speedy, hopefully when it gets pushed to lore-ci we'll see an obvious benefit.

search/utils.py
```diff
@@ -29,6 +30,7 @@
 INDEX_NAME = settings.HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"]
 URL = settings.HAYSTACK_CONNECTIONS["default"]["URL"]
 _CONN = None
+_CONN_VERIFY = False
```
ShawnMilo: nit: I think `_CONN_VERIFIED` is more descriptive

search/utils.py
```diff
@@ -385,7 +427,7 @@ def _create_mapping(conn):
         conn.indices.delete_mapping(index=INDEX_NAME, doc_type=DOC_TYPE)
 
     mapping = Mapping(DOC_TYPE)
-
+    mapping.field("id", "integer", index="not_analyzed")
```
ShawnMilo: Nice fix on indexing the ID, by the way. I'm 99% sure `index="not_analyzed"` is a no-op on an integer type, but it probably doesn't hurt anything to keep it there.

ShawnMilo:
Looks good to me, especially the removal of Haystack's signals and the reduction in query counts during tests. 

Tests have passed, so I'm good to merge unless you agree on my variable name nit and want to "fix" that. The comment about the `not_indexed` on an integer type you can take or leave; I think if you just leave out the `index` kwarg altogether it'll just do the right thing. But it's up to you since I believe the behavior won't change.

Issue #758, opened 2015-10-09
Vocabularies for all repositories are showing on list page for each repo.


ShawnMilo:
When a repository is viewed, along the left there are filters for vocabularies that exist in other repositories, even if they don't exist for the current repository.

ShawnMilo:
The front-end problem has been addressed in the short term by #759.

The description of this issue assumes cross-repository queries will not happen (they currently do not), but @Ferdi has some information which changes that. Please see him before working on this, unless he's added the new requirements as comments below.

ShawnMilo:
Updated this ticket to be only about the bug, not the mappings, since #759 refers to closing this one. #760 was created to track the separate mapping concern.

pdpinch:
@ShawnMilo @noisecapella can we close this now?

ShawnMilo:
Yes, #759 fixed this.

PR #759, opened 2015-10-09
Limited vocabularies to repository being searched

noisecapella:
Addresses #758

### DIFF
--- a/rest/tests/test_search.py
+++ b/rest/tests/test_search.py
@@ -27,8 +27,12 @@ class TestSearch(RESTTestCase):
     Tests for search endpoint.
     """
 
-    def get_results(self, query="", selected_facets=None, sortby=""):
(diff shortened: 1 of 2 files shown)


ShawnMilo:
To summarize the conversation in HipChat:

I like this, and it cleanly fixes the UI. I think we should still create a mapping per repository as in #758 soon to clean up the backend.

I'm going to test this locally and if it works as expected, I'll give it the thumbs-up.

noisecapella:
Added a test, up for review now

ShawnMilo:
Manually tested, works perfectly.

Test:
- Created two repos
- Added different vocabularies to each
- Tagged multiple items in each repo
- Imported new course into both repos
- Confirmed that both vocabularies appeared in both repos
- Replaced `master` with this branch
- Confirmed that only the appropriate vocabularies appeared on each repo

Good to merge after Travis.

PR #762, opened 2015-10-09
Added remove_content_xml flag and updated resource panel to use it

### LINKED ISSUE #761: Add remove_content_xml param to learning resource API
This param should remove the content_xml field from the information being returned if provided. It would enable better performance on the learning resource panel which currently may get megabytes of content_xml information before it can finish loading.

noisecapella:
Fixes #761

### DIFF
--- a/rest/views.py
+++ b/rest/views.py
@@ -550,6 +550,19 @@ class LearningResourceDetail(RetrieveUpdateAPIView):
         return LearningResource.objects.filter(
             id=self.kwargs['lr_id'])
 
+    def get(self, request, *args, **kwargs):
+        """Override get to filter on content_xml."""
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+
+        remove_content_xml = request.GET.get("remove_content_xml")
+        if remove_content_xml == "true":
+            data = dict(serializer.data)
+            del data['content_xml']
+        else:
+            data = serializer.data
+        return Response(data)
+
     def update(self, request, *args, **kwargs):
         """Override update to remove response."""
         super(LearningResourceDetail, self).update(request, *args, **kwargs)
--- a/rest/tests/test_learning_resource.py
+++ b/rest/tests/test_learning_resource.py
@@ -518,6 +518,33 @@ class TestLearningResource(RESTTestCase):
         self.assertEqual(self.resource.terms.count(), 0)
         self.assertEqual(resource2.terms.count(), 1)
 
+    def test_remove_content_xml(self):
+        """
+        Test behavior of remove_content_xml flag.
+        """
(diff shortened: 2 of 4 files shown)


giocalitri:
this needs to be rebased

rest/views.py
```diff
@@ -550,6 +550,19 @@ def get_queryset(self):
         return LearningResource.objects.filter(
             id=self.kwargs['lr_id'])
 
+    def get(self, request, *args, **kwargs):
+        """Override get to filter on content_xml."""
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+
+        remove_content_xml = request.GET.get("remove_content_xml", None)
```
giocalitri: you don't need to specify `None` here

giocalitri:
Just two minor things, but what's the purpose of this PR?

noisecapella:
It allows the resource panel to show up without downloading all of content_xml which is only used on the XML tab

noisecapella:
Any other comments @giocalitri?

giocalitri:
How to you get the XML then? Can you pull only that?

noisecapella:
There are three AJAX calls made for the resource panel, one for each tab. The XML tab would call without the `remove_content_xml` flag which would bring in everything. Since `content_xml` is much bigger than any other field we don't need a flag for only `content_xml`

giocalitri:
ok: I'll check functionality then

giocalitri:
I noticed that when you open the panel there are 4 ajax requests right away:

1- for the metadata without the xml
2- for the static assets
3- for the vocabularies
4- for the  metadata with the xml

I honestly think this approach is wrong, especially with the introduction of this PR. 
IMHO the right approach should be to perform just the ajax request -1 and -3 and perform the others only if the user opens the related tab.

screen shot 2015-10-16 at 4 01 09 pm

giocalitri:
other than the previous comment, this specific piece of functionality works fine.

noisecapella:
I think it's out of the scope of this issue since it's behavior that existed before but feel free to file another issue for it

giocalitri:
I actually do not think it is out of scope, because I do not understand why this PR is useful, then: is it because the first request is faster? But then we pay that with one more request for each panel...

giocalitri:
I think I already talked about that kind of behavior several times, but I'll check and open an issue

noisecapella:
It's useful because now it's only the XML tab which is blocked by the large AJAX request, the user can still interact with the other two panels in the meantime. It speeds up the user experience a little.

noisecapella:
Since you're opening an issue for this, am I fine to merge this?

giocalitri:
yep! go for it

PR #766, opened 2015-10-16
Wrapped course imports in transaction

### LINKED ISSUE #756: If course import fails the course is half imported
We should do imports in a transaction so we can cleanly roll back if they don't work.

noisecapella:
Fixes #756

### DIFF
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -8,12 +8,12 @@ from bs4 import BeautifulSoup
 from shutil import rmtree
 import logging
 from tempfile import mkdtemp
-from os.path import join, exists, isdir
+from os.path import join, exists
 from os import listdir
 
 from archive import Archive, ArchiveException
-from django.core.files import File
 from django.core.files.storage import default_storage
+from django.db import transaction
 from lxml import etree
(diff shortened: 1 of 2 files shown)


importer/api/__init__.py
```diff
@@ -109,7 +108,8 @@ def import_course_from_path(path, repo_id, user_id):
     )
     bundle.import_from_directory(path)
     static_dir = join(path, 'static')
-    course = import_course(bundle, repo_id, user_id, static_dir)
+    with transaction.atomic():
+        course = import_course(bundle, repo_id, user_id, static_dir)
```
giocalitri, line 112: this can be a lot of stuff in a single transaction... what's the impact on the memory for the DB?

noisecapella, line 112: I don't believe there would be any extra memory usage, I think transactions do their business in a separate blocks in the database and then switch over once the transaction is complete. I talked to @bdero and he was thinking that since we're using Postgres the bad experiences with transactions with edx may not apply.

noisecapella:
I've confirmed that the course is created within `import_course` so I'll merge this

PR #787, opened 2015-10-22
Fixed link click behavior for term edit and delete

### LINKED ISSUE #784: Edit and delete term links don't call event.preventDefault()

noisecapella:
Fixes #784


giocalitri:
I think this is good as is: do I need to check functionality?

noisecapella:
It would be good to check functionality just to make sure this does what it seems to, and it should be a quick check. You should be able to click the edit and delete term icons without a '#' showing up in the URL

giocalitri:
will do now

giocalitri:
Functionality is fine: feel free to merge

noisecapella:
Thanks for your diligence :)

PR #788, opened 2015-10-22
Fixed index mapping of terms and vocabularies

### LINKED ISSUE #779: Vocabularies with spaces and without terms show up incorrectly in listing
The vocabulary slug is shown instead of the vocabulary name. It looks like this is due to a for loop over terms in `convert_aggregate()` which will skip vocabularies that don't contain any terms.

### LINKED ISSUE #785: Renamed term doesn't update mapping properly
To reproduce, add a term, assign a term to a resource, then rename the term. There are a couple things to notice:
- the old term name is used in the facet counts
- When searching the index only knows about the old term, resources are only updated if manually resaved.

One way to fix this is to use term ids as values for the mapping instead of slugs since they don't change

### LINKED ISSUE #786: Editing a vocabulary doesn't update mapping properly
Unlike in #785 you do see the updated vocabulary name in the facets but we only reindex resources where we would have removed terms from the vocabulary, so you will see that the new name doesn't have any facet counts for its terms even though they should exist.

If we used vocabulary ids here, something like 'vocab_1' for the index key, I think this problem would be solved since ids are immutable.

noisecapella:
Fixes #786 
Fixes #785 
Fixes #779

### DIFF
--- a/rest/tests/test_vocabulary.py
+++ b/rest/tests/test_vocabulary.py
@@ -24,7 +24,7 @@ from rest.serializers import (
     VocabularySerializer,
     TermSerializer,
 )
-from taxonomy.models import Vocabulary, Term
+from taxonomy.models import Vocabulary, Term, make_vocab_key
 from learningresources.models import LearningResourceType, LearningResource
 
 log = logging.getLogger(__name__)
@@ -668,8 +668,8 @@ class TestVocabulary(RESTTestCase):
                     self.get_results()['facet_counts'][vocab_key]['values']]),
             []
         )
-        self.assert_results([], vocab_key, term1.slug)
-        self.assert_results([], vocab_key, term2.slug)
+        self.assert_results([], vocab_key, term1.id)
+        self.assert_results([], vocab_key, term2.id)
 
         resource1 = LearningResource.objects.all()[0]
         resource2 = LearningResource.objects.all()[1]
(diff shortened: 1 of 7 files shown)


rest/tests/test_vocabulary.py
```diff
@@ -896,6 +896,134 @@ def test_vocab_num_queries(self):
...
+        vocab_id = vocab_result['id']
+        vocab_key = make_vocab_key(vocab_id)
+
+        self.assertEqual(
+            self.get_results()['facet_counts'][vocab_key]['facet']['label'],
+            'name with spaces'
+        )
+
+        self.create_term(self.repo.slug, vocab_slug)
+        self.assertEqual(
```
ShawnMilo: What is this assertion supposed to test? It's the same as the one just above (line 1016), and I'm missing how adding a term could change the output of `get_results()`.

ShawnMilo:
Very minor nit: I think `"brand new name"` should be a variable instead of repeating the string.

ShawnMilo:
One tiny nit and one question. The rest looks good.

rest/tests/test_vocabulary.py
noisecapella: There was a bug where if a vocab had no terms the vocab slug would show up in place of the vocab label (#779)

ShawnMilo: Okay, I didn't make the association. Thanks. It's good to merge, if you want to ignore the nit.

noisecapella: Oh, I guess this assertion doesn't test that, it's the one above that does.

PR #791, opened 2015-10-22
Added custom slugify function to allow any name for Repo, Vocab, Term

### LINKED ISSUE #773: User can create empty slug using a character like '('
If something which uses slugs, like a repository, vocabulary, or term, uses invalid characters for a name or label, the newly generated slug will be an empty string. This can cause many problems and we should change our slug generation methods to add some prefix to the slug to guard against this case.

I think this is a severe bug but not enough to delay the next release

giocalitri:
With this PR it will be possible to use names like "%^*&^" for Repo, Vocab and Term

Fixes #773

### DIFF
--- a/rest/util.py
+++ b/rest/util.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 from django.contrib.auth.models import User
 from django.http.response import Http404
 from django.shortcuts import get_object_or_404
+from django.utils.text import slugify
 from rest_framework.fields import BooleanField, empty
 
 from roles.permissions import BaseGroupTypes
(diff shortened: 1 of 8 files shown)


rest/util.py
```diff
@@ -60,3 +61,19 @@ def dispatch(self, request, *args, **kwargs):
             CheckValidMemberParamMixin,
             self
         ).dispatch(request, *args, **kwargs)
+
+
+def default_slugify(label, default_name, exists_func):
```
noisecapella, line 66: I'm not sure on the usefulness of `default_name` here. Why not just use 'slug' as a default term no matter where the slug is assigned? No strong feelings however

giocalitri, line 66: the reason was because it looked weid to me to have a term and a repository have the same base name.

rest/util.py
```diff
@@ -60,3 +61,19 @@ def dispatch(self, request, *args, **kwargs):
...
+    """
+    Function that extends the Django `slugify` to add a default string in case
```
noisecapella, line 68: The arguments and return value need documentation here

giocalitri, line 68: ok

noisecapella:
This needs tests for repository slugs, other than that everything looks fine

giocalitri:
Comments addressed

noisecapella:
I forgot to mention this earlier but we should also have migrations to rename existing empty slugs

giocalitri:
ok. I'm starting to work on it

giocalitri:
Migrations added

justinabrahms:
Looks pretty good to me, though I don't have much context on the overarching code base. I agree with @noisecapella that the API is a little awkward to pass a default name in, but that it's also not a huge deal. 

Good stuff! :smile_cat:

noisecapella:
Migrations work for me, feel free to merge. You'll have to repush due to Travis intermittent failures

giocalitri:
@noisecapella you do not need to re-push to restart travis: if you login in travis (with the github credentials) you can see a button to restart the job.

giocalitri:
thanks, @justinabrahms

giocalitri:
Tests passed: merging

PR #793, opened 2015-10-22
Set Elasticsearch log level higher during testing

### LINKED ISSUE #792: Tests crash due to memory use under limited memory Docker container
This is consistently reproduceable, you should see a sudden exit from tox and a non-zero error code. We likely won't see this on Travis due to its much higher memory limits, but it would be good to investigate and fix in case it's an indicator of some bigger problem

noisecapella:
Fixes #792. It looks like the increased memory use is just due to Elasticsearch logs being stored in memory.


giocalitri:
I think that if you have `docker-compose up` running in one shell and run `tox` in another, this should not happen because the ES logs are showed in the command line. 
Anyway, I think this is good as is.

PR #801, opened 2015-10-26
Moved content_xml to separate model

### LINKED ISSUE #800: Create LearningResourceContentXml table, move content_xml field there leaving original blank
I think this would help in making our LearningResource queries perform well without explicitly opting out of the `content_xml` field on every use. The REST API should return the same stuff as before to lower the amount of work involved, and since the learning_resources API already has parameters to remove this field.

I'm not sure how this obeys the "old code works with new data" rule, it shouldn't break anything on a schema level at least.

noisecapella:
Fixes #800 

I'm not sure if the performance gain is worth the extra complexity, I'll do some performance comparisons before marking it 'Needs Review'

### DIFF
--- a/learningresources/api.py
+++ b/learningresources/api.py
@@ -16,7 +16,8 @@ from django.db import transaction
 from guardian.shortcuts import get_objects_for_user, get_perms
 
 from learningresources.models import (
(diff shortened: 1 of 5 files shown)


noisecapella:
I'm thinking a piecemeal approach to this might be better for now for simplicity's sake.

Issue #111, opened 2015-06-05
As a curator, I'd like to know the state of my import


pdpinch:
Large uploads can take a while. We need to track the state of an upload, and create a ui to update the user on status. 
- [ ] determine the states of an import: upload, extract, 
- [ ] REST end point that returns status
- [ ] UI that polls status and displays to user (on the listing page?)

Since imports can currently take hours, the user should still be returned to listing page.

pdpinch:
Add a model to the database to track status. Celery's built-in status is limited to pending/accepted/error

carsongee:
Did we want to keep this information forever, or would tracking tasks in the session be ok/preferred?

carsongee:
Also, what do we want to use for the javascript? backbone, jquery, react, ember? Also, where should a view that returns json about the current import task be? If `ui`, no problem, if it is `rest`, that may be a bit tricky, because that hasn't landed yet.  It seems to me it should be in the `rest` app, can we maybe just PR the piece of that that adds the folders and urls, since #106 is large, unless that is expected to merge very soon

pdpinch:
As an example, here's the UI from edX studio import:

![screen shot 2015-06-16 at 8 43 08 pm](https://cloud.githubusercontent.com/assets/430126/8197628/7c7fcab2-1468-11e5-8657-c5d7296d12ac.png)

Ferdi:
@carsongee 

> Did we want to keep this information forever, or would tracking tasks in the session be ok/preferred?

I don't see a reason to keep that info around.

> where should a view that returns json about the current import task be? 

It's not part of our public rest api so UI would be fine

> Also, what do we want to use for the javascript? backbone, jquery, react, ember?

If jquery works , that's fine. We are going to use react. Maybe you can start using it first here.

ShawnMilo:
I just tried to import a course on the Heroku site. It failed (invisibly, of course), because of my permissions to the repository. I was able to import a course to a repository I had created.

Should this be considered part of this issue, or should I create a new issue?

pdpinch:
Can you open a new issue please? 

Definitely the user needs to know that the import errored, but I think there's something else that's not quite right in what you describe -- like how did you get to be in the position of importing a course into a repository you don't have permissions on?

carsongee:
We have not yet been able to reproduce this.  We thought we had with some unexpected permission denied messages, but those were due to #207

PR #802, opened 2015-10-26
Added tasks API

noisecapella:
Required for #111 

This adds `/api/v1/tasks/` to provide a place to queue different types of tasks. Tasks have a `task_type` to distinguish them, currently there's only `resource_export`.


jamiefolsom:
- [x] Functionality
- [x] Code
- [x] Tests
- [x] Questions

jamiefolsom:
@noisecapella what would be the best way to test functionality? I've got it running in my dev environment, and a GET request to: /api/v1/tasks works fine; should I just have a look at OPTIONS, and try POST, DELETE, PUT? Are there other endpoints? Thanks.

jamiefolsom:
Nevermind -- just found the docs!

noisecapella:
Test that the exports still work, and do it multiple times to make sure the old one is cleared properly. The exports UI was switched to use /api/v1//tasks/

jamiefolsom:
OK, so the end user experience should be unchanged, correct? That appears to be the case after a first pass. I will hammer on it some more. Thanks.

noisecapella:
Yes, everything should behave like it did before

jamiefolsom:
OK, yep, confirmed -- that was the easy part I guess.

rest/tests/base.py
```diff
@@ -772,11 +772,58 @@ def create_learning_resource_export(self, repo_slug, input_dict,
...
+            return as_json(resp)
+
+    def delete_task(self, task_id, expected_status=HTTP_204_NO_CONTENT):
+        """
+        Delete a task and remove it from task list.
+        """
+        resp = self.client.delete('/api/v1/tasks/{task_id}/'.format(
+            task_id=task_id
+        ))
+        self.assertEqual(expected_status, resp.status_code)
```
jamiefolsom, line 820: Should this test also assert that the new list of tasks is actually the one we expect, ie that it doesn't contain the one which we intended to delete, but that it does contain the rest? Overkill?

noisecapella, line 820: That's a good idea, I'll add this

rest/views.py
```diff
@@ -815,54 +789,34 @@ def post(self, request, *args, **kwargs):
...
-        # Put new task in session.
-        self.request.session[EXPORT_TASK_KEY][repo_slug][result.id] = {
-            "id": result.id,
-            "initial_state": result.state,
-            "url": url,
-            "collision": collision
-        }
-        self.request.session.modified = True
+        for task in export_tasks:
+            remove_task(self.request.session, task['id'])
```
jamiefolsom, line 801: Nit: would it be simpler just to remove the TASK_KEY from the session, rather than deleting each task?

ui/jstests/test_lr_exports.jsx
```diff
@@ -48,7 +52,7 @@ define(['QUnit', 'jquery', 'lr_exports', 'react',
         }
       });
       TestUtils.initMockjax({
-        url: '/api/v1/repositories/repo/learning_resource_export_tasks/',
+        url: '/api/v1/tasks/',
```
jamiefolsom, line 55: Big improvement in the API, in my opinion!

rest/views.py
noisecapella, line 801: The user may have an export task for another repository in progress (unlikely but possible)

jamiefolsom, line 801: Good point.

jamiefolsom:
:+1: from my point of view. I have a problem with my dev environment w/r/t pytest, but since travis gives it the green light, they are clearly passing. If you want another set of eyes on this of course, by all means.

rest/tests/base.py
noisecapella, line 820: This is actually a helper function, the test using it already checks for successful deletion

PR #810, opened 2015-10-28
Implemented import status

### LINKED ISSUE #111: As a curator, I'd like to know the state of my import

noisecapella:
Fixes #111

### DIFF
--- a/ui/static/ui/js/listing/import_status.jsx
+++ b/ui/static/ui/js/listing/import_status.jsx
@@ -1,0 +1,130 @@
+define("import_status", ["React", "lodash", "jquery", "utils",
+  "react_spinner", "status_box"],
+  function (React, _, $, Utils, ReactSpinner, StatusBox) {
+    'use strict';
+
+    return React.createClass({
+      getInitialState: function () {
+        return {
+          importTasks: {},
+          hasRefreshed: false
+        };
+      },
+      updateImportStatus: function () {
+        var thiz = this;
+
+        Utils.getCollection("/api/v1/tasks/").then(function (tasks) {
+          // Only deal with tasks on this repo that deal with imports.
+
+          tasks = _.filter(tasks, function(task) {
+            return (task.task_type === 'import_course' &&
+              task.task_info.repo_slug === thiz.props.repoSlug);
+          });
+          var tasksMap = {};
+          _.each(tasks, function (task) {
+            tasksMap[task.id] = task;
+          });
+          thiz.setState({importTasks: tasksMap});
+
+          var notProcessing = _.filter(tasks, function (task) {
+            return task.status === 'success' || task.status === 'failure';
+          });
+
+          var deferred = $.Deferred();
+
+          // Delete one after another to avoid race conditions with
+          // Django session.
+          _.each(notProcessing, function (task) {
+            deferred.then(function () {
+              return $.ajax({
+                method: 'DELETE',
+                url: '/api/v1/tasks/' + task.id + '/'
+              });
+            });
+          });
+          deferred.resolve();
+
+          var numProcessing = _.filter(tasks, function (task) {
+            return task.status === 'processing';
+          }).length;
+          if (numProcessing > 0) {
+            thiz.setState({hasRefreshed: true});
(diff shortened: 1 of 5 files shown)


ui/static/ui/js/listing/import_status.jsx
```diff
@@ -0,0 +1,130 @@
...
+          // Django session.
+          _.each(notProcessing, function (task) {
+            deferred.then(function () {
+              return $.ajax({
+                method: 'DELETE',
+                url: '/api/v1/tasks/' + task.id + '/'
+              });
+            });
+          });
+          deferred.resolve();
```
giocalitri, line 45: what if some call fails?

ui/static/ui/js/listing/import_status.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+          var numProcessing = _.filter(tasks, function (task) {
+            return task.status === 'processing';
+          }).length;
+          if (numProcessing > 0) {
+            thiz.setState({hasRefreshed: true});
+            setTimeout(thiz.updateImportStatus, 3000);
```
giocalitri, line 52: Nit: I would make this timeout a bit more visible, maybe adding a comment or moving it into a variable with a meaningful name.

ui/static/ui/js/listing/import_status.jsx
```diff
@@ -0,0 +1,130 @@
...
+          return task.status === 'processing';
+        }).length;
+
+        var importWord = function (n) {
+          if (n === 1) {
+            return "import";
+          } else {
+            return "imports";
+          }
+        };
```
giocalitri, line 78: extra minor nit: you can just append the "s" to the word

ui/static/ui/js/listing/import_status.jsx
noisecapella, line 45: The code above attaches handlers for the success case of `deferred` but doesn't actually resolve or reject the initial promise until this line. We need to resolve the initial promise so that it can start evaluating the other success callbacks. To illustrate:

```
deferred -> task1 -> task2 -> task3
```

`deferred` is always resolved. It doesn't do anything but we need to attach the task chain to something we have control of. Then `task1` is evaluted. It returns a promise that's either rejected or resolved depending on how the AJAX request went. If `task1` succeeded `task2`'s success handler is then executed.

If `task1` failed it would try to execute `task2`'s failure handler, or `task3`, or whatever the next in line with a failure handler is. None of them have failure handlers so we just stop executing the chain in the event of a AJAX failure.

ui/static/ui/js/listing/import_status.jsx
noisecapella, line 78: It would work here but string concatenation for plurals should be avoided in general since it complicates localization

ui/static/ui/js/listing/import_status.jsx
giocalitri, line 45: ok, got it

giocalitri:
The UI does not look great:
screen shot 2015-11-12 at 11 53 12 am

@pdpinch ?

ui/static/ui/js/listing/import_status.jsx
giocalitri, line 78: I agree:

giocalitri:
There is no way to dismiss the message:
screen shot 2015-11-12 at 12 03 40 pm

giocalitri:
Only two comments about the UI and one minor in the code.

pdpinch:
Hi George. How close is this to being done? All things being equal I'd like to include it in the next release, but it's not critical if it's still need a lot of work.

noisecapella:
@giocalitri uncovered an edge case where the imports in progress wouldn't disappear properly if the Celery server had stopped the worker. He also wanted changes to the UI as mentioned above. I don't think there's a lot of work left but it's not quite there yet

PR #817, opened 2015-10-30
Renamed listing.js to JSX

### LINKED ISSUE #687: Use JSX for listing.js
This would let us get rid of the loader functions and use React components directly from within listing.jsx.

noisecapella:
This just allows us to use the JSX syntax within listing.js (renamed to listing.jsx)

Fixes #687


ui/templates/repository.html
```diff
@@ -66,6 +65,9 @@
 <script type="text/jsx"
         src="{% static "ui/js/utils.jsx" %}">
 </script>
+<script type="text/jsx"
+        src="{% static "ui/js/listing.jsx" %}">
```
jamiefolsom, line 69: Is this indentation standard? Looks like 8 spaces.

noisecapella, line 69: It matches everything else in the file, but there's no particular standard for HTML indentation in LORE

jamiefolsom, line 69: Oh, I see the attribute names are horizontally aligned.

jamiefolsom, line 69: Looks fine to me

PR #818, opened 2015-10-30
Bind demo branch to lore-demo app deployment

bdero:
So we can collaboratively evaluate lore-demo releases using pull requests. @pdpinch


bdero:
I've just set the hash of `demo` to be the current commit deployed on lore-demo: https://github.com/mitodl/lore/tree/demo

pdpinch:
:shipit:

PR #821, opened 2015-10-30
Moved React components to separate files in new directories

### LINKED ISSUE #816: Move React classes into separate files

noisecapella:
Fixes #816


noisecapella:
This PR is missing manage_taxonomies.jsx which will be handled in a separate PR

noisecapella:
There are now two commits, one which moves tests to separate files and the other which moves implementation code to separate files

jamiefolsom:
Safe to say that exercising the functionality and the test suite would be enough to confirm that this PR is good to merge? In other words, this is pure refactoring, no new or removed features, right?

noisecapella:
Yes, no changes should be noticed. I took it out of 'Needs review' by the way, I'm moving a few more things around

jamiefolsom:
Roger.

noisecapella:
The split up of files slows down Django Compressor, at least locally. If I can't find an easy workaround we may want to just split the tests and leave the implementation files as is.

noisecapella:
This should not have the slowdown anymore. Ready for review @jamiefolsom

jamiefolsom:
@noisecapella what change did you make to address the slowdown?

noisecapella:
There's now a cache for the React transform (#827) so you should only get the slowdown on first load

jamiefolsom:
- [x] Functionality
- [x] Code
- [x] Tests
- [x] Questions

jamiefolsom:
@noisecapella this looks fine to me. The one thing I would suggest is that the nomenclature be made even more consistent. 

It's probably not warranted to hold up this PR, but a good example from my point of view is in `taxonomy/`, where you've got files named: `vocabulary_component.jsx` and `add_vocabulary.jsx`: very clear! 

On the other hand, In `utils/`, and `learning_resources/`, the files are named in a wider range of conventions, like `select2_component.jsx` vs `status_box.jsx` and there's even a `utils/utils.jsx`. 

It's not a big deal, but maybe we adopt a naming convention before our next big react project (_cough cough_)?

Otherwise, functionality works fine, tests pass, no questions: :+1:

noisecapella:
Yeah, the naming needs work. `lr_exports` is named like that because `exports` is a reserved keyword and Requirejs has trouble with it. select2_component is named to avoid collision with the select2 module being imported. Requirejs does allow you to just use file paths and we should probably do just that for the most part in the next project

PR #826, opened 2015-11-03
Removed django-compressor-requirejs requirement

### LINKED ISSUE #823: Remove django-compressor-requirejs
We don't use this dependency so we should remove it

noisecapella:
Fixes #823

### DIFF
--- a/lore/settings.py
+++ b/lore/settings.py
@@ -210,7 +210,6 @@ STATICFILES_DIRS = (
     os.path.join(BASE_DIR, 'ui', 'static'),
 )
 COMPRESS_PRECOMPILERS = (
-    ('text/requirejs', 'requirejs.RequireJSCompiler'),
     (
         'text/jsx',
         'node node_modules/react-tools/bin/jsx < {infile} > {outfile}'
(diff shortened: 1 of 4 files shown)


bdero:
Looks like the test failure is unrelated and due to list ordering differences: https://travis-ci.org/mitodl/lore/builds/89042237#L342

noisecapella:
I filed #828 for it, for this branch I'll repush and hope it goes away

bdero:
Cool, if it doesn't pass, maybe we can implement this: https://github.com/auchenberg/volkswagen

noisecapella:
Or this for Python: https://pypi.python.org/pypi/pytest-vw

noisecapella:
I spent most of today trying to investigate the test failure. The root of the problem is that we never install the JSX transformer in Travis to use in the Python tests. If I delete `node_modules` locally I don't get the error but it looks like if `DEBUG` is false the exception is suppressed and no compress happens, the non-transformed script tags are returned instead.

I don't know why Travis is different than my local machine in that regard and I don't know how removing `django-compressor-requirejs` causes this, but it seems like the solution is to run `npm install` on tox for the Python tests to populate the `node_modules` directory. This should only add a half second or so if the directory already exists.

noisecapella:
Any other comments @bdero?

bdero:
:+1:

Issue #814, opened 2015-10-30
Bad preview links in production


pdpinch:
Reported by Chen in https://odl.zendesk.com/agent/tickets/2557

> The "Missing Title" issue might be more complicated than that. Try searching for "angular momentum homework" in 8.MReVx, and looked for problems. On the first page you'll see something like this:
> 
> If you click preview on the "Missing Title" ones, they'll actually bring you to a blank course page. The vertical "checkpoint 2: Momentum of a System" only contains two problems, both have names and are found. The two "Missing Title" ones are actually ghost components that don't exist (or at least invisible to me).

In my investigation, it appears that some of the (problem) learning resources have bad preview links that are missing the `/jump_to_id/` followed by `url_name`. 

For example, from this search:

http://lore.odl.mit.edu/repositories/classical-mechanics/?selected_facets=course_exact%3A8.MReVx&selected_facets=resource_type_exact%3Aproblem

these resources have bad preview URLs:

http://lore.odl.mit.edu/api/v1/repositories/classical-mechanics/learning_resources/21192/
http://lore.odl.mit.edu/api/v1/repositories/classical-mechanics/learning_resources/21271/

I can't imagine why, but it seems that every resource with a missing title has a bad preview URL.

pwilkins:
I'm finding this difficult to test because I don't have a good way to identify the same learning resource in different repositories.  I can reproduce the problem in production, and so far I haven't been able to reproduce it on lore-ci.  @pdpinch @noisecapella

noisecapella:
It's difficult to test because there are 3 PRs which may overlap in solving the problem. Currently on lore-ci you should be able to import Classical Mechanics and the preview link should be fixed. However when PR #835 is merged you won't even see these extra resources anymore after importing.

pdpinch:
Can you load 8.MReV in lore-rc?

pwilkins:
I will do that now.

pwilkins:
@pdpinch 8.MReV is loaded in lore-rc in the 'release-testing' repository.

pdpinch:
Here is a resource in 8.MReV that should not have been created:

https://lore-rc.herokuapp.com/repositories/release-testing/?selected_facets=course_exact%3A8.MReVx https://lore-rc.herokuapp.com/repositories/release-testing/?selected_facets=course_exact:8.MReVx

It has a bad Preview link. 

Here is a resource from CC.801 that is legit:

[...]

PR #834, opened 2015-11-04
Added parsing for display_name so we can use it if url_name isn't present

noisecapella:
Fixes part of #814. It makes it more likely that the preview links will have something relevant to work with


ShawnMilo:


PR #835, opened 2015-11-04
Fixed double resource import bug

noisecapella:
Fixes part of #814. In that issue there are two `<problem>` tags, one nested in the other. The inner resource has no url_name so its preview link would be broken. This PR prevents the second problem tag from being imported (it's still present in the content_xml)

### DIFF
--- a/importer/api/__init__.py
+++ b/importer/api/__init__.py
@@ -144,6 +144,19 @@ def import_course(bundle, repo_id, user_id, static_dir):
     return course
 
 
+def is_leaf_tag(tag):
+    """
+    Should we look for resources within elements with this tag?
+
+    Args:
+        tag (unicode): Element tag
+    Returns:
+        bool: Whether tag is leaf tag
+    """
+    return tag in {'video', 'html', 'problem', 'discussion'}
+
+
+# pylint: disable=too-many-branches
 def import_children(course, element, parent, parent_dpath):
     """
     Create LearningResource instances for each element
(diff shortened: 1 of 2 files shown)


importer/api/__init__.py
```diff
@@ -144,6 +144,19 @@ def import_course(bundle, repo_id, user_id, static_dir):
...
+def is_leaf_tag(tag):
+    """
+    Should we look for resources within elements with this tag?
+
+    Args:
+        tag (unicode): Element tag
+    Returns:
+        bool: Whether tag is leaf tag
+    """
+    return tag in ('video', 'html', 'problem', 'discussion')
```
ShawnMilo: ultra-nit: This should be a `set()` instead of a tuple.

importer/api/__init__.py
```diff
@@ -230,6 +243,10 @@ def import_children(course, element, parent, parent_dpath):
         ]
     )
 
-    for child in element.getchildren():
-        if child.tag in DESCRIPTOR_TAGS:
-            import_children(course, child, resource, dpath)
+    # Try to protect against bad data, specifically <problem><problem>...
+    # imports. The two tags will still appear in content_xml but there will
+    # be only one resource for the outer one.
+    if not is_leaf_tag(element.tag):
```
ShawnMilo, line 249: Preference (almost-a-nit).

Change this to:

``` python
if is_leaf_tag(element.tag):
    continue
```

That way it's easier to read, and there's no unnecessary indentation of the important and most-commonly-executed logic.

importer/api/__init__.py
noisecapella: I'm not sure it's more performant for four elements but it's probably cleaner. By the way this can be formatted this as a set literal instead of using `set()`: `{'video', 'html', 'problem', 'discussion'}`

ShawnMilo: Yeah, a set literal is great. It doesn't have to be `set()` -- I was just referring to a set.

And yeah, it's only four elements. That's why I said "ultra-nit." 😉

importer/api/__init__.py
noisecapella, line 249: The outer block is a if statement so the continue wouldn't work here

ShawnMilo, line 249: Yep. My bad. It was a nit anyway.

PR #836, opened 2015-11-04
Fixed preview link to use parent resource link if child has none

### LINKED ISSUE #814: Bad preview links in production

noisecapella:
Fixes #814

### DIFF
--- a/importer/tests/test_import.py
+++ b/importer/tests/test_import.py
@@ -10,22 +10,30 @@ from shutil import rmtree
 from tempfile import mkstemp, mkdtemp
 import zipfile
 import logging
+from lxml import etree
 
 from django.core.files.storage import default_storage
 from django.core.files.base import ContentFile
 import mock
+from xbundle import XBundle
 
 from importer.api import (
     import_course_from_file,
     import_course_from_path,
     import_static_assets,
+    import_course,
 )
 from importer.tasks import import_file
-from learningresources.api import get_resources
+from learningresources.api import get_resources, create_repo
 from learningresources.models import (
-    Course, StaticAsset, static_asset_basepath, LearningResource,
+    Course,
+    StaticAsset,
+    static_asset_basepath,
+    LearningResource,
+    get_preview_url,
 )
 from learningresources.tests.base import LoreTestCase
+from lore import settings
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


learningresources/models.py
```diff
@@ -216,12 +216,25 @@ def get_preview_url(resource, org=None, course_number=None, run=None):
         run=run,
     )
 
-    if resource.url_name is None:
+    url_name = None
```
ShawnMilo: nit:

Lines 219 - 221 can just be `url_name = resource.url_name`, and line 222 can be `if url_name is None`. The `else` makes it _slightly_ more difficult to follow.

importer/tests/test_import.py
```diff
@@ -291,3 +299,46 @@ def get_counts():
...
+        self.assertEqual(html_resources.count(), 1)
+        html_resource = html_resources.first()
+        self.assertEqual(
+            get_preview_url(html_resource),
+            "{base}courses/{org}/{course}/{run}/jump_to_id/{url_path}".format(
+                base=settings.LORE_PREVIEW_BASE_URL,
+                org=html_resource.course.org,
+                course=html_resource.course.course_number,
+                run=html_resource.course.run,
+                url_path="2015_Summer"
```
ShawnMilo, line 364: This test is good. Do you think having a test for the case of no title anywhere in the tree would be beneficial?

noisecapella, line 364: It's a good idea, I'll add it

noisecapella:
Your comments should be addressed now, anything else?

ShawnMilo:
No, nothing else. 

PR #839, opened 2015-11-04
Fixed elasticsearch in docker-compose

### LINKED ISSUE #838: Docker setup is not functioning properly for people setting up from scratch
It looks like elasticsearch needs a setting to tell it to listen to all IP addresses. Since we only use Docker for development I don't see any security risk in making this change.

noisecapella:
Fixes #838

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,7 @@ redis:
     - "6379"
 elastic:
   image: elasticsearch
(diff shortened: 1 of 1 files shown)


noisecapella:
Assigning to @giocalitri since he ran into this problem

giocalitri:
:+1: fixes the problem. Merging

PR #853, opened 2015-11-25
Release 0.14.1

pwilkins:
Release v0.14.1
- Made `requires.io` dependency update.
- Added manual test plan.
- Made changes to accommodate Docker 1.9.0.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,13 @@
 Release Notes
 -------------
 
+Version 0.14.1
+==============
+
+- Made ``[requires.io]`` dependency update.
+- Added manual test plan.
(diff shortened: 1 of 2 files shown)


bdero:
It looks like these changes already got pushed to the release-candidate branch.

pdpinch:
I think that's been our process -- is that a problem?

bdero:
Nope, but what's this PR for?

bdero:
Ah, so the process is that the `rc/` branch is merged into `master` rather than `release` into `master`? We might want to merge `release` into `master` after `release` is verified deployed instead - just so we can catch any possible deviation between `release` and `master` in the PR.

PR #857, opened 2015-11-30
Fix new pylint 1.5.0 violations

bdero:


bdero:
Fun stuff. ;)
screen

roles/tests/test_user_models.py
```diff
@@ -49,8 +49,8 @@ class FooObject(object):
             def __init__(self, foo_var):
                 self.foo_var = foo_var
 
-        user_group2 = FooObject('bar')
-        self.assertFalse(user_group1.__eq__(user_group2))
+        foo_object = FooObject('bar')
```
noisecapella, line 52: Why is this change necessary? Just a better name?

noisecapella:
I left a comment but it's not important :+1:

roles/tests/test_user_models.py
bdero, line 52: pylint now errors when it sees a variable reassignment with a type different than the initial assignment type. Also, I couldn't name it `baz` because that's a "forbidden identifier". :)

bdero:
:8ball:

PR #872, opened 2016-01-22
Added sorting by relevance (_score)

### LINKED ISSUE #868: Change default sort order to include score
Currently sort defaults to id number, which makes sense for searches with no keyword. 

We can change the default to be score, then id. For searches with no keyword, presumably the scores will all be the same, and the effective sort will be by id. 

related to #866

giocalitri:
Fixes #868

### DIFF
--- a/search/sorting.py
+++ b/search/sorting.py
@@ -14,8 +14,9 @@ class LoreSortingFields(object):
     SORT_BY_NR_ATTEMPTS = ('nr_attempts', 'Number of Attempts (desc)', '-')
     SORT_BY_AVG_GRADE = ('avg_grade', 'Average Grade (desc)', '-')
     SORT_BY_TITLE = ('titlesort', 'Title (asc)', '')
+    SORT_BY_RELEVANCE = ('_score', 'Relevance', '')
 
-    DEFAULT_SORTING_FIELD = SORT_BY_NR_VIEWS[0]
+    DEFAULT_SORTING_FIELD = SORT_BY_RELEVANCE[0]
 
     # base sorting field in case the applied sorting is working on equal values
     BASE_SORTING_FIELD = 'id'
@@ -29,7 +30,8 @@ class LoreSortingFields(object):
             cls.SORT_BY_NR_VIEWS,
             cls.SORT_BY_NR_ATTEMPTS,
             cls.SORT_BY_AVG_GRADE,
-            cls.SORT_BY_TITLE
+            cls.SORT_BY_TITLE,
+            cls.SORT_BY_RELEVANCE,
         ]
 
     @classmethod
--- a/search/utils.py
+++ b/search/utils.py
@@ -6,7 +6,7 @@ from __future__ import unicode_literals
 
 from collections import defaultdict
 import logging
-from itertools import islice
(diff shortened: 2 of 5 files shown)


search/utils.py
```diff
@@ -168,21 +169,35 @@ def search_index(tokens=None, repo_slug=None, sort_by=None, terms=None):
         search = search.query("match", repository=repo_slug)
     if sort_by is None:
         # Always sort by ID to preserve ordering.
-        search = search.sort("id")
+        search = search.sort(LoreSortingFields.BASE_SORTING_FIELD)
+    # Temporary workaround; the values in sorting.py should be updated,
+    # but for now Haystack is still using them. Also, the hyphen is
```
pwilkins, line 175: Didn't we replace Haystack with elasticsearch_dsl?

giocalitri, line 175: I simply copied the comment, but as far as I can see, there is some backwards compatibility left in the code

pwilkins, line 175: Looking at all the occurrences of the term "haystack" (http://pastebin.com/DTVkGSk6) it appears that the use of the name was not replaced since the replacement had no impact on whether the code worked.  It's still confusing, though.  I'll create an issue proposing the replacement of the term "haystack" so the value of the change can be considered separately.

pwilkins:
As @giocalitri warned, this may get complicated.

Setting the default sort order to "relevance" moves the learning resource titled "Astronauts playing catch" to the top of the result list, but adding "Astronauts playing catch" to the description of another LR bumps the titled LR to second spot.

screen shot 2016-01-22 at 2 04 01 pm

Since I'm doing this locally, the fact that there are no LR descriptions (other than the one I added) may have unbalanced its relevance.

search/sorting.py
```diff
@@ -14,8 +14,9 @@ class LoreSortingFields(object):
     SORT_BY_NR_ATTEMPTS = ('nr_attempts', 'Number of Attempts (desc)', '-')
     SORT_BY_AVG_GRADE = ('avg_grade', 'Average Grade (desc)', '-')
     SORT_BY_TITLE = ('titlesort', 'Title (asc)', '')
+    SORT_BY_RELEVANCE = ('_score', 'Relevance (asc)', '')
```
noisecapella: Why the underscore here? Just curious

giocalitri: It is the name of the special field in the ES Index: see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html

giocalitri:
@pwilkins was the search string "Astronauts playing catch" ?
Can you check the score of the two learning resources: it is in the JSON returned by search API

noisecapella:
Code looks good, testing functionality now

noisecapella:
I was just wondering, shouldn't it be descending relevance since the most relevant item should be first?

giocalitri:
it actually is: I will remove the string (just relevance)

noisecapella:
Sounds good, feel free to merge after :+1:

giocalitri:
@noisecapella done: will merge after tests pass again

giocalitri:
Travis finally worked: merging

PR #894, opened 2016-03-11
Replaced the status app with django-server-status

### LINKED ISSUE #889: Integrate django-server-status
We have mitodl/django-server-status now which usurps the status app in this repo. We should port to that, which should be pretty effortless.

giocalitri:
#### What are the relevant tickets?

fixes #889 
#### What's this PR do?

replaces the built in status app with the `django-server-status` library
#### Where should the reviewer start?

`settings.py`
#### How should this be manually tested?

rebuild the container and check that the new response from the status contain the keys `celery` and `status_all`


noisecapella:
This generally looks good, although the celery portion of the app isn't working because there's no `USE_CELERY` setting defined. Is that a blocker?

giocalitri:
This generally looks good, although the celery portion of the app isn't working because there's no USE_CELERY setting defined. Is that a blocker?

Probably it is: I will look into it.

giocalitri:
@noisecapella: Now everything works.

@pdpinch this needs attention before releasing to production to check if everything works fine for the imports and everything that uses `celery` in general.

noisecapella:
Looks good now :+1:

